PR #37, opened 2026-06-17
spec: optional agent-identity binding in the Trust Record (#33)

### LINKED ISSUE #33: Carry the bound agent identity (manifest_id, agent_id) in the Trust Record (cMCP #302)
## Spec section
Section 3.1 (Trust Record logical schema) and 3.3 (Verification); schema/trace-claim.json; TrustRecord in src/agentrust_trace/models.py.

## Problem
The Trust Record subject is the gateway session SPIFFE URI (for example spiffe://cmcp.gateway/session/<uuid>). It carries no agent identity, so verification proves that a session ran an approved policy and catalog, but not that the agent the operator reviewed (the signed Agent Manifest agent_id) is the agent that acted. cMCP issue agentrust-io/cmcp#302 (Option A, confirmed) binds the manifest to the session at the gateway. For that binding to be verifiable offline (the Option B defense-in-depth cross-check), the bound identity has to travel in the Trust Record.

## Proposed change
Add an optional agent-identity block, distinct from subject (which stays the gateway session):

    "agent": {
      "agent_id": "spiffe://factory.example/agent/material-movement/dev",
      "manifest_id": "0197739a-8c00-7000-8000-000000000001",
      "binding": "svid-matched"
    }

- All fields optional; records without the block remain valid (backward compatible). subject is unchanged.
- A verifier holding the offline manifest can then cross-check: agent.agent_id == manifest.agent_id, agent.manifest_id == manifest.manifest_id, and the manifest bound policy/catalog hashes == record.policy.* . That is the Option B self-checking layer.
- Spec: new subsection under 3.1 for the agent block, and under 3.3 for the cross-check. Schema: add the optional agent object. Model: add an optional agent field to TrustRecord.

## Impact
- Backward compatible: yes (optional fields, existing records validate unchanged).
- Affects conformance level(s): Level 1+ (identity binding).
- Conformance tests that need updating: add present/absent agent-block cases; existing tests unaffected.
- Regulatory mapping impact: supports identity and accountability traceability (informational).

## Alternatives considered
- Overload subject to carry the agent identity: rejected, subject is the gateway session and conflating them loses the very distinction the binding must prove.
- Keep agent identity only in the cMCP audit chain, not the Trust Record: rejected, the Trust Record is the portable evidence a third party verifies offline, so the binding must travel with it.

aryanputta:
Closes #33.

## Problem
The Trust Record `subject` is the gateway session SVID. It proves a session ran an approved policy and catalog, but not that the agent the operator reviewed (the signed Agent Manifest `agent_id`) is the agent that acted. cMCP #302 (Option A) binds the manifest to the session at the gateway; for the Option B offline cross-check, the bound identity has to travel in the Trust Record.

## Change
Adds an OPTIONAL `agent` block, distinct from `subject`:

```json
"agent": {
  "agent_id": "spiffe://factory.example/agent/material-movement/dev",
  "manifest_id": "0197739a-8c00-7000-8000-000000000001",
  "binding": "svid-matched"
}
```

- **Spec** (`spec/trace-v0.1.md`): schema-table row, new §3.1.1, and an optional offline agent-identity cross-check in §3.3. Normative additions marked with `<!-- CHANGED: #33 -->`.
- **Schema** (`schema/trace-claim.json`): optional `agent` object with `agent_id` (SPIFFE/DID pattern), `manifest_id`, `binding`; `additionalProperties: false`.
- **Model** (`src/agentrust_trace/models.py`): `AgentIdentity` + optional `agent` field on `TrustRecord`, exported.
- **Tests**: present / absent / bad-scheme / extra-field cases.
- **CHANGELOG**: Unreleased entry.

## Compatibility
Backward compatible: all fields OPTIONAL, `subject` unchanged, existing records validate unchanged. A present-but-uncheckable `agent` block is unverified, not invalid. Affects Level 1+ (identity binding); existing conformance cases unaffected.

DCO signed-off. 41/41 tests pass locally.


imran-siddique (commented):
Good direction and the backward-compatibility story is sound. Before this merges there are several things that need addressing -- some are correctness issues, others are gaps a verifier implementer would hit.

**1. `binding` is a free string with no enumeration or registry**

The only value shown anywhere is `"svid-matched"`. If `binding` carries semantic meaning for verifiers (and it must, or why carry it?), a free string breaks interoperability: two implementations using `"svid-matched"` and `"SVID_MATCH"` describe the same thing but won't compare equal. Either enumerate the initial allowed values in the spec (e.g. `svid-matched`, `manifest-presented`, `operator-asserted`) and define an extension registry, or clarify that `binding` is informational-only and verifiers MUST NOT make trust decisions based on its value. Right now the spec implies it is meaningful but gives verifiers nothing to act on.

**2. `AgentIdentity` allows all three fields to be `None` simultaneously**

The model sets all three fields optional individually. That means `agent: {}` and `agent: {"binding": "svid-matched"}` (no `agent_id`, no `manifest_id`) are valid records. An `agent` block with no `agent_id` and no `manifest_id` is unverifiable and carries zero evidence -- it is noise. The spec should state whether a present `agent` block MUST have at minimum `agent_id` + `manifest_id` to be considered binding evidence (even if `binding` is optional). If a partial block is valid and "just informational," say so explicitly and specify what a verifier does with it.

**3. `manifest_id` has no format constraint**

The example shows a UUID v7. The field is a bare `str` with no pattern. A verifier implementing the §3.3 cross-check (step 2: `agent.manifest_id` equals manifest `manifest_id`) needs to know the format to reliably compare values. Is it a UUID? A URI? A hash? If UUID v7 is the intended format, add a pattern. If it is format-agnostic, state that explicitly and note that comparison is byte-equal string matching.

**4. §3.3 cross-check step 3 references a catalog hash field that does not exist in the schema**

> "The policy and catalog hashes the manifest binds equal `policy.bundle_hash` (and the catalog hash where the profile carries one)"

The current TRACE schema has `policy.bundle_hash` but no catalog hash field anywhere. "Where the profile carries one" implies a future or profile-specific extension. A verifier implementing step 3 today has nothing to check for the catalog half of this claim. Either remove the catalog hash part (it is unimplementable against v0.1 records), mark it explicitly as a future extension (with a TODO note and issue reference), or add the field to the schema now.

**5. No example JSON showing the `agent` block**

The `examples/` directory has `intel-tdx.json`, `amd-sev-snp.json`, `nvidia-h100.json`. None will show the `agent` block after this merges. A verifier implementer reading the spec needs a concrete end-to-end example. Add one -- either a new example file or an annotated variant of one of the existing ones.

**6. The spec says `subject` and `agent.agent_id` are "distinct" but does not forbid equality**

In the cMCP reference design the session SVID and agent identity are always different, but nothing in the spec prevents `agent.agent_id == subject`. Should a verifier reject or warn on this? If it is always a mistake, add a MUST NOT. If there are valid deployments where they can be equal, say so.

**7. `__pycache__` files committed**

The diff includes binary `.pyc` files (`src/agentrust_trace/__pycache__/`, `tests/__pycache__/`). These should not be committed. Add `__pycache__/` and `*.pyc` to `.gitignore` and amend.

**8. CHANGELOG conflict with #38**

Both #37 and #38 add a new `## [Unreleased]` section at the same position in CHANGELOG.md. Merging both as-is will produce a conflict or a duplicate section. One of them should rebase on the other so the changelog entries are merged into a single `[Unreleased]` block. Given that #38 is guidance-only and #37 changes the schema, #38 should rebase on #37.

---

The core idea is solid. Resolve 1-4 (the correctness and spec-completeness issues) first. 5-8 are mechanical but should be done before merge.

imran-siddique (commented):
Good work on the direction here - separating the agent identity from the gateway session subject is exactly right. A few things to work through before this is ready to merge.

**Compiled bytecode files in the diff**

The diff includes `src/agentrust_trace/__pycache__/*.pyc` and `tests/__pycache__/*.pyc`. These should never be committed; add `__pycache__/` and `*.pyc` to `.gitignore` (or fix the existing `.gitignore` if it has them already), then strip these files from the branch before merge. Committing bytecode makes the diff unreadable, bloats history, and will cause spurious conflicts on any Python version change.

**Spec-to-implementation misalignment**

The spec `agent` block has three fields: `agent_id`, `manifest_id`, `binding`. But the cMCP reference implementation (PR #315) carries `gateway.agent_identity` with seven fields: `manifest_id`, `agent_id`, `authenticated_subject`, `issuer`, `issuer_key_id`, `policy_bundle_hash`, `tool_catalog_hash`. These do not align. `authenticated_subject`, `issuer`, `issuer_key_id`, `policy_bundle_hash`, and `tool_catalog_hash` are present in the implementation but absent from the spec's `agent` block. Either the spec needs to expand the `agent` block to match what cMCP actually emits, or the implementation needs to be trimmed to match the spec. As it stands, a verifier reading the TRACE spec will not know to look for `policy_bundle_hash` in the `agent` block; a verifier reading a cMCP Trust Record will see fields the spec does not describe.

**`binding` is a free-form string with one documented value**

`binding` is described as "how the gateway established the binding (e.g. `svid-matched`)". There is only one example value. For TRACE to be interoperable, binding types need to be an enumerated set with defined semantics. `svid-matched` is presumably "the authenticated SVID subject matched `agent_id`" but this needs to be normative text, not just an example. A verifier cannot meaningfully interpret `binding` without a definition.

**Empty `agent` block passes validation**

All three fields in `AgentIdentity` are optional (`None` by default). This means `"agent": {}` is valid. An empty `agent` block carries no information but claims "agent identity binding was configured." Add a model validator that requires at least `agent_id` and `manifest_id` when the `agent` block is present, or make both required.

**Test coverage gap**

The test `test_agent_id_rejects_http_scheme` covers the SPIFFE/DID pattern enforcement. Add a parallel test for a valid DID URI (e.g. `did:web:factory.example`) to confirm DID-type identities are accepted. Currently only the rejection path is tested.

aryanputta:
Thanks for the thorough pass. Addressed all points; force-pushed a single clean commit.

1. **`binding` semantics** — now normative: `binding` is OPTIONAL and **informational only** (verifiers MUST NOT base trust on it), with an initial registered set `svid-matched` / `manifest-presented` / `operator-asserted` and an unrecognized-value rule. Registry maintenance tracked as §7 open question 8.
2. **Empty / partial `agent` block** — `agent_id` and `manifest_id` are now required when the block is present (pydantic + both schemas); `{}` and single-field blocks are rejected. Test: `test_agent_block_requires_agent_id_and_manifest_id`.
3. **`manifest_id` format** — documented format-agnostic (UUID/URI/digest), compared by byte-equal match; `minLength: 1`.
4. **Catalog-hash in §3.3 step 3** — removed from the implementable steps and called out as a future extension (no v0.1 field), tracked in §7 open question 9 + cMCP #302.
5. **Example** — added `examples/agent-bound-tdx.json`, wired into both schema- and model-validation test params.
6. **`subject` == `agent.agent_id`** — spec now permits equality explicitly; verifiers MUST NOT reject on it. Test: `test_agent_id_may_equal_subject`.
7. **Bytecode** — added `.gitignore`; stripped `__pycache__` and rebuilt the branch as one clean commit.
8. **CHANGELOG** — #38 is now rebased on this PR; both entries share one `[Unreleased]` block, no conflict.

**Spec-to-impl misalignment (#315):** rather than expand the canonical block, I documented the layering: the canonical `agent` block is the minimal *portable* identity, and a profile MAY carry richer evidence in its own addenda — cMCP's `gateway.agent_identity` (`authenticated_subject`, `issuer`, `issuer_key_id`, `policy_bundle_hash`, `tool_catalog_hash`) is now named explicitly in §3.1.1 as exactly that. This keeps PKI and catalog concerns out of the vendor-neutral record and needs no churn on #315. Happy to flip to a canonical superset if you'd prefer the spec own those fields.

**Coverage gap** — added `test_agent_id_accepts_did_uri` (positive DID path).

Also synced `src/agentrust_trace/schema/trace-v0.1.json` (the schema `validate()` actually loads), which the first revision missed. 47 tests pass, ruff clean.

aryanputta:
Rebased this branch on current upstream main and resolved the model conflict by keeping upstream's stricter subject validation while preserving the optional agent identity block. I also confirmed there are no tracked __pycache__ or .pyc files.

Verification run locally:
- PYTHONPATH=src python3 -m pytest -> 50 passed
- python3 -m ruff check . -> passed
- API smoke validating and signing examples/agent-bound-tdx.json -> passed
- git diff --check -> clean

PR #38, opened 2026-06-17
spec: external execution evidence verification guidance (#34)

### LINKED ISSUE #34: Recognize external execution evidence (controller-signed receipts) on the verification side (cMCP #301)
## Spec section
Verification (Section 3.3), Scope (2.4 and 3.4), and the cMCP reference implementation (Section 5).

## Problem
cMCP issue agentrust-io/cmcp#301 (Option A, confirmed) adds an optional external_execution_evidence receipt to a cMCP audit entry: an independent authority (for example a safety controller) signs an assertion about a specific call (issuer, issuer_key_id, signature, evidence_hash, evidence_type, linked_call_id), bound by linked_call_id to the entry call_id. It is distinct from response_payload_hash (what the gateway forwarded). cmcp-verify now checks the receipt when configured with the issuer trusted key. TRACE should describe how a verifier treats such evidence so the spec and cmcp-verify stay aligned.

## Proposed change
Add verification-side guidance that:
- External execution evidence is out-of-band evidence signed by a non-gateway authority, distinct from gateway-produced fields. It is only as trustworthy as the issuer key, whose trust is established out of band (a PKI / trust-anchor concern, the same shape as the manifest issuer anchor discussed for cMCP #302).
- A conforming verifier, when supplied an issuer trust anchor, verifies the issuer signature over the canonical receipt (all fields except signature) and checks linked_call_id against the bound call.
- Verifying external execution evidence is optional and does not affect the validity of the gateway-produced Trust Record. A present-but-unconfigured receipt is unverified, not invalid.
- This stays within TRACE's permanent scope boundary: TRACE binds evidence, it does not assert that a physical action occurred or that it is functional-safety certified.

No required-field change to the Trust Record. The receipt lives on the cMCP audit entry (cMCP schema), within the audit chain the Trust Record already commits, so this is verification guidance rather than a wire-format change.

## Impact
- Backward compatible: yes (guidance plus optional verification; no Trust Record format change).
- Affects conformance level(s): Level 1+ optionally.
- Conformance tests that need updating: optionally add an external-evidence verification case in trace-tests; existing tests unaffected.
- Regulatory mapping impact: supports accountability for independent downstream actions (informational).

## Alternatives considered
- Put the receipt into the Trust Record itself: rejected, the receipt is per-call and belongs on the cMCP audit entry; the Trust Record already commits the audit chain, so referencing it there is sufficient and keeps the Trust Record gateway-scoped.
- Treat a present-but-unverifiable receipt as a hard record failure: rejected, that would let an unconfigured verifier reject otherwise-valid records; unverified-optional is the right default.

aryanputta:
Closes #34.

## Problem
cMCP #301 (Option A, confirmed) adds an optional external execution evidence receipt to a cMCP audit entry: an independent authority (e.g. a safety controller) signs an assertion about a specific call (`issuer`, `issuer_key_id`, `signature`, `evidence_hash`, `evidence_type`, `linked_call_id`), bound by `linked_call_id` to the entry `call_id`. `cmcp-verify` checks the receipt when configured with the issuer trusted key. TRACE should describe how a verifier treats such evidence so the spec and cmcp-verify stay aligned.

## Change
Adds **§3.3.1 External execution evidence (OPTIONAL)** under Verification:

- External execution evidence is out-of-band evidence signed by a non-gateway authority, distinct from gateway-produced fields, only as trustworthy as the issuer key (trust established out of band, a PKI / trust-anchor concern).
- A conforming verifier, when supplied an issuer trust anchor, verifies the issuer signature over the canonical receipt (all fields except `signature`) and checks `linked_call_id` against the bound entry.
- Verification is OPTIONAL; a present-but-unconfigured receipt is **unverified, not invalid**, and does not affect the validity of the gateway-produced Trust Record.
- Stays within TRACE's permanent scope boundary (§2.4): TRACE binds evidence, it does not assert a physical action occurred or that it is functional-safety certified.

## Compatibility
Backward compatible: guidance plus optional verification, **no Trust Record format change**. The receipt lives on the cMCP audit entry, within the audit chain the Trust Record already commits, so no schema or model change is needed. Affects Level 1+ optionally; existing conformance tests unaffected.

DCO signed-off. 37/37 tests pass locally.


imran-siddique (commented):
The scoping here is correct -- treating the receipt as evidence the Trust Record binds rather than a field the Trust Record carries is the right design. A few things need to be resolved before this can merge.

**1. Broken internal reference: "steps 1-7 above" does not exist in this section**

The closing paragraph says:

> "the verification of steps 1-7 above is unchanged"

§3.3.1 defines exactly 2 steps. There are no steps 1-7 in this section. The §3.3 base verification protocol has 7 steps, but "above" in the context of §3.3.1 refers to this subsection. This will confuse any reader who tries to follow the reference. Either change "steps 1-7 above" to "the standard verification protocol in §3.3" (naming the actual section), or restructure the sentence to say explicitly that the gateway-produced record validation is unaffected.

**2. "Canonical receipt" uses the word canonical without defining canonicalization**

> "Verifies the issuer signature over the canonical receipt (every field except `signature`)"

TRACE uses JCS (RFC 8785, canonical JSON serialization) for the main record signing. Is JCS also the canonicalization for the receipt? If yes, state it explicitly. If the receipt uses a different byte representation for signing (e.g. sorted-key JSON without whitespace, or CBOR), define it here. A verifier implementing step 1 needs to know exactly what byte sequence to verify the signature over. "Every field except `signature`" is not sufficient without specifying the serialization.

**3. No normative reference to the receipt schema**

The receipt fields (`issuer`, `issuer_key_id`, `signature`, `evidence_hash`, `evidence_type`, `linked_call_id`) are named here but their types, formats, and constraints are defined in cMCP #301, not in this spec. A TRACE verifier implementing §3.3.1 cannot parse or verify a receipt without knowing the schema. Add a normative reference: something like "The receipt structure is defined in [cMCP audit spec, §X]" with a link or citation. Without it, §3.3.1 is non-implementable standalone.

**4. Step 2 does not define how `linked_call_id` resolution works**

> "Checks `linked_call_id` resolves to the audit entry it claims to bind"

This is underspecified. The Trust Record commits the audit chain via `tool_transcript.hash`. A verifier needs to: (a) obtain the audit chain, (b) locate the entry whose `call_id` equals `linked_call_id`. Step 2 should say something like: "Confirms that the audit chain committed by `tool_transcript.hash` contains an entry with `call_id` equal to `linked_call_id`." Without this, "resolves" is ambiguous about where the verifier looks and how.

**5. No failure behavior defined**

The base §3.3 protocol specifies what happens when verification fails. §3.3.1 defines when to verify and what to check, but says nothing about what a verifier MUST or SHOULD do when step 1 or step 2 fails. Should it mark the external evidence invalid while keeping the Trust Record valid? Should it fail the entire record? Return a structured error status alongside the main verification result? Specify at minimum whether a step 1 or step 2 failure affects the validity of the enclosing Trust Record.

**6. CHANGELOG conflict with #37**

Both PRs add a new `## [Unreleased]` block at the same position in CHANGELOG.md. This will conflict on merge. This PR should rebase on #37 and merge its changelog entry into the single `[Unreleased]` block.

**7. `__pycache__` files committed**

Same issue as #37 -- binary `.pyc` files in the diff. Add `__pycache__/` to `.gitignore` and amend.

---

Issues 1, 2, 3, and 5 are correctness or implementability issues -- a verifier author cannot implement §3.3.1 correctly from this text alone. 4 is a precision gap. 6 and 7 are mechanical. Address 1-5 before requesting re-review.

imran-siddique (commented):
The scoping here is right - framing external execution evidence as "TRACE binds evidence; it does not assert that a physical action occurred" keeps the spec honest about what the TEE can and cannot prove. A few things to address:

**Compiled bytecode files**

Same issue as PR #37: `src/agentrust_trace/__pycache__/*.pyc` and `tests/__pycache__/*.pyc` are included. Strip these before merge. (If you fix `.gitignore` in #37 first, make sure the fix is rebased into this branch too so it does not reappear here.)

**Broken cross-reference**

§3.3.1 ends with: "verification of steps 1-7 above is unchanged." §3.3 does not have numbered steps. This reference is broken. Either add explicit numbered steps to §3.3 or replace "steps 1-7" with a concrete reference to what is unchanged.

**"Canonical receipt" is undefined**

The spec says verifiers should "verify the issuer signature over the canonical receipt (every field except `signature`)". What is the canonicalization algorithm? Is this RFC 8785 (JCS)? Alphabetically sorted JSON? Something custom? Without a defined canonicalization, two compliant implementations will produce different byte strings and one will reject signatures the other accepts. This needs a normative definition (even a one-liner pointing to RFC 8785 would be sufficient if that is the intended algorithm).

**`evidence_hash` pre-image is still undefined**

The guidance describes verifying the issuer signature over "the canonical receipt" but does not specify what bytes produce `evidence_hash`. The cMCP audit entry schema defines `evidence_hash` as a "hash of the evidence" without further specification. For a TRACE-level spec to be useful, it needs to say what is hashed, in what encoding, with what canonicalization. Otherwise `evidence_hash` is an opaque blob that implementers cannot independently reproduce.

**Interaction with TRACE's audit chain commitment**

The spec notes the receipt "lives on the cMCP audit entry, inside the audit chain the Trust Record already commits to." That is correct. But §3.3.1 does not say how a verifier locates the relevant audit entry given a `linked_call_id`. The Trust Record carries `audit_chain_tip` and `audit_chain_root` but not the full chain. Does the verifier need the exported audit bundle? If so, say so. If the receipt is meaningful without the bundle, say that instead.

aryanputta:
Thanks — addressed all points; force-pushed clean (rebased on #37).

1. **Broken `steps 1-7 above`** — replaced with an explicit reference to the §3.3 protocol (steps 1–7), and the failure-behavior paragraph now names §3.3 as what establishes record validity. No more ambiguous self-reference.
2. **`canonical receipt` undefined** — now defined as the receipt with `signature` removed, serialized with **RFC 8785 (JCS)**, matching §3.2.2 (with a profile-override escape hatch).
3. **`evidence_hash` pre-image** — clarified that `evidence_hash` is an opaque digest defined by the receipt schema and the issuer's `evidence_type`; **TRACE does not specify its pre-image and a verifier does not recompute it**. The issuer signature (step 1) commits `evidence_hash`, so signature verification is what binds it; interpreting the artifact is out of scope (§2.4).
4. **Normative ref to the receipt schema** — §3.3.1 now states the receipt fields/formats are defined normatively by the cMCP audit-entry schema (#301), not by this document.
5. **Locating the entry via `linked_call_id`** — step 2 now says the verifier obtains the exported audit bundle (`/audit/export`) for the chain committed by `tool_transcript.hash` and confirms exactly one entry with `call_id == linked_call_id`; explicitly notes the Trust Record alone carries only tip/root, not the full chain.
6. **Failure behavior** — added: receipt verification is reported separately from the Trust Record verdict; a step 1/2 failure marks the receipt **invalid** but MUST NOT by itself reject the gateway-produced record (a relying party MAY escalate by its own policy). A present-but-unconfigured receipt is **unverified, not invalid**.
7. **CHANGELOG + bytecode** — rebased on #37, so `.gitignore` carries over (no `__pycache__`) and both changelog entries share one `[Unreleased]` block.

aryanputta:
Rebased/restacked this branch on top of the refreshed agent-identity PR and current upstream main. The branch now only adds the external execution evidence guidance on top of #37, limited to CHANGELOG.md and spec/trace-v0.1.md.

Verification run locally:
- PYTHONPATH=src python3 -m pytest -> 50 passed
- python3 -m ruff check . -> passed
- API smoke validating, signing, and verifying examples/agent-bound-tdx.json -> passed
- git diff --check -> clean

PR #39, opened 2026-06-17
docs: define external execution evidence verification

### LINKED ISSUE #34: Recognize external execution evidence (controller-signed receipts) on the verification side (cMCP #301)

carloshvp:
## Summary
- add TRACE verification guidance for optional external execution evidence
- clarify that untrusted/unconfigured receipts are unverified addenda, not core Trust Record failures
- document the physical-execution and functional-safety boundary in the spec and limitations

Closes #34.

## Validation
- `PYTHONPATH=src python3 -m pytest`

Note: `python3 -m ruff check .` currently reports an unused `json` import in `tests/test_sign.py` on fresh `origin/main`; this PR does not touch that file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,11 @@ Seven open questions requiring founding-member input before v0.2 are documented 
   Previously `^spiffe://` only; now `^(spiffe://|did:)`. Additive, backward-compatible.
   DID-native runtimes (e.g. AGT `did:mesh:` identities) no longer require a parallel SPIFFE identity.
   Closes: microsoft/agent-governance-toolkit ADR-0032, agentrust-io/trace-spec#35.
+- Add verification guidance for optional external execution evidence bound by
+  profile-specific audit-chain entries. A verifier may check controller-signed
+  receipts when configured with the issuer trust anchor; otherwise the receipt
+  is unverified addenda and does not affect core Trust Record validity.
+  Closes: agentrust-io/trace-spec#34.
 
 ### Schema
 
--- a/LIMITATIONS.md
+++ b/LIMITATIONS.md
@@ -16,6 +16,9 @@ The `policy.bundle_hash` field attests that a specific policy was in force at ru
 **What happened inside the model**
 The call transcript records tool invocations, arguments, and responses that are observable at the gateway boundary. It does not record the model's internal chain-of-thought, intermediate reasoning, or context window contents. Reasoning that influences behavior without producing a tool call is not captured.
 
+**Physical execution or functional safety**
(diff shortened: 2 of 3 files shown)


imran-siddique:
Closing in favour of #38 — that PR covers the same issue (#34) with the complete implementation (schema update, models.py, tests, and spec section §3.3.1). Merging both would duplicate the change and create conflicts. Thanks for the contribution; the approach in your docs-only change is sound and the substance is captured in #38.

PR #48, opened 2026-06-18
Add tpm2.json configuration file

harshnair75567-cloud:
## What this changes

Adds examples/tpm2.json . Complements existing intel-tdx, amd-sev-snp, and nvidia-h100 examples. Validates against schema/trace-claim.json .
also removing import from line 4 of test_sign.py which was an unused json import causing a ruff F401 lint error.

## Type of change

- [ x ] Editorial (typo, link fix, clarification — no normative effect)
- [ ] Schema change
- [ x ] Example addition

## Spec section
 example addition only, no normative text modified
## Checklist

- [ x ] DCO sign-off on all commits (`git commit -s`)

### DIFF
--- a/examples/tpm2.json
+++ b/examples/tpm2.json
@@ -41,13 +41,12 @@
     "timestamp": 1750676203
   },
   "transparency": "https://registry.agentrust.io/claim/trace-2026-06-23T09:16:40Z-tpm2-b0c1d2",
-  "cnf": {
-    "jwk": {
-      "kty": "EC",
-      "crv": "P-256",
-      "x": "f4OxZX_x51sR3pCsmUQjM_-iX894U3FGuCcQMFG5eWI",
-      "y": "bFHAgHvSiSASfHAhbHGpFGZVnaSqHL1vVXHnqBypNPk",
-      "kid": "tpm2-workload-key-2026-06-23"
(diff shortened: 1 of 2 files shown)


imran-siddique (changes requested):
One issue to fix before merge: the `cnf.jwk` in this example uses EC P-256, but the TRACE spec requires Ed25519 (OKP crv=Ed25519).

From the test suite (TR-SIG-001):
- Positive case: `{"kty":"OKP","crv":"Ed25519","x":"..."}`
- Negative case: ES256, RS256 — P-256 falls into the negative case

The TPM2 platform uses its AK/EK for attestation internally, but the signing key that goes into `cnf.jwk` is an Ed25519 key generated by cMCP (bound to the TPM quote's `qualifying_data`). Please update the example to use:

```json
"cnf": {
  "jwk": {
    "kty": "OKP",
    "crv": "Ed25519",
    "x": "<base64url-encoded Ed25519 public key>",
    "kid": "tpm2-workload-key-2026-06-23"
  }
}
```

Everything else looks correct — the `tpm2` platform value, SLSA 3, the measurement format, the firmware_version field. Just the key type needs updating.

Once that's fixed this is good to merge.

PR #68, opened 2026-06-28
fix(models): block private key parameters in cnf.jwk

harshnair75567-cloud:
## What this changes

JWK._require_key_material only enforced that public key fields were present — it never blocked private key parameters (d, p, q, dp, dq, qi, k) from being stored in cnf.jwk. With extra="allow" on the model, a record carrying private key material passed model_validate() silently. Added _JWK_PRIVATE_PARAMS constant and a check in the validator that raises if any private parameter is present. Added one regression test.

## Type of change

- [x ] Non-breaking spec change (new optional field, new platform profile, informative addition)
- [ ] Schema change
- [ ] Example addition

## Spec section

 §3.2.1 Signing and key management

## Checklist

- [ ] DCO sign-off on all commits (`git commit -s`)

### DIFF
--- a/src/agentrust_trace/models.py
+++ b/src/agentrust_trace/models.py
@@ -5,6 +5,7 @@ from typing import Annotated, Literal
 from pydantic import BaseModel, ConfigDict, Field, model_validator
 
 _DIGEST_RE = r"^sha(256:[0-9a-f]{64}|384:[0-9a-f]{96})$"
+_JWK_PRIVATE_PARAMS = frozenset({"d", "p", "q", "dp", "dq", "qi", "k"})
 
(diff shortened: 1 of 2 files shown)


imran-siddique:
Thank you, this is a real catch. Blocking private key params (`d, p, q, dp, dq, qi, k`) from `cnf.jwk` under `extra="allow"` is a genuine security hardening and we want it in.

Process note: changes to the trace-spec models are normative, and per our contribution policy we land them through internal spec review rather than direct external merges. So I won't merge this PR directly, but the fix is going in with credit to you, tracked in #70. Please keep poking at the models, this is exactly the kind of review we want.

PR #72, opened 2026-06-30
fix(models): reject private key material in cnf.jwk

### LINKED ISSUE #70: fix(models): block private key params in cnf.jwk (extra=allow leak)
Reported by @harshnair75567-cloud in #68. `JWK._require_key_material` only checks that public key fields are present; under `extra="allow"` a record carrying private key parameters (`d, p, q, dp, dq, qi, k`) passes `model_validate()` silently. Add a `_JWK_PRIVATE_PARAMS` check that rejects any private parameter, plus a regression test. Implementing internally per the spec contribution policy, with credit to @harshnair75567-cloud. Refs #68.

imran-siddique:
`JWK` uses `extra="allow"` (JWK members vary by key type per RFC 7517), which meant private/secret params (`d, p, q, dp, dq, qi, k`) on a `cnf.jwk` passed `model_validate()` and were stored silently. A `cnf.jwk` is a public proof-of-possession key, so the validator now rejects any private member. Regression test added.

Originally reported with a fix by @harshnair75567-cloud in #68; implemented here with credit, per the spec contribution policy (normative model changes land through internal review). Closes #70.


harshnair75567-cloud:
Thanks for the credit and the clear explanation on the contribution process — makes sense given it's a normative change. Glad the catch was useful. Will keep digging through the models for anything else that needs hardening.

PR #92, opened 2026-07-05
docs: clarify action receipt verification boundary

carloshvp:
## Summary

Refs #66.

- clarifies the action-receipt verification boundary for embodied workflows in the spec
- adds verification-guide taxonomy for accepted receipts, valid rejections, missing required receipts, and invalid receipts
- adds informative action-receipt fixture shapes without changing the TRACE schema

## Validation

- `.venv/bin/ruff check src tests`
- `.venv/bin/mypy src/agentrust_trace`
- `.venv/bin/pytest --cov=agentrust_trace --cov-report=term-missing`
- `git diff --check`

Docs build note:

- `.venv/bin/mkdocs build --strict --site-dir /tmp/trace-action-receipts-site-check` currently fails before rendering changed pages because the existing MkDocs config uses `docs_dir: .`, which MkDocs rejects as the parent directory of the config file.

### DIFF
--- a/docs/verification.md
+++ b/docs/verification.md
@@ -107,6 +107,36 @@ agentrust-trace verify-scitt session.trace.json \
 
 A valid SCITT receipt proves the record was included in the log and cannot be retroactively removed or modified.
 
+## Action receipts and embodied workflows
+
+Some deployments attach per-action receipts below the session layer. For
+example, an embodied-agent controller can sign a receipt that says a specific
+call was accepted, rejected, aborted, or handed off to another authority. These
+receipts extend the audit chain; they do not replace Trust Record verification.
+
+Keep the verification results separate:
+
+| Evidence layer | What to verify | What not to infer |
+|---|---|---|
+| Session evidence | TRACE signature, freshness, policy hash, runtime measurement, transcript hash | Complete physical-world state |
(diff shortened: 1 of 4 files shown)


imran-siddique (approved):
Approving. The new §3.3.2 (informative) and the verification-guide additions are accurate and well-scoped.

- Purely informative and additive: no change to `schema/trace-claim.json`, and the new `examples/action-receipts/` fixtures are explicitly marked as non-Trust-Records not validated by the schema (`examples/README.md` + `examples/action-receipts/README.md`), so `additionalProperties: false` is not at risk.
- The verification boundary is stated correctly and consistently with the existing §3.3.1 trust-boundary paragraph: a signed controller rejection is *valid negative evidence*, not a TRACE verification failure, while a signed acceptance does not prove physical completion or functional-safety certification unless a stricter profile defines and trusts that external claim. The four outcomes (`receipt_valid_accepted` / `receipt_valid_rejected` / `receipt_missing_required` / `receipt_invalid`) are coherent and match between spec, verification.md, and the fixture README.
- Correctly keeps the action-receipt requirement axis independent of build-provenance depth, and the verifier steps (recompute digest, trusted issuer key not the embedded key, chain ordering, session/call binding, report missing/stale separately) mirror the §3.3.1 rules.

CI green. LGTM.

PR #94, opened 2026-07-06
docs: clarify cnf signing preimage

### LINKED ISSUE #77: spec: decide and document the cnf-in-preimage canonical pre-image (normative)
Tracked follow-up from the June 2026 org-wide security hardening review. Wave 1/2 fixes are merged and published; this is remaining hardening.

The SDK signs over the record including `cnf` (correct), but some doc snippets historically excluded it. Make the normative pre-image definition explicit in the spec so all implementations agree (recommend: include `cnf`, exclude only `signature`). Confirm SDK + spec + all docs match.

carloshvp:
## Summary

Fixes #77.

- Clarify that the TRACE signature pre-image includes `cnf` / `cnf.jwk` and excludes only the embedded `signature` field.
- Align the spec, verification docs, schema descriptions, model docstring, and signing utility docs on the same rule.
- Add a regression test showing that tampering with `cnf.jwk` invalidates the record signature.

## Validation

- `/tmp/trace-spec-venv/bin/python -m pytest -q` -> 79 passed
- `/tmp/trace-spec-venv/bin/ruff check src tests`
- `/tmp/trace-spec-venv/bin/mypy src/agentrust_trace`
- `git diff --check`

Docs build note:
- `/tmp/trace-spec-venv/bin/mkdocs build --strict --site-dir /tmp/trace-spec-site-check` currently fails before page rendering because the existing MkDocs config uses `docs_dir: .`, which MkDocs rejects as the parent directory of the config file. This PR does not change that config.

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -19,7 +19,7 @@ JSON Schema for the TRACE v0.1 Trust Record. Source: [`schema/trace-claim.json`]
 | `appraisal` | object | **yes** | Verifier judgment |
 | `transparency` | string | **yes** | SCITT transparency log anchor URI (empty string if not anchored) |
 | `cnf` | object | **yes** | Confirmation method — contains the `jwk` signing key |
-| `signature` | string | **yes** | Base64url Ed25519 / ES256 / ES384 signature over the record |
+| `signature` | string | **yes** | Base64url Ed25519 / ES256 / ES384 signature over the canonical record with only `signature` absent; `cnf` is included |
 
 ## `model`
 
--- a/docs/verification.md
+++ b/docs/verification.md
@@ -8,7 +8,7 @@ This is the normative protocol from [§3.3 of the spec](../spec/trace-v0.1.md).
 
 ### Step 1 — Parse the envelope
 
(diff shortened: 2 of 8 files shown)


imran-siddique (approved):
Approving. Verified the clarification is technically accurate against the implementation, not just internally consistent.

- `src/agentrust_trace/sign.py` `sign_record()` computes the preimage as `_canonical_bytes({k: v for k, v in payload.items() if k != "signature"})` where `payload` already has `cnf.jwk` merged in — so the signed preimage is the full record with **only** `signature` removed and `cnf` included. The new prose in `docs/verification.md`, `spec/trace-v0.1.md` (§3.2.2 canonical-form step 1), `docs/schema.md`, both `schema/*.json` descriptions, and the `models.py` docstring now all state this correctly.
- This also fixes a prior contradiction: `docs/verification.md` previously said the preimage removed `signature` *and* `cnf`, yet the code snippet right below it excluded only `signature`. Now aligned.
- The added regression test `tests/test_sign.py::test_verify_record_raises_for_tampered_cnf_jwk` proves `cnf.jwk` is inside the signed preimage (tampering raises `InvalidSignature`), and it passes in CI (79 passed, all checks green).

Accurate crypto clarification, fully consistent across spec/docs/schema/code. LGTM.

Issue #97, opened 2026-07-08
Related work / interop: mapping Acta decision receipts into TRACE evidence


tomjwxf:
Spec feedback in the spirit of the contribution table (no product code, just interop groundwork).

TRACE's evidence layer and the Acta receipt format (draft-farley-acta-signed-receipts, revision 02 on the IETF datatracker) sit at adjacent layers of the same problem: TRACE binds evidence to hardware-attested runtime identity; Acta receipts bind per-decision facts (tool, decision, policy digest, chain position) under an Ed25519 signature over canonical JSON, designed to verify offline with no issuer contact. Deployments that want both properties (attested runtime AND per-decision evidence a counterparty can verify independently) currently have no documented way to compose them.

Two concrete suggestions:

1. **A short Related Work or Interop section in the spec** acknowledging software-signed decision-receipt formats as a complementary evidence class, and stating TRACE's position on carrying or referencing them (e.g. an external-evidence claim carrying a digest of a receipt chain head). Even a sentence saying "out of scope, reference by digest" removes ambiguity for implementers who face both.

2. **A field-level mapping as an appendix or companion doc**: Acta receipt fields to EAT claims (the receipt's issuer/kid to the attester identity, the decision payload as a claims set, the chain head as evidence continuity). I am happy to draft this mapping. The Acta side has a shared conformance suite (ScopeBlind/agent-governance-testvectors, Apache-2.0, independent implementations cross-verifying byte-identical canonical output) that a mapping could reference for testable fixtures, which fits the verification culture of the trace-tests suite.

Context and disclosure: I author the Acta draft and maintain protect-mcp; I am also an agentrust-io org member. The aim here is the composition being specified once rather than reinvented per deployment; the two evidence classes strengthen each other (hardware attestation says the runtime is what it claims; decision receipts say what it decided, verifiably, after the fact).

carloshvp:
Issue triage

This seems worth tracking, but I would scope it as interop guidance/conformance mapping rather than a new normative receipt format for v1.0. The active spec anchors are #66 for `verification.action_receipts`, #95 for action-receipt conformance cases, and the closed external-execution-evidence discussion in #34.

Suggested acceptance bar:

- TRACE should carry or reference Acta as external evidence by digest/chain head plus issuer/key metadata, not embed a second receipt schema into the core Trust Record.
[...]

PR #98, opened 2026-07-08
docs(crosswalks): Acta decision receipts as a second action-receipt profile

tomjwxf:
Addresses the acceptance bar @carloshvp set in #97, one point at a time:

> TRACE should carry or reference Acta as external evidence by digest/chain head plus issuer/key metadata, not embed a second receipt schema into the core Trust Record.

Done as a cross-walk doc (`docs/crosswalks/acta-decision-receipts.md`), following the exact pattern of the existing OWASP cross-walk: non-normative, no schema/wire-format change, registered in `mkdocs.yml` the same way. Acta stays an external, independently specified format (draft-farley-acta-signed-receipts, rev 02); nothing here touches `schema/trace-claim.json`. The doc includes a concrete (illustrative, non-normative) Trust Record evidence entry showing exactly that shape: chain head + issuer/key metadata, where the chain head is the real, recomputable digest of the fixture chain's latest receipt.

> A mapping should state exactly which Acta fields satisfy the action-receipts verifier obligations: canonical preimage, signature verification against a trusted or manifest-bound key, chain ordering, and TRACE/cMCP session or call binding.

The field-mapping table walks the exact shape from spec §3.3.2 / `examples/action-receipts/README.md` (issuer, issuer_key_id, linked_call_id/session_id, evidence_hash, previous_receipt_hash, decision, signature) against the exact Acta envelope fields, and restates the five verifier checks from the embodied profile for this one, plus a sixth (`policy_digest` freshness) that is not a signature check and should not be conflated with one.

> Any appendix should include negative fixtures: wrong kid/key, broken chain, mismatched session/call binding, and a valid receipt that represents rejection rather than successful physical or business outcome.

Six fixtures in `examples/action-receipts/acta/`, all real (an actual Ed25519 signer over JCS-canonical bytes, not hand-typed JSON; generator included): valid accepted, valid denied (the rejection case), signature/key mismatch, broken chain, stale policy_digest, and mismatched session/call binding. The last two are deliberately receipts whose signatures verify cleanly but which fail a non-cryptographic check (stale authorization basis, wrong session), to keep those distinctions explicit rather than folding them into the crypto check.

> Keep the boundary explicit: Acta/TRACE composition verifies decision/action evidence integrity; it does not prove physical completion or safety certification.

Restated directly against §2.4 and the #34 external-execution-evidence boundary, generalized to the software (non-embodied) case this profile covers.

Every fixture and every doc-internal anchor link was checked against this repo's actual spec headings and fixture format before commit, not assumed. Happy to adjust anything that does not fit the maintainers' preferred shape for this.


carloshvp (changes requested):
Requesting changes

1. **The fixtures are not `draft-farley-acta-signed-receipts-02` envelopes, despite being presented as exact/independently verifiable Acta fixtures.** Draft-02 §2 requires a top-level `{ "payload": {...}, "signature": { "alg", "kid", "sig" } }` envelope, with common fields such as `type`, `issued_at`, and `issuer_id` inside `payload`; §3.1 defines the access-decision fields as `type: "protectmcp:decision"`, `tool_name`, `decision`, `reason`, etc. These fixtures instead put `v`, `type`, `algorithm`, `kid`, `issuer`, and `issued_at` at the top level, use a bare signature string, and place a different custom shape under `payload`. A conforming draft-02 verifier cannot decode them, so the cross-walk's field mapping and the README's independent-verification claim are currently misleading.

Please regenerate the fixtures using the draft-02 envelope and field names (or explicitly define this as a separate profile rather than claiming draft-02 conformance), then update the mapping and verifier instructions accordingly. Add an automated check that parses each fixture against the claimed Acta schema and verifies the expected positive/negative result; otherwise this drift can recur while the repository's Python tests still pass.

Validation: compared the PR diff and generated fixtures against the published draft-02 receipt structure and access-decision schema; repository Python 3.11/3.12 and CodeQL checks pass, while the contributor gate currently fails. I did not run an Acta verifier because the PR does not add one and the envelopes fail schema inspection before signature verification.

carloshvp (changes requested):
Requesting changes

The cross-walk direction is useful, but two fixtures do not currently exercise the verifier obligations they claim to cover:

1. `examples/action-receipts/acta/03-signature-key-mismatch.json` is byte-for-byte identical to `01-valid-accepted.json`. Following the README's independent-verification steps with the published `signer-public-key.txt`, fixture 03 verifies successfully. The unrelated key exists only transiently inside `gen.mjs`, so the committed fixture set gives an independent verifier no mismatched key or key-resolution input with which to reproduce the advertised failure. Please commit the mismatched public-key/key-set context (or encode a resolvable `kid` mismatch) and make the verification instructions deterministic.

2. `04-broken-chain.json` changes `parent_receipt_hash` after signing, so it only proves that payload tampering invalidates an Ed25519 signature. It does not test chain verification. A valid issuer can sign a receipt whose `parent_receipt_hash` does not match the actual predecessor; that signature must pass while the separate chain-link check fails. Please generate fixture 04 by signing an intentionally wrong parent hash, then verify both conditions independently: signature valid, predecessor digest mismatch.

Validation performed on head `82e56463aaf0a641933d93a68d891d4066573cbd`:

- Independent Ed25519 verification over compact, recursively key-sorted payload JSON with the committed signer public key: `01` PASS, `03` PASS, `04` FAIL, `05` PASS, `06` PASS.
- Confirmed `01-valid-accepted.json` and `03-signature-key-mismatch.json` are byte-identical.

I stopped at these conformance blockers because they need to be corrected before the fixtures can support the cross-walk's interoperability claims.

tomjwxf:
@carloshvp both reviews were correct, and thank you for validating against the published draft rather than taking the PR's description at its word. All three findings are fixed in 9053f36:

**1. Envelope conformance.** You were right: the fixtures used an internal implementation shape, not the draft-02 envelope. They are now conforming receipts: `{ payload, signature: { alg: "EdDSA", kid, sig } }`, access-decision payload per s3.1, `issuer_id` matching `signature.kid`, `kid` in the s2.1.1 recommended `sb:issuer:<base58-12>` format, PureEdDSA directly over JCS(payload) per s5.6, and `previousReceiptHash` as the bare-hex s5.7 hash over the predecessor's entire envelope including its signature. The crosswalk doc and mapping table are rewritten against the real draft field names (this also produced a better mapping: `evidence_hash` now maps to the draft's own `payload_digest` rather than a hand-wave).

**2. Fixture 03.** Now signed by a second, committed key (`mismatched-signer-public-key.txt`) while claiming the signer's `kid`, so resolving `kid` to the published `signer-public-key.txt` fails deterministically from committed bytes alone. No transient state in the generator is needed to reproduce the failure, and a test asserts the mismatch signature is a genuine signature by the committed second key rather than random bytes.

**3. Fixture 04.** Your framing was the correct one and the docs now say so: it is validly signed over a wrong `previousReceiptHash` (a real receipt hash, just the wrong predecessor's, so it models reordering rather than garbage). Signature check passes, chain recomputation fails, and the previous claim that chain and signature failures collapse is gone; under s5.7 they are distinct checks and 04 exists to prove it.

**Automated check.** `tests/test_acta_fixtures.py` verifies every fixture's structure against the draft-02 envelope and its expected positive/negative result from a machine-readable `expected.json`, using only dependencies this project already declares (`rfc8785` for JCS, `cryptography` for Ed25519). Full suite passes locally, 97 tests. A side benefit: the Python RFC 8785 reference now cross-validates the JS generator's canonicalization on every run.

On the `gate` status: as far as I can tell that is the maintainer-approval gate and clears on approval of the head commit, so nothing for me to action there. Happy to adjust anything further.

carloshvp (commented):
Comment review

Re-reviewed current head `9053f367cb7eec1d1324e94a77966d8c8f21b6b7`. This update resolves the findings from both earlier change-request reviews:

- The fixtures now use the draft-02 `{payload, signature}` envelope and access-decision field names.
- Fixture 03 is signed by the committed second key while retaining the claimed `kid`; verification with the resolved signer key fails deterministically.
- Fixture 04 is validly signed but points to the wrong predecessor; signature verification passes and the independent chain-link check fails.
- `tests/test_acta_fixtures.py` now checks envelope structure and the declared signature, chain, policy-freshness, and session-binding outcomes.

Validated locally:

- `uv run --project /tmp/trace-spec-pr98 --extra dev python -m pytest -q tests/test_acta_fixtures.py` → `18 passed`.
- `uv run --project /tmp/trace-spec-pr98 --extra dev python -m pytest -q tests` → `97 passed`.
- GitHub checks on this head: Python 3.11, Python 3.12, CodeQL, and Python analysis pass. The remaining failing `gate` check is the maintainer-approval gate.

No further blocker from my re-review. This single follow-up supersedes and consolidates my two earlier reviews.

imran-siddique (commented):
Comment, not a block. This needs a maintainer decision (cc @carloshvp) on one governance point.

Technically the crosswalk is accurate and correctly non-normative: the anchors resolve, the field names match the fixtures, the six Ed25519 fixtures self-verify in CI, and the schema is untouched per #97.

The decision before merge: the crosswalk elevates a non-working-group individual I-D (`draft-farley-acta-signed-receipts-02`) and points to a company-owned test-vector repo as "the shared conformance suite," framed as a peer profile. The PR does not disclose that the submitter is the draft's author. Whether trace-spec wants to give an external individual draft peer standing is a maintainer call.

Two tightening items if it proceeds:
- `evidence_hash -> payload_digest` maps to a field present in none of the six fixtures, so it is asserted, not demonstrated.
- Acta has no per-call identifier, so the `linked_call_id` obligation from #97 is only partially satisfiable; call it a gap rather than a clean mapping.

Minor: "exact TRACE terms as used in section 3.3.2" overstates it; 3.3.2 is prose and the field names live only in the informative examples README.

carloshvp (changes requested):
I support an informative Acta interoperability note in TRACE. The current text gives the individual draft and its test-vector repository more standing than the evidence supports. Please make two changes before approval.

1. **Governance framing.** Carry the disclosure from #97 into the PR or crosswalk: the PR author also authors the Acta individual draft and maintains protect-mcp, and ScopeBlind owns the linked vector repository. Describe the material as an experimental, non-normative crosswalk to an external individual draft. Replace “second profile” and “Shared conformance suite” with neutral terms such as “interop example” and “external test vectors maintained by ScopeBlind.” State that inclusion does not endorse the draft or grant it TRACE peer-profile status.

2. **Mapping accuracy.** Mark `evidence_hash -> payload_digest` as an optional mapping that none of these fixtures exercises, or add a fixture that does. State that Acta provides session binding but no native per-call identifier, so the current fixtures do not satisfy `linked_call_id`; a deployment needs a digest or extension. Replace “exact TRACE terms as used in section 3.3.2” because those field names come from the informative fixture README.

The fixture, signature, and chain findings from my prior reviews remain resolved on head `9053f36`. I can approve after this framing pass.

imran-siddique:
@carloshvp is driving this review and I am deferring to his call on it. His 2026-07-27 review set two conditions before approval, the governance framing and the mapping accuracy, and those still stand: the latest commit here predates that review, so the ball is with @tomjwxf.

Flagging one thing for the record rather than as a new demand, since @carloshvp already raised it: the disclosure point matters more than the wording. The PR author also authors the Acta individual draft and maintains protect-mcp, and ScopeBlind owns the linked vector repository. An informative crosswalk to an external individual draft is welcome, and stating those relationships plainly in the document is what keeps it informative rather than promotional.

No action needed from me here. @carloshvp, over to you.

lywinged:
I have been building conformance vectors and completeness checks for the receipt fixtures in this repo, so I ran the same measurement over this set. One thing that is right and worth saying so, and one bug with a fix I have verified end to end.

**Scope, so you can weight this:** I checked the fixture set's internal consistency and the canonicalization. I did **not** review the crosswalk document against draft-farley-acta-signed-receipts-02 — I have not read the draft, so I have no opinion on mapping accuracy, which is most of what @carloshvp's bar in #97 asks for.

## The hand-rolled canonicalizer is correct, and the README should say why

`gen.mjs` rolls its own `canonicalize()`, which is usually where this kind of set goes wrong. It does not. I reimplemented it faithfully and differential-tested against `rfc8785` over ASCII, non-ASCII values, supplementary-plane values, supplementary-plane **keys**, nested arrays, large integers, floats, exponent-form floats, empty containers and escape sequences — identical bytes in every case, and identical digests on the actual fixtures.

That is by construction rather than luck: JCS is defined in terms of ECMAScript, so `Object.keys().sort()` (UTF-16 code units) and `JSON.stringify` on primitives (Number-to-String, raw UTF-8) land on it. One sentence in the README saying so would save the next reader who spots a hand-rolled canonicalizer in a repo that mandates RFC 8785 from "fixing" it into divergence.

## Every fixture carries the same `policy_digest`, so `05` does not isolate what it claims

```
current_policy_digest              sha256:bda8d8f6...
01-valid-accepted.json             sha256:b5af974a...  != current   declared=pass
02-valid-denied.json               sha256:b5af974a...  != current   declared=pass
03-signature-key-mismatch.json     sha256:b5af974a...  != current   declared=n/a
04-broken-chain.json               sha256:b5af974a...  != current   declared=pass
05-stale-policy-digest.json        sha256:b5af974a...  != current   declared=fail
06-session-binding-mismatch.json   sha256:b5af974a...  != current   declared=pass

distinct policy_digest values across all six fixtures: 1
```

By the comparison the README states for `05` — "`policy_digest` names a policy bundle that is no longer the one in force (`current_policy_digest` in `expected.json`)" — `01`, `02`, `04` and `06` fail it too while being declared `pass`. `issued_at` does not rescue them: there is no validity window in the payload and the stated check is a digest comparison.

This lands exactly on what the set is for. The README says `04`, `05` and `06` "make one point three ways", but on the freshness axis `05` is indistinguishable from every other fixture — and a second implementation running `expected.json` as the portable contract disagrees with it on four of six.

**The cause is `decisionPayload()`**, which hardcodes `policy_digest: POLICY_V1` with no override, so `r05` gets V1 like everything else — even though the comment above it says V1 is what makes `05` stale. The intent is in the comment; the code does not carry it out.

## Fix, verified

Before changing anything I ran `gen.mjs` unmodified and confirmed it reproduces all seven committed files **byte for byte**, so the regenerated output below can be trusted.

```diff
-function decisionPayload({ tool, decision, reason, sessionId, issuedAt, prevHash }) {
+function decisionPayload({ tool, decision, reason, sessionId, issuedAt, prevHash, policyDigest = POLICY_V2 }) {
   const p = {
     type: 'protectmcp:decision',
     tool_name: tool,
     decision,
-    policy_digest: POLICY_V1,
+    policy_digest: policyDigest,

 const r05 = envelope(decisionPayload({
   tool: 'run_shell', decision: 'allow', sessionId: SESSION,
   issuedAt: '2026-07-08T09:00:12.000Z',
+  policyDigest: POLICY_V1, // the only fixture carrying the superseded policy
 }), SIGNER_PRIV);
```

The cascade handles itself, since `envelopeHash(r01)` is computed rather than pinned: signatures change, the chain head moves from `99c36f17…` to `ddf7efb0…`, and `02` and `04` relink automatically — `04`'s deliberately wrong pointer stays deliberately wrong. After regenerating, `policy_digest` has two distinct values, `05` is the only fixture failing freshness, and the existing suite passes unchanged.

## Why CI did not catch it, and the test that would have

`expected.json` declares four axes and the README says the results are re-verified in CI. Two are: `signature` for all six, `chain` for the two entries in `EXPECTED["chain"]`. The other two are read by no test. `test_stale_policy_digest_is_valid_but_stale` and `test_session_binding_mismatch_is_valid_but_misbound` assert against fixture contents directly, are hardcoded to one fixture each, and assert **inequality** — which passes whether one fixture is stale or all six are. `gen.mjs`'s own self-check has the same asymmetry: it verifies `05` fails freshness and never verifies that the others pass.

Measured rather than inferred: flipping each of the 24 declared cells in `expected.json`, 16 flips change no test outcome, and every `policy_freshness` and `session_binding` declaration is among them. Changing `expected_session_id` or `current_policy_digest` outright to unrelated values also fails nothing.

```python
@pytest.mark.parametrize("name", FIXTURES)
@pytest.mark.parametrize("axis,field,reference", [
    ("policy_freshness", "policy_digest", "current_policy_digest"),
    ("session_binding", "session_id", "expected_session_id"),
])
def test_comparison_result_matches_expected(name, axis, field, reference):
    """The two comparison axes, checked for every fixture rather than only the failures."""
    expected = EXPECTED["results"][name][axis]
    if expected == "n/a":
        assert not _signature_verifies(_load(name)), (
            f"{name}: {axis} is n/a, which is only justified when the signature fails"
        )
        return
    actual = "pass" if _load(name)["payload"][field] == EXPECTED[reference] else "fail"
    assert actual == expected, f"{name}: {axis} is {actual}, expected {expected}"
```

On the current fixtures this fails exactly the four above; with the generator fix, the suite is 30 passed. `session_binding` has the identical structural gap but is currently **correct** — only `06` carries a different `session_id` — so that column is a latent hole rather than a live inconsistency, and the same test closes both.

Happy to open this as a PR against your branch, or leave it with you, whichever is less friction.

**What this does not establish:** I only mutated the declared expectations, so a check passing for the wrong reason is outside what this sees, and I have not looked at the crosswalk document. The set is otherwise the most carefully separated one in this repo — committing the second signing key rather than using random bytes for `03`, and giving the chain check its own fixture where the signature deliberately passes, are both right and neither is common.

imran-siddique (approved):
Meets the bar set in #97: Acta stays an external, independently specified format, carried by digest and issuer metadata rather than embedded as a second receipt schema. Non-normative crosswalk, no change to schema/trace-claim.json, registered in mkdocs the same way as the OWASP crosswalk.

The fixtures are the strongest part. Real Ed25519 signatures with the mismatched key committed, and 04/05/06 making one point three ways: a valid signature over a wrong chain pointer, a stale policy basis, or another session's receipt all pass the cryptographic check and fail only the check a verifier might have skipped. Thanks for the care.

Issue #67, opened 2026-06-26
[pre-v1.0] Key revocation: TraceRevocation claim + entry-ID anchor


imran-siddique:
Tracking issue for the key-revocation design discussed in #46. Maintainer-driven editorial work for v1.0, not a contributor schema merge.

## Decision

Anchor revocation to **SCITT log entry ordering**, not to `iat`.

A compromised record-signing key also controls the `iat` field it signs, so any rule of the form "reject if key revoked and `iat > revoked_time`" is defeated by backdating. We therefore anchor to the Merkle-bound log entry ID instead.

### `TraceRevocation/1.0` claim type (new §3.2.4)

```json
{
  "type": "TraceRevocation/1.0",
  "compromised_key_id": "<key fingerprint>",
  "last_valid_entry_id": "<SCITT log entry ID>",
  "revoked_after_entry": "<next entry ID>",
  "reason": "key compromise",
  "revocation_key_id": "<org recovery key fingerprint>",
  "sig": { "alg": "ed25519", "value": "..." }
}
```

### Verifier rule

> A record signed by a revoked key is valid if and only if its SCITT inclusion entry ID is <= `last_valid_entry_id` in the applicable revocation statement.

- Entry IDs are monotonic and cryptographically bound to the Merkle structure: no clock trust, no notary trust beyond the existing SCITT profile.
- **Fallback:** for records with no usable SCITT receipt, binary revocation (reject all records from the key). No external anchor means no reliable way to distinguish pre- from post-compromise.

### Distribution (offline-verifiable)

- Revocation statements are anchored in the same SCITT log as the records they govern (preserves the no-callback property of §3.3).
- Verifiers cache a signed revocation **bundle** with a `valid_until` field, same max-age model as §3.2.2. A verifier offline can assert "verified against revocation bundle valid at T".
- This replaces the original "MUST check a well-known endpoint" proposal, which would break offline verification.

### Signing-key independence (MUST)

> A revocation statement for key K MUST be signed by a key at a higher level in the §3.2.1 key hierarchy than K, or by a designated org/recovery key whose compromise domain is independent of K.

## Scope of the editorial change

1. New **§3.2.4** (or §3.3.x): `TraceRevocation` claim type + entry-ID revocation rule + binary fallback.
2. `docs/verification.md`: upgrade the receipt framing so inclusion entry ID is a normative ordering/anchor source (MUST), not only an inclusion/immutability proof.
[...]

PR #113, opened 2026-07-30
fix(verify): enforce key revocation at verification time (#76)

### LINKED ISSUE #76: security: enforce revocation/transparency at verify time
Tracked follow-up from the June 2026 org-wide security hardening review. Wave 1/2 fixes are merged and published; this is remaining hardening.

verify_record checks signature + freshness but does not consult any revocation/status or SCITT transparency state, so a revoked/compromised key keeps verifying. 

Scope: add a revocation/status check to the verify path (caller-provided revocation store or CRL/STH lookup), and document that pure offline verification cannot prove non-revocation. Fail closed when a record references a revoked key.

zohebk8s:
## What this changes

Closes #76.

`verify_record()` checked signature and freshness but never consulted revocation or transparency state, so a record signed by a revoked or compromised key kept verifying. This is a gap against the spec's own normative text, not a new requirement: §3.2.1 already says *"Verifiers MUST consult current revocation status at verification time"*.

A signature stays cryptographically valid forever and nothing inside a record can retract the key that made it, so the verifier is the only place withdrawal of trust can be applied.

### The revocation store

`verify_record(..., revocation=...)` accepts either form:

```python
# A revocation list the caller already holds.
verify_record(record, trusted_jwk, revocation={"kPrK_qmxVWaYVA9wwBF6Iuo3vVzz7TxHCTwXBygrS4k"})

# Or a live CRL / status endpoint / SCITT lookup.
verify_record(record, trusted_jwk, revocation=lambda key_id: crl.is_revoked(key_id))
```

| Outcome | Result |
|---|---|
| Key listed as revoked | Rejected (`ValueError`) |
| Store raises (endpoint down, timeout) | Rejected; an unavailable source is not evidence a key is unrevoked |
| Key absent from the store | Verification continues |
| No `revocation` passed | Check skipped; offline verification, unchanged |

Failing closed on an unreachable store is the part that matters: if a lookup failure fell through to "verified", an attacker holding a revoked key would only need to make the status endpoint unreachable.

### Two design points worth review

**The check reads the trusted key, not `record["cnf"]["jwk"]`.** The embedded key is attacker-controlled until the signature verifies, so keying the lookup on it would let a revoked issuer present an unlisted thumbprint and pass. There is a test pinning this.

**Identifiers are the RFC 7638 JWK Thumbprint or `kid`.** Thumbprints work when an issuer publishes no `kid`, and are stable across JWKs differing in optional members. `jwk_thumbprint()` is exported so a caller can key a list on the same identifier the verifier derives; it is pinned against the RFC 8037 Appendix A.3 known-answer vector.

### Scope

Deliberately **not** in this PR:

- No normative text, schema, or record field changed. This implements an existing MUST in the library, so it is not a spec change and needs no comment period.
- `revocation` defaults to `None` rather than being required. Making it mandatory is a breaking API change for every consumer (cMCP, ca2a) and is better decided alongside #67, which designs the `TraceRevocation` claim and the entry-ID anchor. This PR gives the verify path the hook; #67 can decide what feeds it.
- Transparency/SCITT state is reachable through the callable form, but no SCITT client is added here.

Documented rather than left implied: `docs/verification.md` opens with "independently verifiable offline", which is true and is exactly why non-revocation is the one thing it cannot establish. That page and `LIMITATIONS.md` now say so directly.

## Type of change

- [ ] Schema change
- [ ] Example addition

None of the listed categories fit, so I have left them unticked rather than claim one inaccurately. This is a library and docs change that implements an existing normative MUST. No spec text, schema, or record field is modified. Happy to tick whichever box you consider correct.

## Spec section

3.2.1 (Signing and key management), implemented rather than modified. No spec text changed.

## Checklist

- [x] `CHANGELOG.md` updated
- [ ] Breaking changes marked with `<!-- CHANGED: #NNN — description -->` in spec text (not applicable, no breaking change)
- [x] Backward compatibility statement included: additive keyword argument defaulting to `None`, so offline verification is byte-for-byte unchanged

## Verification

- 105 tests pass, 14 new. Every new line covered (`sign.py` misses are pre-existing).
- `ruff check src tests` and `mypy src/agentrust_trace` (strict) clean.
- Backward compatibility checked against a downstream consumer: the cMCP suite passes unchanged with the patched package installed, 893 passed and 8 skipped, before and after.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,16 @@ Format: [Semantic Versioning](https://semver.org/). Spec versions follow `MAJOR.
 
 ## [Unreleased]
 
+### Added
+
+- **`verify_record(..., revocation=...)` enforces key revocation at verification time (#76).** §3.2.1 has always required that "Verifiers MUST consult current revocation status at verification time", but `verify_record()` checked only signature and freshness, so a record signed by a revoked or compromised key kept verifying. The new `revocation` parameter accepts either a container of revoked key identifiers or a callable performing a live CRL, status-endpoint, or SCITT lookup. A listed key is rejected, and a store that cannot answer is also rejected: an unavailable revocation source is not evidence that a key is unrevoked.
+
+  Keys are identified by RFC 7638 JWK Thumbprint or `kid`. The check reads the *trusted* key rather than `record["cnf"]["jwk"]`, which is attacker-controlled until the signature verifies.
+
+  Additive and backward compatible: `revocation` defaults to `None`, which leaves verification purely offline and unchanged. That mode cannot prove non-revocation, now stated in `LIMITATIONS.md` and `docs/verification.md`. No normative text, schema, or record field changed.
+
+- **`jwk_thumbprint(jwk)`**: RFC 7638 JWK Thumbprint (RFC 8037 §2 for OKP), exported so callers can key a revocation list on the same identifier the verifier derives.
+
 ## [0.5.1] — 2026-07-28
 
 ### Fixed
--- a/LIMITATIONS.md
+++ b/LIMITATIONS.md
@@ -25,6 +25,9 @@ Hardware attestation proves the TRACE signing key and policy engine were measure
 **Revocation of the signing key after issuance**
 If the TRACE signing key is compromised after records are issued, existing records remain cryptographically valid. Key monitoring, rapid revocation, and transparency log integration are the required controls — TRACE provides the anchoring mechanism but cannot detect compromise itself.
 
+**Pure offline verification cannot prove non-revocation**
+Signature validity is permanent; trust is not. Nothing inside a record can retract the key that signed it, so a record signed by a since-revoked key verifies offline forever. Spec §3.2.1 accordingly requires verifiers to consult current revocation status at verification time, which is by definition an online step. `verify_record()` takes a `revocation` store, either a container of revoked identifiers or a callable performing a live CRL, status-endpoint, or SCITT lookup. It rejects a listed key, and fails closed when the store cannot answer. Without that store, verification is offline and its result means "this record was validly signed by this key", not "this key is still trusted".
+
 ## What Level 0 does not provide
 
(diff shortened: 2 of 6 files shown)


zohebk8s:
Fork CI is green on this exact commit (283383d), both matrix legs:
https://github.com/zohebk8s/trace-spec/actions/runs/30584031843

test (3.11) and test (3.12): lint, type check, and tests all pass.

imran-siddique (approved):
Reviewed the diff carefully, including against the individual-contributor spec policy. This is **not** a normative change: `spec/trace-v0.2.md` and `schema/` are untouched, and the changed files are the reference implementation plus non-normative docs. It implements normative text that §3.2.1 already carries rather than adding a requirement, so the policy that reserves spec changes to known orgs does not apply here.

On the substance, the security-critical decisions are all the right way round:

- The check is keyed on the **trusted** key, never `record["cnf"]["jwk"]`. That is the one detail that decides whether this is a real control or theatre, since the embedded key is attacker-controlled until the signature verifies, and keying on it would let a revoked issuer present an unlisted thumbprint. The `allow_embedded_key=True` case is called out rather than glossed.
- A store that cannot answer is a rejection, not a pass. An unreachable CRL is not evidence a key is unrevoked.
- `revocation=None` stays the default, so offline verification is unchanged and the feature is additive. `LIMITATIONS.md` and `docs/verification.md` now say plainly that offline mode proves "validly signed by this key", not "this key is still trusted". That honesty is the part I care most about.
- `jwk_thumbprint` hashes only the RFC 7638 required members per `kty`, so it is stable across JWKs differing in `kid`/`alg`/`use`. Using JCS for the serialization is fine given member names are ASCII, and the docstring already notes why that is byte-identical to RFC 7638 §3.3 ordering. Omitting `oct` is correct for a module that only handles public confirmation keys.
- Matching on thumbprint or `kid` covers issuers that publish neither consistently.

Placement before the signature check is right: the trusted key's status does not depend on the record, so there is nothing to gain by verifying first.

Only failing check is the maintainer-hold gate.

PR #122, opened 2026-08-05
test: conformance vectors for seven unexercised receipt rules

lywinged:
## What

Seven conformance fixtures, one per receipt rule that the verifier applies and that no
fixture previously exercised.

| Rule | What an implementation can omit today and still pass |
|---|---|
| `action_ref_invalid` | Recomputing the action reference instead of trusting the declared value |
| `call_id_mismatch` | Checking the receipt is bound to *this* call |
| `session_id_mismatch` | Checking it is bound to *this* session |
| `evidence_hash_mismatch` | Recomputing the evidence digest |
| `issuer_key_untrusted` | Checking the signing key against a pinned set at all |
| `receipt_from_future` | Rejecting a receipt issued after the verification time |
| `decision_invalid` | Refusing to read an unknown verb as accept or reject |

## Why these two matter beyond tidiness

Without `issuer_key_untrusted`, a receipt authenticates itself. A signature verifies
against whatever key it names, and only a pinned set decides whether that key was ever
entitled to speak.

Without `evidence_hash_mismatch`, the signature covers a digest whose document can be
swapped. The receipt signs `evidence_hash`, not the evidence body, so only recomputation
catches a substituted body.

The other five are ordinary binding and freshness checks, and the same argument applies
in weaker form: a suite that never exercises a rule certifies implementations that skip
it.

## How they were found

By walking the verifier's source for every failure code it can emit and comparing that
against the codes the fixtures expect. The seven above were emitted by the verifier and
expected by nothing.

I am happy to contribute the check itself as a follow-up if it is wanted — it recovers
the rule inventory from source with `ast` rather than from a hand-maintained list, so a
rule added without a fixture fails rather than passing quietly. Keeping it out of this PR
so the fixtures can be judged on their own.

## Shape

One fixture per rule, each triggering exactly that rule and nothing else, so a failure
names the check that broke rather than a set of them.

They pin their own deterministic test key. The private half of the key behind `01`–`09`
is not published, and each fixture already carries its own `trusted_issuer_keys`, so a
separate key changes nothing about how the set is verified.
`gen_rule_coverage_vectors.py` regenerates the set byte-for-byte; only public JWKs appear
in the files. Worth stating plainly that the keys are deliberately deterministic test
keys rather than leaving a reviewer to notice the pattern.

## Scope

Additive. No existing fixture changes, no verifier changes, no schema or spec changes.
The only edits to existing files are the filename list in `test_fixture_set_is_complete`
and a table in the examples README.

Under `GOVERNANCE.md` this is a conformance-test addition: no normative text, so no
sponsor and no comment window. Please tell me if you read it differently.

## Checks

- 112 tests pass (105 before, plus 7)
- `ruff check src tests` and `mypy src/agentrust_trace` clean
- Verified on Python 3.11, 3.12, 3.13 and 3.14
- DCO signed

### DIFF
--- a/docs/verification.md
+++ b/docs/verification.md
@@ -156,14 +156,15 @@ Keep the verification results separate:
 | Action issuance evidence | Canonical action digest, receipt signature, trusted issuer key, session or call binding, chain order | Successful physical completion |
 | Outcome evidence | Controller or monitor decision carried by the receipt payload | Functional-safety certification unless the issuer and profile explicitly claim it |
 
-For action receipts, a verifier should distinguish four common outcomes:
+For action receipts, a verifier should distinguish five common outcomes:
 
 | Outcome | Meaning |
 |---|---|
 | `receipt_valid_accepted` | The receipt is well-formed, trusted, bound to the call, and reports acceptance. |
 | `receipt_valid_rejected` | The receipt is well-formed, trusted, bound to the call, and reports controller or policy rejection. This is valid negative evidence. |
 | `receipt_missing_required` | The profile required a receipt, but none was present for the consequential action. |
-| `receipt_invalid` | The receipt is present but fails signature, digest, issuer, freshness, ordering, or call-binding checks. |
+| `receipt_invalid` | The receipt is present but fails signature, digest, freshness, ordering, or call-binding checks against a key the verifier holds. |
+| `receipt_unverified` | The receipt names an issuer key the verifier has not pinned, and nothing else failed. Per section 3.3.1 of the spec this is unverified, not invalid: the receipt confers no trust and proves no wrongdoing, surfaced with an advisory rather than a failure. |
 
 The key boundary is that a valid rejection is not malformed evidence. It is
 evidence that the downstream authority declined the action. A valid acceptance
--- a/examples/action-receipts/README.md
+++ b/examples/action-receipts/README.md
@@ -93,14 +93,18 @@ three operations:
 | `11-call-id-mismatch.json` | `receipt_invalid` | unknown | An authentic receipt bound to a different call. |
 | `12-session-id-mismatch.json` | `receipt_invalid` | unknown | An authentic receipt from a different session. |
 | `13-evidence-hash-mismatch.json` | `receipt_invalid` | unknown | The receipt is authentic but the detached evidence was swapped after signing. |
-| `14-receipt-issuer-key-untrusted.json` | `receipt_invalid` | unknown | The issuer key is not in the verifier's pinned set. |
+| `14-receipt-issuer-key-unknown.json` | `receipt_unverified` with advisory | unknown | The issuer key is not in the verifier's pinned set. Unverifiable is not invalid (spec §3.3.1): no trust is conferred and no forgery is proven, surfaced as an `issuer_key_unknown` advisory. |
 | `15-receipt-from-future.json` | `receipt_invalid` | unknown | Issued after the verification time, so an upper bound on age never rejects it. |
 | `16-decision-not-in-enum.json` | `receipt_invalid` | unknown | An unrecognised decision verb, which must not read as accept or reject. |
 
 Fixtures `10`–`16` each pin down one rule that the verifier applies and that no fixture
 previously exercised. Every one was a check a conforming implementation could have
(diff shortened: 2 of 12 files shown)


lywinged:
Flagging something against my own PR, found while reading §3.3 rather than the verifier.

**`22-receipt-issuer-key-untrusted.json` pins behaviour that diverges from §3.3.1.**

The spec says, of a receipt whose issuer key the verifier does not hold:

> A receipt whose issuer key is unknown to the verifier is **unverified, not invalid**. The
> Trust Record's gateway-produced evidence (signature, audit-chain hash, policy hash, TEE
> measurement) is unaffected. Verifiers SHOULD surface an advisory status (e.g.
> `external_evidence_unverified`) rather than silently ignoring the receipt.

The fixture verifier appends `issuer_key_untrusted` to `failures`, which resolves to
`receipt_invalid`. My fixture records that as the expected outcome:

```json
"status": "receipt_invalid",
"failures": ["issuer_key_untrusted"],
"warnings": []
```

Advisory and invalid are different claims. §3.3.1 draws the line deliberately — an
unconfigured key is a gap in the verifier's configuration, not a defect in the evidence,
and the distinction is the one the section spends a paragraph on.

**The divergence is not mine.** `issuer_key_untrusted` has been a failure since before this
branch; the fixture only records what the verifier already does. But that is the part worth
raising: **a conformance vector that pins a divergence makes it harder to correct, because
it converts an accident into an expectation.** Left as is, changing the verifier to match
§3.3.1 would show up as a failing conformance test, which is backwards.

§3.3.2 is informative and this fixture set declares its own test profile, so nothing here
is strictly violated. It diverges without saying so, which is the smaller version of the
same problem.

## What I would suggest, though it is your call

Three options, in the order I would rank them:

1. **The verifier is wrong.** Emit `issuer_key_untrusted` as a warning with a status that
   is not `receipt_invalid`, matching §3.3.1. That is a behaviour change rather than a test
   addition, so it does not belong in this PR — but the fixture should then be written
   against the corrected behaviour, not merged first and rewritten after.
2. **The spec's rule does not reach this profile**, because §3.3.2 is informative and the
   profile may be stricter on purpose. Then the fixture is right and the divergence is
   worth one sentence in `examples/README.md`, so the next reader does not have to
   rediscover it.
3. Drop `22` from this PR and land the other six, leaving that rule unguarded until the
   question is settled.

I have no view on which of the first two is correct — it is a question about intent, not
about code. Happy to do any of the three; tell me which and I will push it.

The other six fixtures are unaffected: none of them touches a rule §3.3.1 or §3.3.2 speaks
to differently from the verifier.

lywinged:
Pushed the correction for the flag above, as option 1 (`ff6ad93`): the verifier now surfaces an unknown issuer key as an `issuer_key_unknown` advisory in `warnings`, with a fifth outcome `receipt_unverified`, and fixture 14 is regenerated against that behaviour — renamed to `14-receipt-issuer-key-unknown.json`, since the old filename carried the same conflation. The other six fixtures are byte-identical, and the outcome table in `docs/verification.md` gains the new row.

I said above that the behaviour change was not this PR's to make, and the call is still yours. But every reading I could construct came out the same way — an unpinned key is an inability to check, not evidence of forgery, and §3.3.1 spends its paragraph on exactly that line — and of the three options, merging a vector that pins the divergence looked strictly worst. So the branch now carries the spec-aligned behaviour as the default rather than the divergence. If you would rather take option 2 or 3, reverting is one commit: say so and I will push it.

lywinged:
One more note, to make this easier to dispose of rather than to add to it.

The six vectors other than `14` touch nothing §3.3.1 or §3.3.2 speaks to differently from the verifier. They were reviewable before I raised the issuer-key question and they still are. If it is useful, **say the word and I will drop `14` and the behaviour change from this branch entirely**, leaving six fixtures that are pure test additions with no decision attached. The issuer-key question then lives on as its own issue, where it can take as long as it needs without holding anything up.

I would rather that than have a question I raised become the reason six unrelated vectors sit unmerged.

imran-siddique (approved):
Seven rules the verifier applies that no fixture exercised. An implementation could skip recomputing the action reference or the evidence digest, or never pin the issuer key, and still pass the suite. Closing that is worth doing before the spec moves and other-language verifiers start certifying against it.

Issue #116, opened 2026-08-02
verification.md: format-versioning and verifier-compatibility obligations (evidence outlives verifiers)


lywinged:
**Spec gap.** Compliance evidence is verified years after issuance — the artifact routinely outlives the verifier build — and the profile/enum surface is already evolving pre-1.0 (#114 proposes a new profile URI precisely to avoid breaking existing verifiers). But `verification.md` currently has no normative statement of what a verifier must do about version skew. Without one, every future revision risks silently orphaning existing evidence — or worse, verifiers "helpfully" accepting artifacts under semantics they don't actually implement.

Production context: in our per-action receipt format the hash-computation order is load-bearing — any change orphans every previously issued artifact. What held up operationally: a mandatory format-version field, plus a verifier obligated to check every version it claims to support and to say *which version it verified under* in its output.

**Proposed obligations for `verification.md`:**

1. Every record/receipt carries an explicit format/profile version (the profile-URI direction in #114 already implies this — this makes it a *verifier* obligation, not just an emitter convention).
2. A conformant verifier declares the set of versions it supports and MUST refuse versions outside that set — refuse, not best-effort.
3. The verification statement includes the version under which verification ran.
4. Any downgrade/fallback behavior MUST be disclosed in the statement; silent fallback is non-conformant.

**Conformance vectors for trace-tests:**
- unknown-version artifact → must refuse (fail-open here is the dangerous default)
- known-version artifact → verified under correct semantics, version echoed in the statement
- downgrade-with-disclosure → conformant
- silent downgrade → must fail conformance

Happy to draft the section plus the vectors — this composes with the verifier-obligations ADR discussed in #66 and the offline verification statement discussed in #76.

imran-siddique:
Thanks for this, and for the detail in the write-up.

A process note on format-versioning and verifier-compatibility obligations when evidence outlives its verifiers.

Normative spec text, meaning anything carrying an uppercase RFC 2119 keyword, needs an organizational sponsor accountable for the requirement. That is a provenance rule about who stands behind a requirement others implement against, not a judgment on the proposal or on you. Anyone may propose a normative change, and where a proposal is accepted without a sponsor, a maintainer carries the PR.

So this stays open. I have labeled it `policy: take-internal`, which means the substance goes through review here and a maintainer carries any resulting spec change. This issue gets updated with the outcome either way.

[...]

PR #125, opened 2026-08-07
fix(verify): enforce the profile cutover at verification time

lywinged:
## What

`verify_record()` never reads `eat_profile`. A record carrying the superseded v0.1 identifier, a future version, a foreign tag, or no profile at all verifies exactly as a v0.2 record, provided its signature checks out.

The unreleased changelog already declares the intended behaviour — a v0.2 verifier *"requires the new URI and rejects the old one; it does not accept both"* — and `spec/trace-v0.2.md` §2 makes the same requirement normative (*"Cutover, not coexistence"*). This PR makes the reference implementation carry it out: the profile is checked first, before any cryptographic work, and anything other than the v0.2 identifier raises `ValueError`. The v0.1 identifier gets a message that says why it is rejected. A missing profile is refused for the same reason a wrong one is: a verifier cannot supply the semantics by assumption, and a valid signature over semantics the build does not implement is not evidence.

Same shape as the revocation fix (#76): an already-merged spec requirement the reference implementation did not enforce, needing no new normative text.

## Scope, deliberately minimal

`TRACE_PROFILE_V0_2` is exported so callers can pin the constant the verifier accepts; the v0.1 tag is deliberately **not** exported. There is no configurable accepted-profiles set and no verification-statement return — #116 proposes designs in that space, and this PR takes no position on them. It only carries out the MUST that is already merged.

## Verification

- 110 tests pass (105 before, plus 5 — including one proving the refusal does not depend on cryptographic work: a wrong-profile record with a garbage signature is refused for its profile, not its signature)
- Neutralizing the check fails 4 of the 5
- `ruff check src tests` and `mypy src/agentrust_trace` clean
- Verified on Python 3.11 and 3.14
- `docs/verification.md` step 4 (the manual profile assert) now notes the library does this itself
- No normative text, schema, or record field changed
- DCO signed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,10 @@ Format: [Semantic Versioning](https://semver.org/). Spec versions follow `MAJOR.
 
 - Other `agentrust.io` URLs moved to `agentrust-io.com`: the registry and verifier hosts in the AGT adapter and the schema `$id`.
 
+### Fixed
+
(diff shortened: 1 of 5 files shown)


imran-siddique (approved):
Correct and well argued. The profile check belongs before any cryptographic work, and the docstring says why that ordering is safe: the only action taken on the unauthenticated value is refusal, and a record that verifies has had its profile covered by the signature.

Enforces what spec/trace-v0.2.md section 2 and the changelog already declared. Landing this before the LF move.

Issue #128, opened 2026-08-07
spec: §3.2.2 states code-point key ordering; RFC 8785 sorts by UTF-16 code unit


lywinged:
**Spec section affected**

§3.2.2, canonical form step 2, first bullet (`spec/trace-v0.2.md` line 212).

**Problem**

The bullet describes the sort order of the algorithm the same sentence mandates, and describes it wrongly:

> 2. Apply RFC 8785 JSON Canonicalization Scheme (JCS) to produce a deterministic byte sequence:
>    - Object keys are sorted in **Unicode code-point order** (ascending).

RFC 8785 §3.2.3 sorts by **UTF-16 code unit**, not by code point. The two orders agree across the BMP and disagree once a key contains a supplementary-plane character, because surrogates occupy D800–DFFF:

```python
# Escapes throughout, so nothing here depends on how your terminal renders it.
# U+1F600 is supplementary-plane (surrogate pair D83D DE00); U+FFFD is high BMP.
>>> keys = ["zk\U0001f600", "zk\uFFFD"]
>>> sorted(keys)                                        # code point (Python str)
['zk\uFFFD', 'zk\U0001f600']                             # 0xFFFD < 0x1F600
>>> sorted(keys, key=lambda s: s.encode("utf-16-be"))   # UTF-16 code unit (RFC 8785)
['zk\U0001f600', 'zk\uFFFD']                             # 0xD83D < 0xFFFD
```

**Nothing is broken today and no record changes.** The reference implementation is correct and knows it is — `_canonical_bytes`'s docstring says "JCS sorts object keys by UTF-16 code unit", and commit `1a3a711` names this divergence as one of the reasons `json.dumps` was replaced. So the code and the specification disagree, in the code's favour.

The exposure is to an implementer working from the document rather than from the Python. §3.2.2 presents that bullet list as the algorithm, and a Go or Rust verifier built to the letter of it sorts by code point, produces a different pre-image for any record with a supplementary-plane key, and fails to verify a valid record with no useful error. `ROADMAP.md` targets Go, Rust and TypeScript verification libraries for v1.0, so the population that would be misled by it does not exist yet but is planned.

Reachable inside a schema-valid record: `cnf.jwk` is the one object in `schema/trace-claim.json` without `additionalProperties: false`, which is correct — RFC 7517 permits additional JWK members — so a member name is the one key the schema does not fix.

**Proposed change**

One word:

```diff
[...]

PR #134, opened 2026-08-08
docs(spec): publish the registry anchor and inclusion-proof format (#111)

### LINKED ISSUE #111: Publish the registry anchor format (inclusion-proof spec) publicly
Raised by @l33tdawg in [agentrust-io/.github discussion #15](https://github.com/orgs/agentrust-io/discussions/15), who suggested trace-spec as the home for it. Opening it here rather than leaving him to file it.

## Problem

To anchor a TRACE record's `transparency` field and verify inclusion independently, a third party needs the Merkle anchor and inclusion-proof format. That format **is** specified: `docs/anchor-format.md` in `agentrust-io/trace-registry` is normative and explicitly written so a conforming verifier can be built from the document alone.

`trace-registry` is private for operational reasons, not because the format is unsettled. The effect is that the one document an external verifier needs is the one they cannot read, which undercuts the "verify without trusting the operator" property the registry exists to provide. An inclusion proof nobody outside can check is not much of a transparency log.

## Proposal

Publish the anchor format as a public document. Options, in rough order of preference:

1. **Move it into `trace-spec`** as a companion spec document, since this is the public spec home and @l33tdawg suggested it. The registry keeps its implementation private; the format it commits to becomes public.
2. Make `trace-registry` public. Cleanest for auditability, largest decision.
3. Publish on the docs site only. Fastest, but splits the spec across two homes.

Whichever way, the reference verifier (`src/trace_verify/_verify.py`, RFC 9162 inclusion check) should be reachable too, since "here is the algorithm, and here is one implementation that is not the definition" is the useful pairing.

## What the document has to state, since these are the parts people get wrong

- **RFC 6962** CT Merkle trees; RFC 9162 inclusion-proof check.
- The anchored unit is the **complete signed claim, signature included**, not a pre-signature payload.
- `canonical_claim_bytes` is **sorted-key JSON**: `json.dumps(claim, sort_keys=True, separators=(",", ":"), ensure_ascii=True).encode("ascii")`, no trailing newline.
- `leaf = SHA-256(0x00 || canonical_claim_bytes)`; interior `node = SHA-256(0x01 || left || right)`.
- Proof shape `{leaf_index, audit_path}` against a checkpoint carrying `merkle_root`, `leaf_count`, `ts`, `producer`, `batch_id`.

**The trap that most deserves prominence:** the registry anchor uses sorted-key JSON, while TRACE record *signing* uses **RFC 8785 (JCS)**. Two canonicalizations at two layers. An implementer who reasonably assumes JCS at the leaf will produce proofs that never verify, with no useful error. That warning belongs in the published document, not only in a discussion thread.

Related: TRACE v0.2 made `transparency` optional below Level 2 (see #109), since an unanchored Level 0/1 record has no receipt to name. Publishing the anchor format is what makes Level 2 independently checkable.

imran-siddique:
Closes #111.

The format was already normative and already written so a conforming verifier could be built from it alone. It lived in `trace-registry`, which is private, so the one document an external verifier needs was the one they could not read. An inclusion proof nobody outside can check is not transparency.

This takes **option 1** from the issue, which @l33tdawg proposed: publish it here, in the public spec home, and leave the registry implementation where it is. That decouples publishing the format from the larger decision about repo visibility.

## `spec/registry-anchor-v1.md`

RFC 6962 trees, RFC 9162 inclusion check, the complete signed claim as the anchored unit, entry fields, proof shape, conformance.

**§0 leads with the trap** rather than burying it, per the issue. TRACE canonicalizes with RFC 8785 (JCS) for *signing* and sorted-key JSON for the *anchor leaf*. The two agree on ASCII-only records carrying integer numbers, which is most records, which is exactly what makes assuming JCS at the leaf dangerous: proofs never verify and the failure has no useful diagnostic. The section names the test that catches it.

**§8 requires the append-only property to be externally checkable, not asserted.** An operator issuing verifiable proofs while publishing nothing an outsider can audit is running a log, not a transparency log.

## Verified, not asserted

The point of the document is that someone can implement from it without reading our code, so I checked that rather than claiming it. An implementation written from the published text alone:

- verifies the registry's real anchored sample (`leaf_count 1`, empty audit path);
- verifies all five proofs of a five-leaf batch built by the registry's own `tools/anchor.py`, which exercises the odd-node **promotion** branch the single-leaf sample never touches (`leaf 4` gets a 1-element path, leaves 0 to 3 get 3);
- rejects a tampered claim against a valid proof.

## Two things found while writing it

**Four documents told readers to send signed records to a domain we do not own.** `docs/integration/agt.md`, `docs/integration/cmcp.md`, `docs/trust-levels.md` and `docs/verification.md` named `registry.agentrust.io`, which resolves to third-party parked addresses. Same defect the v0.2 profile cutover fixed in the identifier, missed in the prose. Moved to `registry.agentrust-io.com`, which is what the SDK's adapters already emit. `spec/trace-v0.1.md` keeps its original values: it is superseded and stands as a record of what was published.

**The anchoring tutorial documented an API that does not exist.** It told readers to POST a signed Trust Record to a SCITT HTTP endpoint at that parked domain and read a `receipt_uri` from the response. There is no such endpoint. Rewritten against the actual mechanism: submit to staging, retrieve the inclusion proof, verify it yourself against the published entry. It also still described `transparency` as a required string, which 0.5.1 changed. The page carries a dated note stating what it used to say, because anyone who built against it should know that nothing they sent was received.

## Not in this PR

Making `trace-registry` public. That is a separate decision with a licensing blocker in front of it, covered in the go-public runbook, and this PR is deliberately independent of it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,20 @@ Format: [Semantic Versioning](https://semver.org/). Spec versions follow `MAJOR.
 
 ## [Unreleased]
 
+### Added
+
+- **`spec/registry-anchor-v1.md`: the registry anchor and inclusion-proof format is now public (#111).** The format was already normative and already written so a conforming verifier could be built from it alone, but it lived in `trace-registry`, which is private. The effect was that the one document an external verifier needs was the one they could not read, and an inclusion proof nobody outside can check is not transparency. It is published here, in the public spec home, as @l33tdawg proposed in the discussion that raised this.
+
+  §0 leads with the trap, because it is the one that costs an implementer a day and gives no useful error: TRACE canonicalizes with **RFC 8785 (JCS)** for *signing* and with **sorted-key JSON** for the *anchor leaf*. The two agree on ASCII-only records with integer numbers, which is most records, which is exactly what makes assuming JCS at the leaf dangerous.
+
+  §8 states conformance, including the requirement that the append-only property be externally checkable rather than asserted. An operator that issues verifiable proofs but publishes nothing an outsider can audit is running a log, not a transparency log.
+
+### Fixed
+
+- **Four documents told readers to send signed records to a domain this project does not own.** `docs/integration/agt.md`, `docs/integration/cmcp.md`, `docs/trust-levels.md` and `docs/verification.md` still named `registry.agentrust.io`, which resolves to third-party parked addresses. This is the same defect the v0.2 profile cutover fixed in the identifier, missed in the prose. Moved to `registry.agentrust-io.com`, which is what the SDK's adapters already emit. The v0.1 spec keeps its original values; it is a superseded document and a record of what was published.
+
+- **The anchoring tutorial documented an API that does not exist.** It instructed readers to POST signed Trust Records to a SCITT HTTP endpoint at the parked domain above and to read a `receipt_uri` from the response. There is no such endpoint. Rewritten against the actual mechanism: submit to staging, retrieve an inclusion proof, and verify that proof yourself against the published entry. It also still described `transparency` as a required string, which 0.5.1 changed. The page carries a dated note saying what it used to say, because anyone who built against it deserves to know nothing they sent was received.
+
 ### Changed
 
 - **The spec, the README and the roadmap named three different standards homes between them.** §6.1 proposed splitting TRACE between CoSAI and the Linux Foundation entity hosting MCP; the README said "Targeting AAIF"; neither is where this is going. TRACE is being formed at the Linux Foundation as its own series, "TRACE Specification, a Series of LF Projects, LLC" (see #127). §6.1 is rewritten, the README line is corrected, and §7 Q1 is marked resolved rather than deleted so a reader tracking it can see how it landed.
--- a/docs/integration/agt.md
+++ b/docs/integration/agt.md
@@ -45,7 +45,7 @@ adapter = TraceAGTAdapter(
     model_id="claude-sonnet-4-6",
     model_version="20251001",
     build_provenance_digest="sha256:e5f6a7b8...",
-    transparency="https://registry.agentrust.io/claim/...",
+    transparency="https://registry.agentrust-io.com/claim/...",
 )
 
 session = AGTSessionResult(
--- a/docs/integration/cmcp.md
+++ b/docs/integration/cmcp.md
@@ -114,7 +114,7 @@ Or let cMCP push it to the transparency registry automatically:
 # cmcp.yaml
 trace:
   emit: true
-  registry: https://registry.agentrust.io
+  registry: https://registry.agentrust-io.com
   scitt_anchor: true
 ```
 
--- a/docs/trust-levels.md
+++ b/docs/trust-levels.md
@@ -81,7 +81,7 @@ Level 2 adds a SCITT transparency log entry to a Level 1 record. The `transparen
 |-------|-------------|
 | `transparency` | Resolvable `https://` URI to a SCITT receipt |
 
-The [SCITT reference implementation](https://github.com/microsoft/scitt-api-emulator) and the [agentrust SCITT registry](https://registry.agentrust.io) are both supported anchors.
+The [SCITT reference implementation](https://github.com/microsoft/scitt-api-emulator) and the [agentrust SCITT registry](https://registry.agentrust-io.com) are both supported anchors.
 
 **Recommended flow:**
 
--- a/docs/tutorials/anchoring-to-the-registry.md
+++ b/docs/tutorials/anchoring-to-the-registry.md
@@ -1,44 +1,43 @@
 # Anchoring a Trust Record to the TRACE registry
 
-After signing a Trust Record, you should anchor it to the TRACE transparency registry. The anchor receipt proves that the record existed at a specific time and has not been altered since — tamper evidence that holds even if the operator who produced the record is later compromised.
+After signing a Trust Record, you can anchor it to the TRACE transparency registry. The anchor proves the record existed at a specific time and has not been altered since, which is tamper evidence that holds even if the operator who produced the record is later compromised.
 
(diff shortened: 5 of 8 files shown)


lywinged:
Reviewed by implementing §1 to §5.1 from the document text alone, since that is the document's own claim, and by checking its two structural assertions rather than reading past them. Both hold. One defect in the package §7 names, one row missing from §0's table, two minor notes.

Escapes throughout below, so nothing here depends on how a terminal renders it.

## Verified

**§3's equivalence claim is true.** "Level by level, promote the odd node unchanged, and this yields the same tree as the RFC 6962 recursive split at the largest power of two" is the kind of sentence that is usually approximately true. I built both constructions independently and compared roots for every batch size from 1 to 200: no mismatch.

**The document is implementable without the code.** §1, §2, §3 and §5.1 written from the text alone, then every leaf of every batch size from 1 to 40 verified against its own root, and a tampered claim rejected against a valid proof. §5.1 is faithful to RFC 9162 §2.1.3.2, including the right-edge promotion shift, which is the part that usually gets dropped.

**§1 matches the shipped verifier exactly.** `trace_verify._verify.canonical_claim_bytes` is the §1 one-liner, character for character.

## `trace-verify` verifies TRACE signatures with the construction §3.2.2 forbids by name

Not in this diff, but §7 is the sentence that makes the package a reference implementation, and §5.2 explicitly hands signature verification to v0.2 §3.3, which the package implements.

`trace_verify._signature.canonical_body_bytes` builds the signature pre-image as:

```python
body = {k: v for k, v in claim.items() if k != "signature"}
json.dumps(body, sort_keys=True, separators=(",", ":"), ensure_ascii=True).encode("ascii")
```

v0.2 §3.2.2 names that exact construction: *"Implementations MUST use an RFC 8785-conformant library. Using `json.dumps(sort_keys=True)` (Python) or equivalent ad-hoc sorting is insufficient."* `agentrust_trace.sign_record` signs the same object -- the claim with `signature` removed -- through `rfc8785.dumps`. Same scheme, different canonicalizer, so the two agree on ASCII and diverge everywhere else:

```
record contents                                   agentrust-trace   trace-verify
ASCII only                                        valid             valid
any string field containing U+00FC                valid             INVALID
any string field containing a CJK character       valid             INVALID
```

Reproduction, against `agentrust-trace` 0.6.0 and `trace-verify` 0.2.0 from PyPI:

```python
from cryptography.hazmat.primitives.asymmetric.ed25519 import Ed25519PrivateKey
from agentrust_trace.sign import sign_record, key_to_jwk
import trace_verify._signature as tvs

key = Ed25519PrivateKey.generate()
record = {
    "eat_profile": "https://trace.agentrust-io.com/profiles/v0.2",
    "sub": "spiffe://trust.example.org/agent/payments",
    "policy_label": "Zahlungsverkehr f\u00fcr M\u00fcnchen",  # escapes; any non-ASCII does it
}
signed = sign_record(record, key)
print(tvs.verify_claim_signature(signed, key_to_jwk(key)))   # False
```

The CLI verifies signatures by default and exits 1 when the check fails, so an external verifier following §7, on a valid record carrying a German policy label or a non-Latin data class, is told the signature is invalid. `verify_claim_against_registry`'s docstring describes it as "the fail-closed check used by the anchoring paths"; if that is accurate, such a record also cannot be anchored, but that path is not visible from outside.

Two readings, and the public material does not distinguish them:

- the registry deliberately signs over a different pre-image than v0.2 §3.2.2, in which case this is a third row for §0's table and the most consequential one, since it is a divergence at the layer §0 states is JCS; or
- it is a defect in the package.

Either way I would not have §7 point an external verifier at it before that is settled, since §7 exists precisely so that people run it.

Worth stating plainly: this is the mistake §0 warns about, in the opposite direction. §0 says do not reuse the signing canonicalizer at the leaf. What is in the shipped code is the leaf canonicalizer reused at the signing layer. The warning is correct and one-directional, and the missing direction is the one that has already happened.

## §0's table is missing a row: key order

§0 lists two divergences between JCS and §1, escaping and non-integer numbers. There is a third. JCS sorts object keys by UTF-16 code unit (RFC 8785 §3.2.3); §1 sorts by Unicode code point. The two differ once a key contains a supplementary-plane character, because surrogates occupy D800 to DFFF.

§1's prose is right, and Python's `sort_keys=True` implements it, because Python compares strings by code point. TypeScript is one of the three verifier languages ROADMAP targets for v1.0, and JavaScript is the counterexample I ran:

```js
// U+1F600 is supplementary-plane (surrogate pair D83D DE00); U+FFFD is high BMP.
Object.keys({"zk\u{1F600}": 1, "zk\uFFFD": 2}).sort()
// -> ["zk\u{1F600}", "zk\uFFFD"]   UTF-16 code unit order, 0xD83D < 0xFFFD
// true code-point order is the reverse, 0xFFFD < 0x1F600
```

So an implementer who reads §1 correctly and writes the idiomatic sort produces different leaf bytes and gets §0's own symptom: a root that differs, with nothing to indicate why.

The test §0 names does not reach it. It catches the failure it was written for. A non-ASCII BMP character sorts identically under both orders and §1 escapes it either way, so those bytes come out identical; only a supplementary-plane character in a key separates them. I checked both cases.

This is the same distinction as #128, still open, and `cnf.jwk` remains the one object in `schema/trace-claim.json` without `additionalProperties: false`, so it is reachable inside a schema-valid record. Suggested: one clause in §1 saying code point and not UTF-16 code unit, and a supplementary-plane key in the test §0 recommends.

## Two minor notes

- `trace-verify`'s PyPI metadata points Homepage, Documentation and Bug Tracker at `github.com/agentrust-io/trace-registry`. All three return 404 to an anonymous reader, which is the audience this document exists to serve.
- The installed distribution is 0.2.0 while `trace_verify.__version__` is 0.1.0, and `--version` prints the latter. Same drift #131 fixed here.

None of the above touches the format itself, which I think is right, and publishing it here is clearly the correct call.

PR #136, opened 2026-08-09
fix(schema): resync the packaged copy, and make the next drift fail

lywinged:
Small one. `validate_json()` loads `src/agentrust_trace/schema/trace-v0.2.json`; the spec, README and CONTRIBUTING all point a reader at `schema/trace-claim.json`. The two had drifted in three places, so the schema someone reads was not the schema their record is checked against — and nothing failed, because both files are individually valid and each is exercised on its own.

## The visible consequence

`subject`. The root file and `models.py` both accept a DID subject; the packaged copy still required `^spiffe://`.

```python
>>> from agentrust_trace import validate_json
>>> rec = json.loads(Path("examples/intel-tdx.json").read_text())
>>> rec["subject"] = "did:web:example.org:agent:payments-processor"
>>> validate_json(rec)
ValidationError: 'did:web:example.org:agent:payments-processor' does not match '^spiffe://'
```

The same record passes `schema/trace-claim.json` and `TrustRecord.model_validate`. So a subject form the specification permits was rejected by the reference validator, and the two places a reader would check to find out why both said it was fine.

## What is resynced

| | Was | Now |
|---|---|---|
| `subject.pattern` / `.description` | packaged: `^spiffe://` | `^(spiffe://\|did:)`, matching the root file and `models.py` |
| `build_provenance.slsa_level.description` | packaged had lost the sentence explaining Level 0 | restored; Level 0 is what `minimum: 0` exists for |
| `$id` | root file: `trace-v0.1.json` | `trace-v0.2.json`, taken from the packaged copy |

The `$id` goes the other way from the other two, because neither file was uniformly right: the root file's `$id` still named v0.1 on a schema whose `eat_profile` const is v0.2. 0.5.1 moved that URL's domain and left its version segment.

## The guard

`tests/test_validate.py` now compares the two as parsed JSON on every run, plus a regression test that both subject forms pass `iter_errors`. Deleting either half of the schema change fails both tests; I checked that rather than assuming it.

Compared as parsed JSON rather than byte for byte on purpose: the two files differ in line endings by long-standing accident — the root file is CRLF, the packaged one LF — and that changes nothing about how either validates a record. A byte comparison would fail today for a reason that does not matter and would push someone toward reformatting a 428-line file.

## Also in here

One changelog entry moved. The `verify_record()` profile-cutover enforcement shipped in 0.6.0, but I filed its entry under 0.5.1 in #125, next to the cutover declaration it implements. That left 0.5.1 with two `### Fixed` blocks and 0.6.0 with no entry for its only behaviour change. Moved to 0.6.0 with its two internal cross-references corrected, since they pointed at neighbours it no longer has. My mistake, and this change edits the same file, so it seemed better than a second PR for one paragraph.

## Checks

205 passed, 1 skipped. `ruff check src tests` and `mypy src/agentrust_trace` clean. No normative text, no schema semantics beyond the three resynced fields, no record field added or removed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,14 @@ Format: [Semantic Versioning](https://semver.org/). Spec versions follow `MAJOR.
 
 ## [Unreleased]
 
+### Fixed
+
+- **The packaged schema had drifted from the normative one, and a DID subject was the casualty.** `validate_json()` loads `src/agentrust_trace/schema/trace-v0.2.json`, while the spec, README and CONTRIBUTING all point a reader at `schema/trace-claim.json`. The two disagreed in three places, so the schema someone reads was not the schema their record was checked against. The visible consequence: the root file and `models.py` both accept a `did:` subject, the packaged copy still required `^spiffe://`, so `agentrust_trace.validate_json()` rejected a subject form the specification permits. Also resynced: the `slsa_level` description, and the root file's `$id`, which still said `trace-v0.1.json` on a schema whose `eat_profile` const is v0.2.
+
+  `tests/test_validate.py` now compares the two as parsed JSON on every run, so the next drift fails instead of shipping. Compared parsed rather than byte for byte because the two files differ in line endings by long-standing accident, which changes nothing about how either validates a record.
+
+- **A 0.6.0 changelog entry was filed under 0.5.1.** The `verify_record()` profile-cutover enforcement shipped in 0.6.0; its entry landed in the 0.5.1 section, next to the cutover declaration it implements, which left 0.5.1 with two `### Fixed` blocks and 0.6.0 with no entry for a behaviour change. Moved, with its two internal cross-references corrected to match where it now sits.
+
 ## [0.7.0] — 2026-08-08
 
 ### Added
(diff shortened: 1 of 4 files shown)


imran-siddique (approved):
Verified rather than taken on trust: rebased onto current `main` (which now carries the `origin` block and its `if`/`then` cross-field rule) and ran the suite. 205 passed, including your new drift test, so the two schema copies are genuinely in sync after both changes land.

The `subject` finding is the real one. The root file and `models.py` both accept a DID, the packaged copy required `^spiffe://`, and `validate_json()` loads the packaged copy — so the SDK rejected a subject form the spec permits, and nothing failed because each file is individually valid. That is the same shape of defect as two schema copies drifting anywhere else: no test compared them, so the drift was invisible.

Comparing parsed JSON rather than bytes is the right call given the CRLF/LF difference, and saying so in the docstring means nobody 'fixes' it back to a byte comparison later.

The misfiled 0.6.0 changelog entry is a good catch alongside it.

lywinged:
One residue from this, which merging cannot reach: the backfilled release notes for v0.5.1 and v0.6.0 were generated from `CHANGELOG.md` before this change, so they still carry the old placement. v0.5.1's body has the `verify_record()` cutover paragraph; v0.6.0's does not. A release body is a snapshot rather than a view of the file, so the fix here does not propagate to either page.

Impact is small: 0.5.1 is superseded twice over, and the wrong answer only reaches someone reading that page to find out which version carries the cutover enforcement. Recording it rather than proposing a schedule -- it was my misfiling in #125, and I have no way to edit a release body.

If it is worth doing, it is one paragraph moved between two pages: the one beginning "`verify_record()` now enforces the profile cutover", which on `main` now sits under `[0.6.0]` with its two cross-references corrected.

PR #137, opened 2026-08-09
fix(fixtures): give the Acta set a policy digest that is actually current

lywinged:
Follows up the second half of my review on #98, which nobody had a chance to answer before it merged. Small, and it is fixture data plus the checks over it: no spec text, no schema, no record field.

cc @tomjwxf, since this touches your set.

## What is wrong

All six fixtures carry the same `policy_digest`, and none of them equals `current_policy_digest` in `expected.json`:

```
current_policy_digest              sha256:bda8d8f6...
01-valid-accepted.json             sha256:b5af974a...  != current   declared=pass
02-valid-denied.json               sha256:b5af974a...  != current   declared=pass
03-signature-key-mismatch.json     sha256:b5af974a...  != current   declared=n/a
04-broken-chain.json               sha256:b5af974a...  != current   declared=pass
05-stale-policy-digest.json        sha256:b5af974a...  != current   declared=fail
06-session-binding-mismatch.json   sha256:b5af974a...  != current   declared=pass

distinct policy_digest values across all six: 1
```

By the comparison the README states for `05` — "`policy_digest` names a policy bundle that is no longer the one in force (`current_policy_digest` in `expected.json`)" — `01`, `02`, `04` and `06` fail it too while being declared `pass`. `issued_at` does not rescue them: there is no validity window in the payload, and the stated check is a digest comparison.

That lands on what the set is for. The README says `04`, `05` and `06` "make one point three ways", but on the freshness axis `05` was indistinguishable from every other fixture, and a second implementation running `expected.json` as the portable contract disagrees with it on four of six.

## The cause, and the fix

`decisionPayload()` hardcoded `POLICY_V1` with no override, so `r05` got V1 like everything else — even though the comment directly above it says V1 is what makes `05` stale. The intent was in the comment; the code did not carry it out.

```diff
-function decisionPayload({ tool, decision, reason, sessionId, issuedAt, prevHash }) {
+function decisionPayload({ tool, decision, reason, sessionId, issuedAt, prevHash, policyDigest = POLICY_V2 }) {
     ...
-    policy_digest: POLICY_V1,
+    policy_digest: policyDigest,

 const r05 = envelope(decisionPayload({
   tool: 'run_shell', decision: 'allow', sessionId: SESSION,
   issuedAt: '2026-07-08T09:00:12.000Z',
+  policyDigest: POLICY_V1, // the only fixture carrying the superseded policy
 }), SIGNER_PRIV);
```

Everything downstream follows on its own, because `envelopeHash(r01)` is computed rather than pinned: signatures change, the chain head moves from `99c36f17` to `ddf7efb0`, and `02` and `04` relink automatically, with `04`'s deliberately wrong pointer staying deliberately wrong. `05` is byte-identical to before.

Before touching anything I ran the generator unmodified and confirmed it reproduces every committed file byte for byte, so the regenerated output can be trusted rather than assumed.

## Why nothing caught it, and what does now

The generator's self-check asserted that `05` fails freshness and `06` fails binding, and nothing about the other four. Asserting the failure alone passes whether one fixture is stale or all six are. That asymmetry is the whole bug.

- **`gen.mjs`** now checks both axes for every fixture against the manifest, skipping only `n/a`.
- **`tests/test_acta_fixtures.py`** read the `policy_freshness` and `session_binding` declarations nowhere at all. The two tests that existed were hardcoded to one fixture each and asserted *inequality*, which passes either way. Replaced with parametrized tests over every fixture; on the pre-fix data they fail for exactly 01, 02, 04 and 06, which I checked rather than assumed.
- **The crosswalk** quotes `01` in full and says its `chain_head` "is real". Both went stale the moment the fixtures were regenerated, and nothing would have noticed. Resynced, and a test now pins the document against the fixtures.

Also dropped: the generator note saying freshness for 01/02/04/06 was "pass relative to their v1-era chain context in the docs narrative". The manifest no longer needs it, and it was never written into `expected.json`, so the portable contract carried those four declarations with nothing qualifying them.

## Checks

214 passed, 1 skipped. `ruff check src tests` and `mypy src/agentrust_trace` clean.

Unchanged and worth repeating from the review: the hand-rolled `canonicalize()` in `gen.mjs` is correct. I differential-tested it against `rfc8785` over ASCII, non-ASCII values, supplementary-plane values, supplementary-plane keys, nested arrays, large integers, floats, exponent-form floats, empty containers and escape sequences — identical bytes in every case. That is by construction rather than luck, since JCS is defined in terms of ECMAScript.

### DIFF
--- a/docs/crosswalks/acta-decision-receipts.md
+++ b/docs/crosswalks/acta-decision-receipts.md
@@ -24,7 +24,7 @@ Per Acta s2.1, a receipt is a two-field envelope: a `payload` and a `signature` 
     "type": "protectmcp:decision",
     "tool_name": "run_shell",
     "decision": "allow",
-    "policy_digest": "sha256:b5af974ae7e1c6e4a656c9637c68ec0755a6cffa7861e196c0aaa2d6a7874cb5",
+    "policy_digest": "sha256:bda8d8f60c90afe5c05fc7f3b1b53d4715a0f7ba72ec8d4b82ad7aba70d9609b",
     "session_id": "ses_8f31ab",
     "issued_at": "2026-07-08T09:00:01.000Z",
     "issuer_id": "sb:issuer:QUGuJV1P6e6c"
@@ -32,7 +32,7 @@ Per Acta s2.1, a receipt is a two-field envelope: a `payload` and a `signature` 
   "signature": {
     "alg": "EdDSA",
     "kid": "sb:issuer:QUGuJV1P6e6c",
-    "sig": "d709be9e8d907c14bd2005df05cc960b070d388a665964074a19a355a2eabb3d..."
+    "sig": "cf4665db8613b3b8b398fae6c1bacf88e4963d300af63cf709a2be1e8a6bda69..."
   }
 }
 ```
@@ -68,7 +68,7 @@ Per the direction set on [trace-spec#97](https://github.com/agentrust-io/trace-s
 ```json
 {
   "evidence_type": "acta/decision-receipt-chain",
-  "chain_head": "99c36f17af17d48c2e6aab51ca21d9a08c53dcffbbb1f5b847c76fda9bf098bd",
+  "chain_head": "ddf7efb089f7504bbe880a623976b34be04b0bd4a991cdec12288841030229e4",
   "receipt_count": 2,
   "issuer_key_id": "sb:issuer:QUGuJV1P6e6c"
 }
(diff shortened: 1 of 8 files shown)


imran-siddique (approved):
Checked the fixtures directly rather than reading the table: on this branch, `01`, `02`, `03`, `04` and `06` now equal `current_policy_digest`, and `05-stale-policy-digest.json` keeps the old value, which is the whole point of that fixture. Suite passes, 214.

The defect is worth naming precisely, because it is the kind that makes a conformance set actively misleading rather than merely incomplete: five fixtures declared `pass` while failing the exact comparison the README defines for `05`. A set where the negative case is indistinguishable from the positives does not test what it says it tests, and a downstream implementer calibrating against it would have concluded the stale-policy check was unimplementable.

Generating the digest from the same source as `expected.json` in `gen.mjs` rather than a `POLICY_V1` constant is the part that stops it recurring.

Merging both this and #136.

Issue #19, opened 2026-06-11
enforcement_mode 'silent' legitimizes no-logging policy mode


imran-siddique:
`schema/trace-claim.json` allows `policy.enforcement_mode: "silent"`, documented as "no logging":

```json
"enforcement_mode": {
  "type": "string",
  "enum": ["enforce", "advisory", "silent"],
  "description": "How policy decisions were applied: enforce (block on deny), advisory (log only), silent (no logging)."
}
```

This undercuts the core value proposition of TRACE as tamper-evident governance evidence. A record can simultaneously claim hardware-attested policy binding (`policy.bundle_hash`) and declare that policy decisions were neither enforced nor logged. Verifiers and badges built on the schema will treat such records as conformant, giving "silent" deployments the same trust mark as enforcing ones.

Recommendation (either):
- Remove `silent` from the enum. A deployment that does not log policy decisions has nothing for the Trust Record to attest about policy application, and should not get a conformant record.
- Or, if `silent` must remain for migration purposes: require in regulated profiles that records with `enforcement_mode: "silent"` carry a non-affirming `appraisal.status` (e.g. `warning` or `contraindicated`) so they can never appraise as fully trustworthy, and state this normatively in the spec.

Generated with Claude Code

PR #143, opened 2026-08-09
feat(spec): enforcement_mode gains "declared", for policy nothing evaluated

imran-siddique:
Both framework adapters hit this independently today, which is what makes it a vocabulary problem rather than an adapter problem.

## The gap

`enforce`, `advisory` and `silent` all assert that **something evaluated the policy** — `enforce` acted on the result, `advisory` did not, `silent` acted with the operational logs suppressed. There was no value for a policy that is named and bound into the signed record with nothing evaluating it.

That is not a corner case. An agent framework has no policy engine, so a record built by observing a LangChain or LlamaIndex run has a policy the operator declares and no evaluation of it anywhere. With three values such a record has to claim an evaluation that never happened.

Both adapters refused to default the field and documented that `advisory` overstates a bare run. Honest, and it still leaves every framework-produced record marginally untrue — which belongs in the vocabulary, not in two READMEs.

## `declared`

The weakest value, and never a default:

- A producer that evaluates policy **MUST NOT** use it.
- A consumer **MUST NOT** read it as evidence that any rule was checked.
- A verifier appraising for enforcement **SHOULD** treat it as it treats an absent enforcement claim.

It says only: this is the policy the deployment states it was operating under.

## Compatibility

Additive to a **closed** enum in the model and in both JSON schema copies — a test pins that adding a value did not turn the field into free text, and `monitor` is still rejected by both.

Same one-directional consequence as `origin`: a verifier older than this release rejects a record carrying `declared`. Producers should not emit it until their consumers are current.

246 tests, `ruff` and `mypy` clean.

Follow-up once released: both framework adapters can default to `declared` instead of refusing, which is the first time either of them will be able to fill that field truthfully without the caller thinking about it.


solloek369-arch:
One narrow clarification on silent across the existing lineage.

[#19](https://github.com/agentrust-io/trace-spec/issues/19) raised the original concern around silent. [#28](https://github.com/agentrust-io/trace-spec/pull/28) then clarified it as allowing the governed action while suppressing operational logs and preserving every would-have-denied decision in the audit chain.

[#143](https://github.com/agentrust-io/trace-spec/pull/143) later describes silent as having “acted” on the policy result. Current TRACE also describes it as [“evaluated and enforced” in the schema](https://github.com/agentrust-io/trace-spec/blob/e7e2ecab68cf3534c7d5fcb7e9a6f089fcb7d592/schema/trace-claim.json#L158-L167), while [§4.3](https://github.com/agentrust-io/trace-spec/blob/e7e2ecab68cf3534c7d5fcb7e9a6f089fcb7d592/spec/trace-v0.2.md#L474-L482) still specifies the audit/log behavior without explicitly stating the deny disposition.

So one narrow question:

policy decision = deny
enforcement_mode = silent

MUST the governed action be allowed or blocked?

Current cMCP [allows a Cedar-denied call through in SILENT](https://github.com/agentrust-io/cmcp/blob/24b05aec7475859f61c4aade36983d0787ba4cc3/src/cmcp_runtime/policy/evaluator.py#L204-L243) and [projects that mode as TRACE silent](https://github.com/agentrust-io/cmcp/blob/24b05aec7475859f61c4aade36983d0787ba4cc3/src/cmcp_runtime/audit/trace_claim.py#L378-L384).

Did [#143](https://github.com/agentrust-io/trace-spec/pull/143) intentionally change the deny disposition established in [#28](https://github.com/agentrust-io/trace-spec/pull/28), or is “acted” / “enforced” only meant to distinguish evaluated silent from declared?

I’m not assuming either answer; I just want to pin the intended semantics.

imran-siddique:
The intended silent-mode behavior remains the allow behavior recorded in #28: a policy deny does not block the action, operational logs may be suppressed, and the audit chain still records every would-have-denied decision. Enforce remains the default; silent requires explicit configuration. The later schema wording saying silent is enforced is inconsistent with that behavior and should be corrected.

PR #146, opened 2026-08-09
fix(provenance): apply the structural rules on the consumer side too

### LINKED ISSUE #142: provenance.verify_record accepts records build_record refuses, including tee-attested with no evidence
**What is wrong**

`agentrust_trace.provenance.verify_record` (at `src/agentrust_trace/provenance.py:178`
— not the same-named function in `sign.py`, which is unrelated and does enforce
freshness on `iat`) accepts records that `provenance.build_record` refuses. Four
structural rules are enforced only on the producer side:

- a `tee-attested` record with `attestation: null`,
- an `endpoint` identity with no `spki_sha256`,
- an `artifact` identity with no `digest`,
- `issued_at` missing, or set to `"tomorrow"`, or negative.

A record does not have to come from `build_record` — anyone can write the JSON and
sign it — so the consumer side is the one that has to enforce these, and it does not.

Reproduction against `671f2a8`. The record is hand-assembled, never passed through
`build_record`, and signed by a key the consumer already trusts (so §5 step 3 is
satisfied by assumption):

```python
from agentrust_trace import provenance as P
from agentrust_trace.sign import generate_key, key_to_jwk

key = generate_key(); jwk = key_to_jwk(key)
tools = [{"name": "search", "description": "search the docs",
          "input_schema": {"type": "object"}}]

forged = {
    "format": P.FORMAT,
    "kind": "tee-attested",                                        # top trust tier
    "identity": {"endpoint": {"url": "https://mcp.acme.example"}}, # no spki_sha256
    "publisher": "did:web:acme.example",
    "tool_catalog": {"hash": P.tool_catalog_hash(tools), "tool_count": 1},
    "attestation": None,                                           # no evidence
    # issued_at omitted entirely
}
signed = P.sign_record(forged, key)

P.verify_record(signed, jwk)        # passes
P.check_tool_catalog(signed, tools) # passes
```

Both steps of §5 pass. The consumer is handed a record claiming `tee-attested` with
`attestation: null`, an endpoint identity that is a bare URL, and no `issued_at`.

The asymmetry, rule by rule:

| Rule | `build_record` | `provenance.verify_record` | Spec |
|---|---|---|---|
| `tee-attested` requires attestation | refuses | **accepts** | §3 table: required when `kind` is `tee-attested` |
| `endpoint` requires `spki_sha256` | refuses | **accepts** | §2, and the code's own message: "A URL on its own is not an identity" |
| `artifact` requires `digest` | refuses | **accepts** | §2 |
| `issued_at` | sets an int | **never inspected** | §3 table: required |

`identity` is tested for truthiness of `artifact`/`endpoint` only; the members inside
are never looked at. `tool_catalog.tool_count` is likewise never checked — it is read
only to build the `ToolCatalogMismatch` message.

Two smaller things in the same function:

- **Trailing newline.** `_PUBLISHER_RE` and `_DIGEST_RE` anchor with `$`, and in
  Python `$` matches before a single final newline, so
  `publisher = "did:web:acme.example\n"` passes, as does a `tool_catalog.hash` with
  one trailing newline. `\Z` is the intended anchor. (The hash case fails safe:
  `check_tool_catalog` then rejects it on the `!=` comparison.)
- **A false negative, opposite direction.** The embedded-key guard is
  `embedded != trusted_jwk`, a dict comparison, so a legitimately signed record whose
  `cnf.jwk` carries an extra member — `kid`, say — is rejected even though the key
  material is identical. Comparing key material, or the JWK thumbprint, would avoid
  refusing honest publishers.

**Spec section or file**

`src/agentrust_trace/provenance.py:178` (`verify_record`), measured against
`spec/server-provenance-v1.md` §2 (Identity), §3 (The record — required-field table),
and §5 (Verification). Tests: `tests/test_provenance.py`.

**Expected behavior**

`verify_record` should refuse every record `build_record` would refuse. The
cleanest form is to move the structural rules into a helper both sides call, so the
two paths cannot drift, and have `verify_record` run it before the signature check —
the same ordering argument as the profile check in #125: the only action taken on the
unauthenticated value is refusal. `issued_at` should be required and type-checked;
the two regexes should anchor with `\Z`; the embedded-key guard should compare key
material rather than dict equality.

**Impact**

Verifiers accept records that are structurally invalid by the spec's own §2/§3. This
is **not** unauthenticated forgery, and should not be read as such: `verify_record`
requires a `trusted_jwk` the consumer already holds, and I confirmed before filing
that stripping `cnf`, omitting its `jwk`, or signing with an untrusted key all get
rejected — the signature is the real gate. The reachable case is a publisher you
already trust asserting a higher tier than it holds evidence for: a
`publisher-asserted` signer presenting itself as `tee-attested` and passing every
check in §5. Since `kind` is closed precisely so "a verifier can key on it", that is
the field's whole value, and nothing downstream re-checks it — `check_tool_catalog`
re-checks only the tool hash.

Why the tests miss it: `test_tee_attested_without_evidence_is_refused`,
`test_endpoint_url_without_a_key_digest_is_refused`, and
`test_artifact_digest_must_be_a_digest` all call `build_record` only. No test in the
file puts a structurally-invalid-but-correctly-signed record through `verify_record`
(the verification tests cover tampered signatures, wrong keys, unsigned records, and
unknown formats — all over structurally valid input). So the checks can be absent
from the consumer path with the suite green: deleting the producer-side checks fails
three tests; the consumer side has no check to delete.

imran-siddique:
Closes #142.

Four structural rules lived only in `build_record`:

- a `tee-attested` record must carry attestation evidence
- an `endpoint` identity must carry `spki_sha256`
- an `artifact` identity must carry `digest`
- `issued_at` must be a usable timestamp

A record does not have to come from `build_record`. Anyone can write the JSON and sign it, so a rule on the producer side is a rule an attacker never runs. @lywinged's reproduction on #142 shows the consequence: a record claiming `kind: "tee-attested"` with `attestation: null`, an endpoint identity that is a bare URL, and no `issued_at`, signed by a key the consumer already trusts, passes both steps of §5. The consumer is handed the top trust tier with nothing behind it.

The rules move into a single `_check_structure` that both `build_record` and `verify_record` call. Copying them into the verifier would have worked today and drifted later, and the copy that matters is the consumer's.

This is the same principle #144 made explicit in the C2PA assertion next door: a check a verifier must not be able to skip is required, not offered. @lywinged drew that connection on the issue.

**Tests.** Nine cases, one per rule plus the reported forgery verbatim and a well-formed control. Run against the currently published `agentrust-trace`, eight of the nine fail and the control passes, so they are guarding the actual defect.

**Compatibility.** This rejects records the verifier previously accepted. Every such record is one `build_record` would have refused to emit, so anything produced through the SDK is unaffected.

### DIFF
--- a/src/agentrust_trace/provenance.py
+++ b/src/agentrust_trace/provenance.py
@@ -91,6 +91,60 @@ def tool_catalog_hash(tools: list[dict[str, Any]]) -> str:
     return "sha256:" + hashlib.sha256(canonical).hexdigest()
 
 
+def _check_structure(
+    *,
+    kind: str,
+    artifact: dict[str, Any] | None,
+    endpoint: dict[str, Any] | None,
+    attestation: dict[str, Any] | None,
+    issued_at: Any,
+) -> None:
+    """The structural rules both the producer and the consumer must apply.
+
+    Shared rather than duplicated because these rules were enforced only in
+    :func:`build_record` for a time, and a record does not have to come from
+    ``build_record``: anyone can write the JSON and sign it. A rule that lives
+    only on the producer side is a rule an attacker simply does not run, so a
+    ``tee-attested`` record with ``attestation: null`` verified cleanly.
+
+    Keeping one implementation is the point. Two copies drift, and the copy that
+    matters is the one on the consumer side.
+    """
+    if artifact is not None:
+        if not artifact.get("package"):
+            raise ProvenanceError("artifact.package is required (a Package URL)")
+        if not _DIGEST_RE.match(str(artifact.get("digest", ""))):
+            raise ProvenanceError(
+                "artifact.digest must be a sha256: digest of the entrypoint. For an "
+                "interpreted server that is the script, not the interpreter: every such "
+                "server on a host shares one interpreter digest."
+            )
+    if endpoint is not None:
+        if not endpoint.get("url"):
+            raise ProvenanceError("endpoint.url is required when endpoint is present")
+        if not _DIGEST_RE.match(str(endpoint.get("spki_sha256", ""))):
+            raise ProvenanceError(
+                "endpoint.spki_sha256 must be a sha256: digest of the Subject Public Key "
+                "Info. A URL on its own is not an identity."
+            )
+    if kind == "tee-attested" and not attestation:
+        raise ProvenanceError(
+            "kind='tee-attested' without attestation evidence is the claim without the "
+            "thing that backs it"
+        )
+    if kind != "tee-attested" and attestation:
+        raise ProvenanceError(
+            f"kind={kind!r} carries attestation evidence. Evidence that is present but "
+            "not claimed invites a consumer to read it as an attestation that was made."
+        )
+    # bool is an int subclass, and True would otherwise pass as a timestamp.
+    if not isinstance(issued_at, int) or isinstance(issued_at, bool) or issued_at < 0:
+        raise ProvenanceError(
+            "issued_at must be a non-negative integer Unix timestamp. A record with no "
+            "issue time cannot be aged, so a consumer has no way to reject a stale one."
+        )
+
+
 def build_record(
     *,
     kind: str,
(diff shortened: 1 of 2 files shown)


lywinged:
Thanks, pulling the structural rules onto the verify path via one _check_structure
is exactly it, and rejecting bool for issued_at is a nice catch I'd missed.

Two smaller things from the bottom of #142 are in the same function and untouched by
this PR, if you want them in the same pass:

- The regexes still anchor with $ (lines 43–44). In Python $ matches before a single
  trailing newline, so publisher = "did:web:acme.example\n" and a tool_catalog.hash
  with one trailing newline both still pass verify_record. \Z is the intended anchor.
  (The hash case fails safe — check_tool_catalog rejects it on the != — but the
  publisher one does not.)

- The embedded-key guard is embedded != trusted_jwk (line 253), a dict comparison, so
  a legitimately signed record whose cnf.jwk carries an extra member like kid is
  refused even though the key material is identical. Comparing key material, or the JWK
  thumbprint, would avoid turning away honest publishers.

Both are minor next to the main fix; flagging only because they're the same few lines
you're already in.

Issue #117, opened 2026-08-02
spec: disclosed receipt gaps — GapDisclosure claim + receipt_gap_disclosed verifier outcome


lywinged:
**Spec section affected**

`docs/verification.md` — "Action receipts and embodied workflows" (the four-outcome table); `spec/trace-v0.2.md` §3.3 (verification protocol, receipt outcomes); `schema/trace-claim.json` (one new optional claim — schema PR to track the spec PR per CONTRIBUTING).

**Problem**

The four action-receipt verifier outcomes conflate two materially different missing-receipt cases under `receipt_missing_required`:

1. an **adversarial or negligent silence** — a receipt for a consequential action is simply absent; and
2. an **operator-disclosed, bounded loss** — the emitter crashed, lost a tail of receipts, and said so.

Under a profile with `action_receipts: required` (#66), completeness is the load-bearing property — but no emitter can be made gap-proof: any high-throughput writer has a crash window in which a tail of receipts is lost. As currently specified, an honest operator who loses 50 ms of receipts in a crash produces the **same verifier outcome** as an adversary who silently deleted the receipts for a consequential action.

This creates a perverse incentive: backfilling or fabricating receipts scores better than disclosing the loss. An evidence standard should reward honesty under failure, not punish it.

**Proposed change**

*1. New optional, chain-bound claim `GapDisclosure`:*

```json
{
  "type": "GapDisclosure/1.0",
  "range_start_after": "<chain position of the last receipt before the gap>",
  "range_end_before": "<chain position of the first receipt after resumption>",
  "cause": "crash | shutdown | backpressure | unknown",
  "receipts_lost_estimate": 3,
  "disclosed_at": "<chain position where this disclosure is sealed>"
}
```

- Gap boundaries MUST be expressed in chain/log order, not timestamps — a key that signs receipts also signs its own timestamps, so self-reported time cannot bound a gap (same rationale as the entry-ID anchoring in #67).
- The disclosure MUST be signed by the receipt-issuing key or its hierarchical parent, and MUST itself be sealed into the receipt chain at the resumption point, so the disclosure is as tamper-evident as the receipts it accounts for.
- `receipts_lost_estimate` MAY be null when the emitter cannot bound the count; the range bounds are still normative.

*2. Fifth verifier outcome, and a re-scope of the fourth:*

| Outcome | Meaning |
|---|---|
[...]

PR #148, opened 2026-08-09
test: rule registry the verifier consumes, and two independent vectors per rule

### LINKED ISSUE #124: test(conformance): single-vector coverage has no margin — define and enforce independence
## Problem

Every rule the action-receipt verifier applies is exercised by at most one fixture. After
#122 lands there are no uncovered rules, and that is the weaker of the two properties
worth having. Coverage with a margin of zero degrades to no coverage the moment a single
fixture is deleted, renamed out of the set, or weakened — and nothing in the repository
notices.

That is not a hypothetical about carelessness. It is the ordinary lifecycle of a fixture
set: a case gets rewritten during a refactor, an expectation gets relaxed to make an
unrelated change pass, a file gets renumbered. Each of those is a normal PR, and any of
them can silently retire a rule from the suite.

## Two distinct failure modes

**One vector can stop testing what it names.** A vector that triggers two rules at once
still passes while only one of them is doing the work; remove the other and it keeps
passing for the wrong reason. This is not theoretical — while preparing #122 I wrote an
untrusted-key vector that fired both `issuer_key_untrusted` and an issuer-binding rule,
and had to restructure it so it isolated one. If it had merged as written, the suite
would have reported a rule as covered that nothing distinguished.

**Loss is silent.** Nothing currently compares what the verifier can emit against what
the fixtures expect, so a rule dropping to zero vectors produces no failure anywhere.

## What "independent" should mean

The word does most of the work in "add independent vectors", and left undefined it
becomes a matter of taste. A definition that is checkable:

> Two vectors for the same rule are **independent** if there exists a single
> implementation defect that causes one to pass and the other to fail.

If every single defect takes both out identically, they are one vector written twice —
whatever their field values differ by.

Concretely, for `issuer_key_untrusted`:

- **A** — the key identifier is absent from the trusted set entirely
- **B** — the identifier is present in the set, but bound to a different issuer

An implementation that looks up the identifier without checking the binding passes B and
fails A. That separation is what makes them two vectors rather than one.

The definition has a useful property: it is **mechanically checkable**. Inject a defect,
run the set, and confirm the two vectors do not fail together. That is mutation testing
applied to the vectors rather than to the implementation.

## Scope

I count **14 rules** in the current verifier — 13 failure codes and one warning — by
walking `tests/test_action_receipt_fixtures.py` with `ast` for every string literal
appended to a failure or warning list. If you are counting 20, you are likely including
something I am not looking at; say which and I will scope to that instead of to my count.

Against those 14, two independent vectors each is roughly 28, of which 9 exist today and
7 more arrive with #122.

## Proposed work

1. **At least two independent vectors per rule**, by the definition above, with the
   independence argued in the fixture description rather than assumed.
2. **A coverage check in CI**, so margin cannot silently return to zero. It should
   recover the rule inventory from the verifier's source rather than from a
   hand-maintained list — a list is correct only until the first person forgets it, and
   that failure is silent in the direction that matters.

I have (2) working and can contribute it separately from the vectors, since it is useful
on its own and reviewable on its own. It answers three questions: does any fixture expect
a code nothing emits, is every emitted code expected by some fixture, and does deleting
each rule change at least one fixture's outcome. The third is the one that catches
non-load-bearing coverage.

## One caveat on who writes the vectors

Vectors and the checker that runs them, written by the same hand in the same sitting,
agree with each other by construction. That agreement is worth very little on its own,
and it is worth less for vectors whose whole purpose is independence.

I would rather this set were not authored entirely by me. If someone else produces a
second set independently, the useful artifact is not either set but the **cross-run**:
their vectors through a verifier they did not write, and mine through one I did not. Where
the two disagree about an expected outcome, that is either a defect or a rule whose
semantics are genuinely ambiguous — and the second is more valuable to find than any
individual vector.

Happy to do it either way. Flagging it because "we wrote more vectors" is not the same
claim as "these vectors are independent", and only the second one is what is being asked
for here.

Refs #122

imran-siddique:
Closes #124. **Authored by @lywinged**, from their branch `lywinged/trace-spec:feat/124-registry-and-second-vectors`. I rebased it onto current main (it was 33 commits behind) and dropped `uv.lock`; the commit is theirs and the authorship is intact. Opened as a maintainer because the branch had no PR.

**What it does.** Every action-receipt obligation was exercised by exactly one fixture, so deleting, renaming or weakening any single vector silently retired a rule and nothing failed. This gives each of the 14 obligations a second, independent vector and makes the margin enforceable.

The design follows @carloshvp's review on the issue. The AST walk for string literals is retired as the inventory source, because `append`, `extend`, constants and helper returns are silent blind spots. In its place the verifier consumes an explicit tuple of named `Rule` entries and emits every code from a single loop, so an unregistered check is a check that never runs. The AST survives inverted: it now asserts that nothing else in the module appends to a failure or warning list, so an emission path bypassing the registry fails the suite.

Independence is executable rather than asserted. The issue's definition is "one implementation defect makes one vector pass and the other fail", which only means something if the defects are named, so each rule declares at least one weakened variant of its check: digest-prefix comparison, case-normalised identifiers, clock tolerance, signature structure validated without cryptography. Two vectors count as independent only when a declared defect moves one and leaves the other alone, and a rule with no declared defect fails the suite.

`tests/vector_margins.json` is the ratchet: 14 rules, all at margin 2, so anything above the floor cannot decay back to it unnoticed.

**Verified on current main after the rebase:** 313 passed, 1 skipped, ruff clean on `tests/` and `examples/`.

**One limit @lywinged states themselves**, worth keeping visible: independence is relative to the declared defect set, so a shortcut nobody thought to declare is still invisible. It is a floor on discrimination, not a proof of it. They also noted they would rather the vectors were not all authored by one hand, and that a second independently written set would be more valuable as a cross-run than either set alone. That remains open and is not a reason to hold this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ htmlcov/
 # Virtualenvs
 .venv/
 venv/
+uv.lock
--- a/examples/action-receipts/README.md
+++ b/examples/action-receipts/README.md
@@ -96,6 +96,20 @@ three operations:
 | `14-receipt-issuer-key-unknown.json` | `receipt_unverified` with advisory | unknown | The issuer key is not in the verifier's pinned set. Unverifiable is not invalid (spec §3.3.1): no trust is conferred and no forgery is proven, surfaced as an `issuer_key_unknown` advisory. |
 | `15-receipt-from-future.json` | `receipt_invalid` | unknown | Issued after the verification time, so an upper bound on age never rejects it. |
 | `16-decision-not-in-enum.json` | `receipt_invalid` | unknown | An unrecognised decision verb, which must not read as accept or reject. |
+| `17-missing-receipt-explicit-null.json` | `receipt_missing_required` | unknown | The receipt supplied as an explicit `null`: 03's absence through a different door, misread by presence-checking implementations. |
+| `18-action-ref-tail-forged.json` | `receipt_invalid` | unknown | The declared `action_ref` matches the recomputed digest in every character but the last. |
+| `19-action-ref-mismatch-in-tail.json` | `receipt_invalid` | unknown | Receipt and action references differ only in the final character. |
+| `20-call-id-case-mismatch.json` | `receipt_invalid` | unknown | The linked call id is this call's id in a different case — a different call. |
+| `21-session-id-case-mismatch.json` | `receipt_invalid` | unknown | The session id in a different case — a different session. |
+| `22-evidence-hash-mismatch-in-tail.json` | `receipt_invalid` | unknown | The evidence hash is wrong only in its final character. |
+| `23-receipt-issuer-key-case-variant.json` | `receipt_unverified` with advisory | unknown | The right key pinned under a case-variant of the receipt's key id: not the key the receipt names. |
+| `24-receipt-signature-malformed.json` | `receipt_invalid` | unknown | Valid base64url decoding to 32 bytes — rejected by structure, where 04 needs cryptography. |
+| `25-stale-receipt-boundary.json` | `receipt_invalid` | unknown | Stale by exactly one second. |
+| `26-receipt-from-future-boundary.json` | `receipt_invalid` | unknown | Issued one second after the verification time. |
+| `27-receipt-chain-gap-in-tail.json` | `receipt_invalid` | unknown | The predecessor link is wrong only in its final character. |
+| `28-physical-completion-claim-case.json` | `receipt_invalid` | unknown | The claim `"None"`: the vocabulary word under case-normalisation, outside it as bytes. |
+| `29-same-party-self-report-rejected.json` | `receipt_valid_rejected` with warning | `rejected` | A gateway self-report on the rejected branch: issuer independence matters on both. |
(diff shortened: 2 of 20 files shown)


lywinged:
Thank you for doing the rebase. That branch was 33 commits behind and I had left it without a PR on purpose, so this is work you took on that was mine.

Confirming the half only the author can: I diffed the rebased head against `feat/124-registry-and-second-vectors`, and `tests/test_action_receipt_fixtures.py`, `tests/test_vector_completeness.py` and `tests/vector_margins.json` are byte-identical. You established that it runs; this is the other half, that it moved without changing.

One thing worth stating before it becomes a question. The branch carries 14 registered rules here and 22 on the fork's `main`. The extra 8 are the `GapDisclosure` rules from #117, an unaccepted proposal, so they are deliberately outside this PR and their fixtures sit under a separate `proposal-117/` directory with their own numbering. The 14 here stand on their own.

Your summary of the limit is accurate, including the part about a second independently written vector set. That one is still open.

PR #149, opened 2026-08-10
fix(provenance): identify a JWK by its key material, not its dict shape

lywinged:
## What this changes

`provenance.verify_record` compared the record's `cnf.jwk` against the trusted key
with dict equality:

```python
if embedded and embedded != trusted_jwk:
    raise ProvenanceError("the record's embedded key is not the trusted key. ...")
```

A JWK is identified by its key material. `kid`, `use` and `alg` are optional members
that carry none of it. `key_to_jwk` emits the bare `{crv, kty, x}`, while a key
resolved from a JWKS endpoint normally carries `kid`, because that is how JWKS
distinguishes keys across rotation. The two are the same public key and not the same
dict, so a consumer holding its trusted key from a JWKS rejected every record:

```python
from agentrust_trace import provenance as P
from agentrust_trace.sign import generate_key, key_to_jwk, _pubkey_from_jwk

key = generate_key(); jwk = key_to_jwk(key)
tools = [{"name": "search", "description": "d", "input_schema": {"type": "object"}}]
record = P.build_record(
    kind="publisher-asserted", publisher="did:web:acme.example", tools=tools,
    endpoint={"url": "https://mcp.acme.example", "spki_sha256": "sha256:" + "a" * 64},
)
signed = P.sign_record(record, key)

from_jwks = {**jwk, "kid": "2026q3", "use": "sig"}          # the same key, from a JWKS
_pubkey_from_jwk(jwk).public_bytes_raw() == _pubkey_from_jwk(from_jwks).public_bytes_raw()
# True

P.verify_record(signed, from_jwks)
# ProvenanceError: the record's embedded key is not the trusted key. A record signed
# by some other key is a record about a server somebody else is describing.
```

The signature verified. Only the comparison failed — and the message accuses the
publisher of substituting a key when it used exactly the right one, which leaves an
implementer with nowhere to look: valid signature, correct key, and an error saying
the key is wrong.

Now compared by RFC 7638 thumbprint. `sign.jwk_thumbprint` already implements it, and
its docstring already states the property this needs:

> stable across JWKs that differ in optional members such as `kid`, `alg`, or `use`

The check itself is unchanged in intent — a record whose embedded key is a *different*
key is still refused, and #146's posture depends on that. Only the comparison widens.
A `cnf.jwk` with no usable `kty` is now a refusal rather than an uncaught `ValueError`.

**Also**, `_PUBLISHER_RE` and `_DIGEST_RE` anchored with `$`, which in Python matches
before one final newline, so `publisher = "did:web:acme.example\n"` and a
`tool_catalog.hash` with a trailing newline both passed `verify_record`. Both now
anchor with `\Z`.

Both were raised on #146 and were not part of that fix; #142 is closed, so neither was
being tracked anywhere.

## Tests

Six, and each is load-bearing rather than merely present — reverting the thumbprint
comparison to dict equality fails three of them, reverting the anchors fails the other
two:

- a trusted key carrying `kid`/`use` still verifies (the deployment case)
- an embedded `cnf.jwk` carrying `kid` still verifies
- a genuinely different embedded key is still refused (the widening does not overshoot)
- an unusable `cnf.jwk` is refused rather than crashing
- `publisher` may not carry a trailing newline
- `tool_catalog.hash` may not carry a trailing newline

The embedded-key case is signed by hand, because `sign_record` builds `cnf` itself from
`key_to_jwk` and this implementation therefore cannot emit a `cnf.jwk` carrying `kid`.
Another implementation can, and the format permits it, so the verifier has to accept it.

329 tests pass; ruff and mypy clean.

## Type of change

- [ ] Schema change
- [ ] Example addition

Strictly a library fix; no spec text changes. Ticking non-breaking as the closest fit —
records and keys that verified before still verify, and some that were wrongly refused
now verify.

## Spec section

None. `spec/server-provenance-v1.md` is unchanged; §3 already lists `cnf.jwk` as "public
key the signature verifies under", which is the reading this restores.

## Checklist

- [ ] `CHANGELOG.md` updated (for any normative change) — not applicable, no normative change
- [ ] Breaking changes marked — not applicable
- [ ] Backward compatibility statement — not applicable

### DIFF
--- a/src/agentrust_trace/provenance.py
+++ b/src/agentrust_trace/provenance.py
@@ -21,7 +21,12 @@ import re
 import time
 from typing import Any
 
-from agentrust_trace.sign import _canonical_bytes, _pubkey_from_jwk, key_to_jwk
+from agentrust_trace.sign import (
+    _canonical_bytes,
+    _pubkey_from_jwk,
+    jwk_thumbprint,
+    key_to_jwk,
+)
 
 __all__ = [
     "FORMAT",
@@ -252,11 +257,23 @@ def verify_record(record: dict[str, Any], trusted_jwk: dict[str, Any]) -> None:
         raise ProvenanceError("record carries no signature")
 
     embedded = (record.get("cnf") or {}).get("jwk")
-    if embedded and embedded != trusted_jwk:
-        raise ProvenanceError(
-            "the record's embedded key is not the trusted key. A record signed by some "
-            "other key is a record about a server somebody else is describing."
-        )
+    if embedded:
+        # Compared by RFC 7638 thumbprint, not by dict equality. A JWK is identified by
+        # its key material; `kid`, `use` and `alg` are optional members that carry none
+        # of it, and a key resolved from a JWKS endpoint normally has `kid` while
+        # `key_to_jwk` emits the bare minimum. Dict equality made that difference fatal
+        # and rejected records signed by exactly the right key.
+        from hmac import compare_digest
+
+        try:
+            matched = compare_digest(jwk_thumbprint(embedded), jwk_thumbprint(trusted_jwk))
+        except ValueError as exc:
+            raise ProvenanceError(f"the record's embedded key is unusable: {exc}") from exc
+        if not matched:
+            raise ProvenanceError(
+                "the record's embedded key is not the trusted key. A record signed by "
+                "some other key is a record about a server somebody else is describing."
+            )
 
     pub = _pubkey_from_jwk(trusted_jwk)
     body = _canonical_bytes({k: v for k, v in record.items() if k != "signature"})
--- a/tests/test_provenance.py
+++ b/tests/test_provenance.py
@@ -8,6 +8,8 @@ produces.
 
 from __future__ import annotations
 
+import base64
+
 import pytest
 
 from agentrust_trace.provenance import (
@@ -20,7 +22,7 @@ from agentrust_trace.provenance import (
     tool_catalog_hash,
     verify_record,
 )
-from agentrust_trace.sign import generate_key, key_to_jwk
+from agentrust_trace.sign import _canonical_bytes, generate_key, key_to_jwk
 
 DIGEST = "sha256:" + "a" * 64
 OTHER_DIGEST = "sha256:" + "b" * 64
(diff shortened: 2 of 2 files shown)


lywinged:
Rebased onto dd1b6e6. 345 pass, ruff and mypy clean.

#157 landed the same comparison in `sign.verify_record` while this was open, so I have rewritten this to match it exactly:

```python
if not compare_digest(jwk_thumbprint(embedded), jwk_thumbprint(trusted_jwk)):
```

The error type still differs, `ProvenanceError` here against `ValueError` there, since each module raises its own. Everything else is now the same expression.

That makes the remaining state easy to state: after #157, main identifies a signing key by its RFC 7638 thumbprint in `sign.py` and by Python dict equality in `provenance.py`. Same repository, same question, two answers. This PR removes the second one.

Still six load-bearing tests on the rebased tree, re-measured rather than assumed: reverting the thumbprint comparison fails three, reverting the `\Z` anchors fails two.

imran-siddique:
The thumbprint comparison is right and matches what #157 landed in `sign.verify_record`, so the two paths agree now. The `kid` case is the real one: `key_to_jwk` emits the bare `{crv, kty, x}` and a JWKS serves the same key with `kid`, so dict equality refused records signed by exactly the right key.

One thing to do before this can merge. I have merged #154, so the `\Z` anchor change and a parametrized test covering all four guarded fields are now on main. This branch carries the same two lines, so it conflicts.

Could you rebase onto main and drop the anchor hunk plus the two newline tests? `test_publisher_may_not_carry_a_trailing_newline` and `test_catalog_hash_may_not_carry_a_trailing_newline` are both subsumed by #154's `test_a_trailing_newline_does_not_satisfy_a_pattern`, which also covers `artifact.digest` and `endpoint.spki_sha256`. That leaves this PR as the thumbprint change and its four key-identity tests, which is the right scope for it.

I did the rebase locally to check it is clean: it resolves to 76 insertions across the two files and the full suite passes at 347 passed, 1 skipped. So there should be no surprises.

lywinged:
Rebased onto 5367970 and narrowed to one thing.

#154 landed the `\Z` anchor change that this branch also carried, so that half is dropped here along with the two newline tests it duplicated — #154's four parametrized cases cover the same fields and one more. The only conflict was the comment above the two patterns, where we had both written one saying the same thing; I kept the merged wording.

What is left is the key comparison. #157 landed the same fix in `sign.verify_record` while this was open, so this is now written the same way rather than merely equivalently:

```python
if not compare_digest(jwk_thumbprint(embedded), jwk_thumbprint(trusted_jwk)):
```

The error type still differs — `ProvenanceError` here against `ValueError` there — since each module raises its own. Everything else is the same expression.

That makes the remaining state easy to say: on main today, `sign.py` identifies a signing key by its RFC 7638 thumbprint and `provenance.py` identifies it by Python dict equality. Same repository, same question, two answers. This removes the second one.

Four tests, each load-bearing, re-measured on the rebased tree rather than carried over: reverting the thumbprint comparison fails three. 347 pass, ruff and mypy clean.

imran-siddique (approved):
The provenance path now identifies JWKs by RFC 7638 thumbprint exactly as sign.verify_record already does, so optional JWKS metadata no longer makes the same key compare unequal. The different-key and unusable-key cases remain fail-closed, and duplicated newline-anchor work was correctly removed after rebase.

Issue #161, opened 2026-08-11
RFC Proposal: Composable Zero-Knowledge Proofs for TRACE


Trivo25:
**Spec section affected**

§2.3 Trusted Computing Base, §3.3 Verification, §4 Standards Composition (§4.1–§4.3)

This proposal is initially provided as an informative RFC rather than a direct modification of the normative TRACE wire format.

**Problem**

TRACE currently centers execution assurance around TEE-backed attestation. This provides strong guarantees about the measured runtime, execution boundary, host confidentiality and protected credentials.

However, TRACE does not currently describe zero-knowledge proofs as an alternative or complementary assurance mechanism.

ZK and TEEs establish different properties:

- A TEE can establish properties about the measured runtime and execution boundary.
- A ZK proof can establish properties about a defined computation, including computation over committed or private inputs.
- ZK can also provide privacy-preserving verification, where a verifier confirms a claim without receiving the complete underlying transcript or sensitive values.
- The two mechanisms can be composed when both runtime assurance and independently verifiable computation are required.

Treating these mechanisms as a single linear trust hierarchy would therefore be misleading. Implementations can require different combinations depending on their threat model, privacy requirements, and deployment constraints.

**Proposed change**

Introduce a proof-system-neutral, composable assurance model for TRACE.

Under this model, ZK and TEE are peer assurance mechanisms rather than levels above or below each other. An implementation can select and combine mechanisms according to the properties it requires.

Example assurance profiles include:

- TRACE + ZK
- TRACE + TEE
- TRACE + TEE + ZK
- TRACE + TEE + SCITT
- TRACE + TEE + ZK + SCITT

The proposed RFC defines the relevant assurance dimensions, including:

- computation integrity;
- runtime integrity and confidentiality;
- input authenticity;
- proof-gated downstream actions;
- external verification and evidence durability;
- privacy-preserving verification.

For example, a ZK proof can be verified by an external gateway before it permits a downstream action, without requiring the verifier or gate to execute inside the TEE.

ZK can also operate over evidence authenticated by a TEE. A TEE can commit to an execution transcript, while a ZK proof can establish selected policy claims over that commitment without disclosing the complete transcript.

The RFC does not require a specific ZK proof system. Selection of the proving system remains an implementation detail.

This RFC defines the assurance model and trust assumptions first. Exact wire-format fields, proof-system identifiers, verification rules, and conformance requirements can be proposed separately after the model is agreed.

**Backward compatibility**

[...]

PR #162, opened 2026-08-11
RFC Proposal: Composable Zero-Knowledge Proofs for TRACE

Trivo25:
## What this changes

Adds an informative RFC proposing **composable zero-knowledge proofs for TRACE**.

The RFC describes ZK and TEE as peer assurance mechanisms that provide different properties and can be used independently or composed depending on the required security, privacy, verification and deployment properties.

In particular, the RFC covers:

- ZK-based computation integrity
- TEE-based runtime integrity and confidentiality
- composition of TEE-authenticated evidence with ZK proofs
- privacy-preserving verification and selective disclosure
- externally verified, proof-gated downstream actions
- external receipts, transparency and evidence durability

The proposal is proof-system neutral and does not require a specific ZK proving system.

This PR is intentionally informative. It does not modify the TRACE wire format, schema, existing trust levels or conformance requirements. Follow-up changes can define normative fields and verification requirements after the assurance model is agreed.

Relates to https://github.com/agentrust-io/trace-spec/issues/161

## Type of change

- [ ] Schema change
- [ ] Example addition

## Spec section

No normative section of `spec/trace-v0.2.md` is modified by this PR.

The RFC is relevant to the specification's trust and verification model and proposes an assurance model that can inform future normative changes related to ZK evidence and its composition with existing TEE-backed TRACE evidence.

## Checklist

This PR is informative and non-breaking, so no `CHANGELOG.md` update, breaking-change marker or breaking-change compatibility statement is required.

### DIFF
--- a/docs/rfcs/composable-zk-assurance.md
+++ b/docs/rfcs/composable-zk-assurance.md
@@ -1,0 +1,326 @@
+# RFC Proposal: Composable Zero-Knowledge Proofs for TRACE
+
+**Status:** Draft issue proposal
+**Author:** Florian Kluge
+**Contact:** florian.kluge@o1labs.org
+**Organisation:** o1Labs
+**Scope:** Non-breaking assurance model extension
+**Target:** TRACE specification
+**Primary topic:** Zero-knowledge proof as a composable mechanism
+
+## Summary
+
+TRACE currently describes trust through software signing, hardware-backed attestation and transparency mechanisms.
+
+This proposal adds **zero-knowledge proofs** as an additional mechanism that can compose with the existing TRACE model.
+
+**TEE attestation and ZK proofs establish different properties.** TEE attestation can establish properties about the measured runtime and execution boundary, while ZK can establish properties about a defined computation. They can therefore be used independently or composed when both types of assurance are required.
+
+The proposal does not define zero-knowledge proofs as a trust level above or below a Trusted Execution Environment. ZK proofs and TEEs provide different security properties and rely on different trust assumptions. A ZK-based profile relies on the soundness of the proof system, the correct identification of the proven program or policy, authenticated and committed inputs and correct proof verification. A TEE-based profile relies on the hardware vendor and attestation chain, the measured workload and the correct deployment of the trusted runtime. In return, ZK can provide independently verifiable computation integrity without trusted hardware, while a TEE can provide runtime isolation, host confidentiality,and protected credential handling. The two mechanisms can be used independently or composed when an implementation requires both sets of properties.
+
+Instead, TRACE should support a **non-linear, composable assurance model**. An implementer can select and combine assurance mechanisms according to the required **security, privacy, verification and deployment properties**.
+
+For example, ZK can provide independently verifiable computation without trusted hardware or selectively prove claims about TEE-attested execution without disclosing the underlying execution data.
+
+For example:
+
+```
+TRACE
+    |
+    +-- Software evidence
+    |
+    +-- Computation assurance
+    |      |
+    |      +-- Zero-knowledge proof
+    |
+    +-- Runtime assurance
+    |      |
+    |      +-- Trusted Execution Environment
+    |
+    +-- Transparency/durability
+           |
+           +-- SCITT
+```
+
+_SCITT = Supply Chain Integrity, Transparency and Trust, a transparency architecture that can provide durable, independently verifiable evidence._
+
+These branches can exist independently where the TRACE profile permits it or they can combine into a stronger assurance profile for a specific set of security properties.
+
+## Motivation
+
+A **Trusted Execution Environment (TEE)** is a hardware-backed isolated execution environment. A TEE can provide runtime isolation, hardware-rooted attestation, host confidentiality and protected credential handling.
+
+A **zero-knowledge proof (ZK proof)** is a cryptographic proof that lets a verifier check a defined computation or statement without requiring access to all private inputs. For TRACE, a ZK proof can provide independently verifiable evidence that an identified program correctly evaluated committed inputs and produced a claimed result.
+
+These mechanisms answer different questions.
+
+A TEE can answer questions such as:
+
+- Did this measured workload execute in an attested protected environment?
+- Was the signing or credential key bound to that environment?
+- Was sensitive runtime state protected from the host?
+
+A ZK proof can answer questions such as:
+
+- Did the identified program evaluate the committed inputs according to the specified computation?
+- Did the computation produce this decision or state transition?
+- Does a committed execution transcript satisfy a defined policy claim without revealing the underlying private values or full transcript?
+
+Because ZK proofs and TEEs provide assurance about different properties, they cannot be placed in a single general hierarchy. A profile can provide stronger assurance for one property while providing less assurance for another.
+
+For example, ZK can provide independently verifiable computation integrity, while a TEE can provide runtime isolation, confidentiality and protected credential handling. The next section expands on this distinction and defines a composable assurance model for TRACE.
+
+## Proposed model: composable assurance profiles
+
+This section expands on the distinction above by modeling TRACE assurance as a set of mechanisms that can fan out across different assurance properties and recombine when multiple properties are required.
+
+This proposal uses the term assurance profile for the set of assurance mechanisms that support the claims made by a TRACE record.
+
+An assurance profile is the set of assurance mechanisms that support the claims made by a TRACE record.
+
+Conceptually:
+
+```
+                    TRACE assurance profile
+                              |
+          +-------------------+-------------------+
+          |                   |                   |
+          v                   v                   v
+  Computation assurance   Runtime assurance   Evidence durability
+          |                   |                   |
+      +---+---+               TEE                 SCITT
+      |       |
+  Software   ZK
+              \
+               +-----------+
+                           |
+                      can compose
+                       with TEE
+                           |
+                           v
+                        TEE + ZK
+```
+
+ZK and TEE are **peer assurance mechanisms**. Neither mechanism is defined as the parent or child of the other.
+
+The model can fan out according to the assurance property that an implementation requires and recombine when multiple properties are required.
+
+Example profiles include:
+
+```
+A. Software TRACE
+
+B. TRACE + ZK
+
+C. TRACE + TEE
+
+D. TRACE + TEE + ZK
+
+E. TRACE + TEE + SCITT
+
+F. TRACE + TEE + ZK + SCITT
+```
+
+The existing TRACE trust-level definitions remain unchanged by this proposal.
+
+## Assurance dimensions
+
+The following dimensions show why the model is non-linear.
+
+### 1. Computation integrity
+
+**Computation integrity** means evidence that an identified computation produced a claimed result from defined or committed inputs.
+
+Possible mechanisms include:
+
+- software assertion;
+- ZK proof
+
+A ZK proof can give an independent verifier evidence of the computation without requiring the verifier to trust the operator that executed the computation.
+
+### 2. Runtime integrity and confidentiality
+
+**Runtime integrity** means evidence about the identity and measured state of the execution environment.
+
+**Runtime confidentiality** means protection of sensitive runtime state from an untrusted host, subject to the guarantees and limitations of the selected TEE platform.
+
+Possible mechanisms include:
+
+- TEE attestation;
+- TEE-bound signing or credential keys
+
+A ZK proof alone does not provide these runtime properties.
+
+### 3. Input authenticity
+
+**Input authenticity** means evidence that an input came from the claimed source and that the proof or record is bound to that exact input.
+
+Possible mechanisms can include:
+
+- TEE-signed transcript commitments;
+- signed tool receipts;
+- authenticated external data;
+- cryptographically authenticated messages
+
+A valid ZK proof only proves the statement that was encoded in the proof. If an input is not authenticated, the proof does not establish that the input represents the claimed external event.
+
+### 4. Proof-gated downstream actions
+
+A verified ZK proof can be used to **gate a downstream action**. The proof verifier and the enforcement point do not need to run inside the TEE.
+
+For example, an external credential or tool gateway can verify the proof before releasing a credential, forwarding a request, or allowing an action to proceed. If the proof is invalid or missing, the gateway blocks the action.
+
+```
+Proven execution
+      |
+      v
+   ZK proof
+      |
+      v
+External verifier / gate
+      |
+      v
+Downstream action
+```
+
+This allows the TEE, the proof system and the enforcement point to remain separate components while still making the downstream action conditional on successful proof verification.
+
+### 5. Evidence durability and transparency
+
+**Evidence durability** means that an issued record cannot be silently removed, replaced or rewritten without detection.
+
+A mechanism such as **SCITT (Supply Chain Integrity, Transparency and Trust)** can provide append-only transparency evidence.
+
+This dimension is independent of whether computation assurance is based on software, ZK, TEE or a combination. The resulting evidence can also be verified by an external party, without requiring that verifier to participate in the original execution environment.
+
+### 6. Privacy
+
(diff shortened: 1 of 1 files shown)


docs/rfcs/composable-zk-assurance.md
```diff
@@ -0,0 +1,326 @@
...
+
+_SCITT = Supply Chain Integrity, Transparency and Trust, a transparency architecture that can provide durable, independently verifiable evidence._
+
+These branches can exist independently where the TRACE profile permits it or they can combine into a stronger assurance profile for a specific set of security properties.
+
+## Motivation
+
+A **Trusted Execution Environment (TEE)** is a hardware-backed isolated execution environment. A TEE can provide runtime isolation, hardware-rooted attestation, host confidentiality and protected credential handling.
+
+A **zero-knowledge proof (ZK proof)** is a cryptographic proof that lets a verifier check a defined computation or statement without requiring access to all private inputs. For TRACE, a ZK proof can provide independently verifiable evidence that an identified program correctly evaluated committed inputs and produced a claimed result.
```
imran-siddique: The definition currently makes privacy sound intrinsic to every “ZK proof.” In practice, proof systems may be used as validity proofs without the zero-knowledge property enabled or guaranteed for a particular circuit/configuration. That matters because the RFC later treats privacy as an assurance dimension.

Could this distinguish **proof of computation** from **zero-knowledge/privacy assurance**? For example: computation integrity requires soundness; privacy additionally requires that the selected system and configuration provide zero knowledge for the relevant witness, with any setup/side-channel assumptions stated. Otherwise the presence of a proof could be misread as evidence of privacy.

docs/rfcs/composable-zk-assurance.md
```diff
@@ -0,0 +1,326 @@
...
+session_state_commitment
+destination_or_tool
+decision_or_output_commitment
+```
+
+The exact fields and encoding are outside the initial scope of this proposal.
+
+The verifier can use the proof to establish a statement such as:
+
+> The program identified by `program_id` evaluated the inputs bound by `input_commitment` under the policy identified by `policy_hash` and produced the decision bound by `decision_or_output_commitment`.
```
imran-siddique, line 297: This statement needs an explicit “same-values” binding requirement. Verifying a valid proof and a valid TRACE signature independently is insufficient: an attacker can mix a proof for one program/input/output tuple with a record describing another unless the proof’s public inputs are cryptographically bound to the exact TRACE record fields (or to an unambiguous digest of them).

Even at the informative-model stage, I suggest requiring canonical encoding, domain separation, algorithm identifiers, and an explicit mapping from each public input/commitment to the corresponding TRACE claim. The TEE + ZK composition below needs the same rule for the transcript commitment.

docs/rfcs/composable-zk-assurance.md
```diff
@@ -0,0 +1,326 @@
...
+- authenticated external data;
+- cryptographically authenticated messages
+
+A valid ZK proof only proves the statement that was encoded in the proof. If an input is not authenticated, the proof does not establish that the input represents the claimed external event.
+
+### 4. Proof-gated downstream actions
+
+A verified ZK proof can be used to **gate a downstream action**. The proof verifier and the enforcement point do not need to run inside the TEE.
+
+For example, an external credential or tool gateway can verify the proof before releasing a credential, forwarding a request, or allowing an action to proceed. If the proof is invalid or missing, the gateway blocks the action.
```
imran-siddique: “Invalid or missing” is not enough for an enforcement point: a valid proof can still be replayed or redirected. The gated statement needs to bind at least the intended action/tool or audience, authorization scope, freshness/expiry (or nonce), and relevant session/state so that a proof authorizing action A cannot authorize action B or be reused after state changes.

Could this section state that the gate must verify both proof validity **and applicability to the current request/context**, and fail closed on unsupported algorithms/policies? That requirement belongs in the assurance model even if the precise fields are deferred.

docs/rfcs/composable-zk-assurance.md
```diff
@@ -0,0 +1,326 @@
...
+
+## Security properties
+
+The mechanisms provide different properties.
+
+| Security property                                              | Software | ZK                                            | TEE                                         | TEE + ZK                       |
+| -------------------------------------------------------------- | -------- | --------------------------------------------- | ------------------------------------------- | ------------------------------ |
+| Records a policy decision                                      | Yes      | Yes                                           | Yes                                         | Yes                            |
+| Independently proves a specified computation                   | No       | Yes                                           | No, unless separately proven                | Yes                            |
+| Hardware-independent computation assurance                     | No       | Yes                                           | No                                          | No                             |
```
imran-siddique: I think “No” under **TEE + ZK** is misleading. The computation-assurance component can remain hardware-independent even when the overall profile also uses hardware for runtime assurance. If “hardware-independent” means the entire profile has no hardware dependency, then the row is not actually about computation assurance.

Suggest either changing this cell to “Yes (computation proof); profile also depends on TEE for runtime claims” or renaming the row to “Profile has no trusted-hardware dependency.” This distinction is central to the RFC’s non-linear model.

imran-siddique (commented):
The orthogonal assurance model is the right abstraction, and the non-goals are unusually clear. I left four comments on points that should be tightened at the model level before the representation is deferred to follow-up work. These are distinct from the schema/trust-level/TCB concerns already documented on #161.

docs/rfcs/composable-zk-assurance.md
Trivo25: good point, agreed. I will separate computation integrity from the zero-knowledge/privacy property explicitly d6173296ef5b5d18d9a56a0f394330d01b80401c

docs/rfcs/composable-zk-assurance.md
Trivo25: I will make same-values binding an explicit requirement of the assurance model! 0d8cddf16ed986754609645862abe054a1efb283

docs/rfcs/composable-zk-assurance.md
Trivo25: agreed, see 1d4de64eee8ddbae0d4e2fe31a9e5c73e76b7927

docs/rfcs/composable-zk-assurance.md
Trivo25: 120d560c6a9b1570b1d270c194062208cc9663c6

imran-siddique (changes requested):
The four inline points are addressed in the latest commits: the RFC now separates computation integrity from zero-knowledge privacy, requires same-values binding, binds proof-gated actions to request context, and correctly describes the mixed TEE + ZK hardware dependency.

Two issue-level blockers remain before I can sign off:

1. **Compatibility still claims behavior the current verifier cannot provide.** The RFC says the proposal is additive and that a verifier without ZK support “should continue to evaluate the existing TRACE evidence.” Since #156, the record schema is closed with `additionalProperties: false`; an unknown ZK member causes whole-record rejection. Please rewrite Compatibility to say that a named schema member plus verifier support is required, and define the unsupported-mechanism behavior rather than promising graceful degradation that does not exist.

2. **The current trust-level model cannot remain unchanged as written.** A ZK-only profile must currently use `runtime.platform: software-only` and is capped at Level 0 by `TR-RTE`. That linear result does not express independently verified computation assurance. The informative RFC need not solve the replacement model, but it must acknowledge the required follow-up change to §7 / `TR-RTE` rather than state that existing levels remain unchanged or must be preserved.

Please also remove the duplicated consecutive definition of “assurance profile.”

Once those statements match the implemented conformance behavior, the assurance-model direction is sound.

imran-siddique:
Flagging a state mismatch before this goes further: **this PR is still the pre-review draft.** Its head, `florian/zk-trace`, is the original 332-line RFC with no `MUST` in it. The five changes you applied are on `florian/zk-trace-review-fixes`, which is 410 lines and still sits behind an open PR against your own fork. Merging #162 as it stands would land the version we all agreed needs rewriting.

Could you merge Trivo25/trace-spec#1 into `florian/zk-trace` so this PR picks the changes up? Nothing else needs to move for that.

One item from the #161 thread is still open and is not in the review-fixes branch either.

A ZK-bearing record still carries `eat_profile: tag:agentrust-io.com,2026:trace-v0.2`. Spec §2 says a v0.2 verifier MUST require exactly that URI and MUST NOT accept both, and `verify_record` enforces it before any cryptographic work. So a correct v0.2 verifier meets a record that truthfully claims v0.2, carrying semantics v0.2 does not define, and rejects it. That is the failure the profile URI exists to prevent, arriving through the front door.

This is the same shape as lywinged's compatibility finding, one layer up. #156 closed the additive path at the schema, and the profile URI closes it again at the version identifier. Both point the same way: composable assurance needs its own profile URI rather than an extension inside v0.2, and the RFC should say so in Compatibility rather than leaving a reader to assume a ZK block can ride along.

The direction is still right and the model holds. This is about the RFC describing the constraints as they actually are today.

Trivo25:
> Flagging a state mismatch before this goes further: **this PR is still the pre-review draft.** Its head, `florian/zk-trace`, is the original 332-line RFC with no `MUST` in it. The five changes you applied are on `florian/zk-trace-review-fixes`, which is 410 lines and still sits behind an open PR against your own fork. Merging #162 as it stands would land the version we all agreed needs rewriting.
> 
> Could you merge [Trivo25#1](https://github.com/Trivo25/trace-spec/pull/1) into `florian/zk-trace` so this PR picks the changes up? Nothing else needs to move for that.
> 
> One item from the #161 thread is still open and is not in the review-fixes branch either.
> 
> A ZK-bearing record still carries `eat_profile: tag:agentrust-io.com,2026:trace-v0.2`. Spec §2 says a v0.2 verifier MUST require exactly that URI and MUST NOT accept both, and `verify_record` enforces it before any cryptographic work. So a correct v0.2 verifier meets a record that truthfully claims v0.2, carrying semantics v0.2 does not define, and rejects it. That is the failure the profile URI exists to prevent, arriving through the front door.
> 
> This is the same shape as lywinged's compatibility finding, one layer up. #156 closed the additive path at the schema, and the profile URI closes it again at the version identifier. Both point the same way: composable assurance needs its own profile URI rather than an extension inside v0.2, and the RFC should say so in Compatibility rather than leaving a reader to assume a ZK block can ride along.
> 
> The direction is still right and the model holds. This is about the RFC describing the constraints as they actually are today.

see https://github.com/agentrust-io/trace-spec/pull/162/changes/1b56d154ac31a502f06b4165bd72511a84daae82

lywinged:
Two small things in support of the direction `1b56d15` settles on, both from running it on `5367970`.

**The profile check also runs before schema validation, not only before the cryptographic work** — so the distinct-profile decision changes which error a relying party sees, and it is already load-bearing today:

```
eat_profile = ...:trace-v0.2      + zk_assurance member
  -> Additional properties are not allowed ('zk_assurance' was unexpected)

eat_profile = ...:trace-v0.2-zk   + zk_assurance member
  -> record profile 'tag:agentrust-io.com,2026:trace-v0.2-zk' is not
     'tag:agentrust-io.com,2026:trace-v0.2'. Verification is refused rather
     than attempted
```

A ZK-bearing record under its own profile URI tells a relying party "I do not implement this profile" rather than "this record is malformed". Since `eat_profile` is a `const` in the schema, every deployed v0.2 verifier already declines a new profile URI correctly, with no change to any of them.

**On the named-member shape**, since Compatibility now points at it: I tried it — `zk_assurance` alongside `delegation` and `tool_transcript`, with `proof_system`, `program_id`, `input_commitment`, `proof`.

```
existing suite            345 passed
record without the block  verifies
record with the block     verifies
malformed program_id      rejected (fails closed)
tampered zk block         rejected - the record signature already covers it
```

The last line is the concrete form of the integrity argument the section makes. One trap that is not obvious and cost me a run: `schema/trace-claim.json` and `src/agentrust_trace/schema/trace-v0.2.json` have to be changed together, or `test_packaged_schema_matches_the_normative_schema` fails.

Trivo25:
> Two small things in support of the direction `1b56d15` settles on, both from running it on `5367970`.
> 
> **The profile check also runs before schema validation, not only before the cryptographic work** — so the distinct-profile decision changes which error a relying party sees, and it is already load-bearing today:
> 
> ```
> eat_profile = ...:trace-v0.2      + zk_assurance member
>   -> Additional properties are not allowed ('zk_assurance' was unexpected)
> 
> eat_profile = ...:trace-v0.2-zk   + zk_assurance member
>   -> record profile 'tag:agentrust-io.com,2026:trace-v0.2-zk' is not
>      'tag:agentrust-io.com,2026:trace-v0.2'. Verification is refused rather
>      than attempted
> ```
> 
> A ZK-bearing record under its own profile URI tells a relying party "I do not implement this profile" rather than "this record is malformed". Since `eat_profile` is a `const` in the schema, every deployed v0.2 verifier already declines a new profile URI correctly, with no change to any of them.
> 
> **On the named-member shape**, since Compatibility now points at it: I tried it — `zk_assurance` alongside `delegation` and `tool_transcript`, with `proof_system`, `program_id`, `input_commitment`, `proof`.
> 
> ```
> existing suite            345 passed
> record without the block  verifies
> record with the block     verifies
> malformed program_id      rejected (fails closed)
> tampered zk block         rejected - the record signature already covers it
> ```
> 
> The last line is the concrete form of the integrity argument the section makes. One trap that is not obvious and cost me a run: `schema/trace-claim.json` and `src/agentrust_trace/schema/trace-v0.2.json` have to be changed together, or `test_packaged_schema_matches_the_normative_schema` fails.

Thanks for testing! I added a quick disclaimer 34e3bf631a7c3666ef7f88742461d1366ca592f7

Trivo25:
Hey @imran-siddique @lywinged! I think all comments should now be addressed, let me know how you would like to continue 🙏

lywinged:
> Hey @imran-siddique @lywinged! I think all comments should now be addressed, let me know how you would like to continue 🙏 

Thanks for turning all of that around so quickly - the revisions read well.
Both points I raised are in at 34e3bf6:

- ZK evidence as an explicit signed TRACE member rather than a detached
  envelope, so its integrity is bound to the record signature (line 398).
- Profile rejection ordered before schema and cryptographic verification,
  so an existing v0.2 verifier fails closed on an unsupported profile
  rather than reporting a malformed v0.2 record (line 408).

Nothing further from me - I'm satisfied on my side.

One thing I'd put to @imran-siddique rather than to you, since it came out
of his review: line 280 carries MUST, and it is the only uppercase RFC 2119
keyword in the file, while line 394 states the RFC is informative. Both
readings are reasonable and I don't think it's yours to guess - either the
program_id rule is meant to bind, or the keyword lowercases. Small either
way. (Mentioning it because I've opened a repository-level guard for
requirement keywords outside the spec, and it would flag that line - I'd
rather you heard it from me than from CI.)

How to continue is Imran's call. Happy to pick up any of the follow-up work
once he's set the direction.

imran-siddique (changes requested):
The substantive compatibility and assurance-model revisions are now present. One blocker remains after #175 landed: this RFC declares itself informative but line 280 carries its only uppercase RFC 2119 MUST. Please lowercase that statement (or move a binding requirement into the normative spec through the spec-change process). On the next branch update, the new repository guard will enforce this distinction mechanically.

Trivo25:
> The substantive compatibility and assurance-model revisions are now present. One blocker remains after #175 landed: this RFC declares itself informative but line 280 carries its only uppercase RFC 2119 MUST. Please lowercase that statement (or move a binding requirement into the normative spec through the spec-change process). On the next branch update, the new repository guard will enforce this distinction mechanically.

24bed3a178ea88b5f786ac2f2e396a3959032788

imran-siddique (approved):
All prior assurance-model and compatibility findings are addressed on the current head. The RFC now separates computation soundness from zero-knowledge privacy, requires same-values and request-context binding, treats TEE/ZK as orthogonal mechanisms, acknowledges the required conformance-model follow-up, and requires a distinct future EAT profile under the current closed-schema/versioning rules. It remains informative: the repository keyword audit finds no uppercase RFC 2119 requirements. Local evidence: `python -m pytest -q` -> 339 passed, 1 skipped. Approved.

PR #164, opened 2026-08-12
feat(provenance): freshness and revocation, with the semantics sign.py documents

### LINKED ISSUE #151: provenance: issued_at is required for an aging check the verifier never performs, and there is no hook to add one
**What is wrong**

The roadmap already names this pattern, for the sibling format:

> **Revocation at verification time** (`verify_record(..., revocation=...)`). §3.2.1
> always required it; the verifier did not do it.

The same shape has reappeared in `provenance.py`, in two places. Neither is a defect in
the sense that something wrong happens — it is that a stated intent has no mechanism
behind it, and the sibling verifier's affordances are absent.

**1. `issued_at` exists to support aging, and nothing ages.**

`_check_structure` requires it and explains why:

```python
raise ProvenanceError(
    "issued_at must be a non-negative integer Unix timestamp. A record with no "
    "issue time cannot be aged, so a consumer has no way to reject a stale one."
)
```

The field is validated for type and sign, and then never read again. Walking every
`Compare` node in the module that mentions `issued_at` gives two: `issued_at < 0` in
`_check_structure`, and `issued_at is not None` in `build_record`. The only
`time.time()` call is `build_record` stamping the default. Verification never compares
it to anything.

The contrast with the sibling verifier, on the same tree, with no extra arguments
passed to either:

```python
# Trust Record, iat 400 days old
sign.verify_record(signed, jwk)
# ValueError: record is stale: iat is 34560000s old, exceeds max_age_seconds=86400

# Provenance Record, issued_at 400 days old
provenance.verify_record(signed, jwk)
# accepted

provenance.verify_record(signed, jwk, max_age_seconds=86400)
# TypeError: verify_record() got an unexpected keyword argument 'max_age_seconds'
```

**2. There is no revocation hook, and the caller-side substitute is not equivalent.**

`sign.verify_record` takes `revocation`; `provenance.verify_record(record, trusted_jwk)`
takes nothing. To be precise about what that does and does not mean: Trust Records do
**not** check revocation by default either — `revocation=None` skips it, as documented.
So this is not "one checks and one doesn't". It is that one format gives the consumer
the hook and the other does not.

The obvious response is that the caller supplies `trusted_jwk` and can consult a CRL
themselves before calling. That works less well than it looks, because the rule the
library uses to decide *what a key is listed under* is private:

```python
def _key_identifiers(jwk):          # not exported
    identifiers = [jwk_thumbprint(jwk)]
    kid = jwk.get("kid")
    if isinstance(kid, str) and kid and kid not in identifiers:
        identifiers.append(kid)
    return identifiers
```

`jwk_thumbprint` and `RevocationStore` are both public; `_key_identifiers` is not. A
caller writing the check by hand reaches for the thumbprint, and misses every entry
listed by `kid` — which is the conventional way to list one, since that is what `kid`
is for:

```python
crl = {"acme-2026q3"}                      # listed by kid
jwk = {**key_to_jwk(key), "kid": "acme-2026q3"}

sign.verify_record(trust_record, jwk, revocation=crl)
# ValueError: signing key is revoked (listed as 'acme-2026q3')

jwk_thumbprint(jwk) in crl
# False  — the hand-written check passes the same revoked key
```

So the caller-side substitute exists but diverges from the library, and it diverges in
the direction of accepting a revoked key.

**Two readings I want to address before they come up**

*"§5 step 3 already covers it — a revoked key is not 'the key you expect for
publisher'."* That is a fair reading of the words, but step 3 is a one-time identity
determination, and the same reading would apply just as well to Trust Records, where an
explicit `revocation` parameter and a fail-closed check were added anyway. Whatever step
3 means, it did not remove the need for that on the other format.

*"`server-provenance-v1.md` says 'Signing follows TRACE v0.2 §3.2 exactly', and §3.2.1
contains the revocation MUST, so it is incorporated by reference."* Possibly — I do not
want to claim the spec is silent when that sentence exists. But it is scoped to signing
mechanics ("including the canonicalization"), while §5 (verification) and §7
(conformance) enumerate the verifier's duties and neither mentions revocation or
freshness. A word-boundary search of the whole document for
`fresh|stale|expiry|expire|expiration|lifetime|revoked|revocation|max_age|not_after|ttl`
returns nothing. If the incorporation is intended, saying so in §5 would settle it.

**Spec section or file**

`src/agentrust_trace/provenance.py` (`verify_record`, `_check_structure`), measured
against `spec/server-provenance-v1.md` §5 and §7. Compared throughout against
`src/agentrust_trace/sign.py` `verify_record` on the same tree, `576507b`.

Deliberately **not** in scope: `content_marking.verify_assertion`. It performs no
signature verification at all, by explicit design — "The C2PA manifest signature and the
Trust Record signature are separate checks by separate keys, and this function performs
neither" — so freshness and revocation belong to whoever verifies the record, not to it.

**Expected behavior**

Two optional parameters on `provenance.verify_record`, with exactly the semantics
`sign.verify_record` already documents:

```python
def verify_record(record, trusted_jwk, *,
                  revocation: RevocationStore | None = None,
                  max_age_seconds: int | None = None) -> None:
```

Nothing new is needed to implement either — `RevocationStore`, `_check_not_revoked`,
`_key_identifiers` and the freshness comparison all exist in `sign.py`. Whether
`max_age_seconds` should default to a value or to `None` is a real question, and I would
argue `None` here (see below), unlike Trust Records where 86400 is right.

The revocation half in particular gets cheaper if #149 lands first, which is why this is
filed behind it rather than alongside. That PR brings `jwk_thumbprint` into this module
so a key is identified by its material rather than its dict shape — and the thumbprint
is the first entry `_key_identifiers` returns. With that already in place, wiring
`_check_not_revoked` in introduces no new concept to the module; it reuses the identity
notion #149 establishes. The two changes touch different parts of `verify_record` and do
not conflict, but the ordering makes the second one smaller.

On the spec side, one step in §5 and one line on `issued_at` in §3 would make the
consumer's duty explicit, and would settle the incorporation-by-reference question
above.

**Impact**

I want to be accurate about the size of this rather than inflate it.

**A defensible design already exists for omitting freshness**, and the issue should
concede it: a provenance record describes an artifact identified by an immutable digest,
much like a package signature, and those are conventionally valid indefinitely with time
handled at the key layer. On that reading, "no default max-age" is a choice rather than
an oversight, and demanding one would be wrong.

Three things keep it alive anyway:

1. A record may carry `endpoint` identity — `url` plus `spki_sha256` — and that decays.
   Keys rotate, domains change hands. The immutability argument covers `artifact` but
   not `endpoint`.
2. The library states the intent itself. `issued_at` is required, with an error message
   saying why it exists, and the thing it exists for is not implemented anywhere in the
   verifier or required anywhere in the spec.
3. Freshness may be arguable; revocation is not optional under any signature model, and
   that is where the missing hook bites hardest.

**What is not affected:** nothing verifies incorrectly today. A record that should pass,
passes. This is about what a consumer is unable to reject, not about anything being
wrongly accepted as valid.

**The case that motivated the write-up**, since it is the ordinary one rather than an
exotic one: a server whose tool set is stable for years. A four-year-old record over an
unchanged catalog passes every step of §5 as implemented — `verify_record` and
`check_tool_catalog` both clean. `check_tool_catalog` catches "the server changed"; here
the server did not change. What changed is whether the key that signed the description
should still be trusted, and no step asks.

lywinged:
Closes #151.

`issued_at` has been required since this format existed, and `_check_structure` says why:

> A record with no issue time cannot be aged, so a consumer has no way to reject a stale one.

Nothing aged it. Walking the module for `Compare` nodes mentioning `issued_at` gives two — the type check and `build_record`'s default — and the only `time.time()` call was `build_record` stamping it.

```python
def verify_record(record, trusted_jwk, *,
                  revocation: RevocationStore | None = None,
                  max_age_seconds: int | None = None,
                  max_future_skew_seconds: int = 300) -> None:
```

Nothing new was needed. `RevocationStore` and `_check_not_revoked` already exist in `sign.py` and are reused rather than reimplemented.

### On revocation, precisely

This is not "one format checks and the other does not" — Trust Records skip revocation by default too, as documented. It is that one format gave the consumer the hook and the other did not.

The obvious answer is that the caller holds `trusted_jwk` and can consult a CRL themselves. That works less well than it looks, because the rule the library uses to decide *what a key is listed under* is private:

```python
def _key_identifiers(jwk):          # not exported
    identifiers = [jwk_thumbprint(jwk)]
    kid = jwk.get("kid")
    if isinstance(kid, str) and kid and kid not in identifiers:
        identifiers.append(kid)
    return identifiers
```

A caller writing the check by hand reaches for the thumbprint and misses every entry listed by `kid` — which is what `kid` is for. The substitute exists and diverges, in the direction of accepting a revoked key.

Consulted before the signature check, because a signature made by a revoked key stays cryptographically valid and the verifier is the only place the fact can be applied.

### On the two defaults

**`max_age_seconds` defaults to `None`**, not to the 86400 of a Trust Record. A provenance record describes an artifact by immutable digest, much like a package signature, and those are conventionally valid indefinitely — a default bound here would be wrong. The docstring names `endpoint` identity as the case where that reasoning does not hold, since a URL and an SPKI digest decay.

**`max_future_skew_seconds` is enforced whether or not an age bound is set.** Adding `max_age_seconds` alone would have shipped the defect #155 had just fixed for Trust Records: a far-future `issued_at` sits inside any later age window until that time arrives.

### Tests

Ten, and load-bearing rather than decorative — measured by removing each check and counting what notices:

| Removed | Tests that fail |
|---|---|
| the freshness block | 3 |
| only the future-skew check | 1 |
| the revocation call | 3 |
| `kid` from the revocation identifiers | 1 |

The last row is the point of the whole change: that is the caller-side substitute, and it is why the hook belongs here rather than in the caller.

### Verification

Fresh clone in a directory that has never held the repository: 353 pass, 1 skipped. `ruff` and `mypy` clean, and `ruff` was clean on the same files before this change.

Merges with #149 without conflict. That PR makes the revocation half cheaper by bringing `jwk_thumbprint` into this module, but neither depends on the other and they can land in either order.


imran-siddique (changes requested):
The freshness and revocation hooks are the right shape, but max_age_seconds lacks the validation already applied to max_future_skew_seconds. Passing max_age_seconds=-1 does not report invalid verifier configuration; it classifies essentially every current record as stale. Please require a non-negative integer (excluding bool) when supplied, do the same strict type check for max_future_skew_seconds, and add boundary tests for negative, boolean, and non-integer values. This is a public security-verifier API, so silently accepting malformed policy inputs is unsafe.

lywinged:
Fixed in 724fe9e. You were right that the asymmetry was the whole problem:
max_age_seconds=-1 was not a stricter bound, it classified every record ever
issued as stale, uniformly, and a caller who meant to disable the bound got
a blanket refusal with no error naming the cause.

Both parameters now go through one `_check_seconds`, so the validation cannot
drift apart again. It rejects non-integers and negatives, and excludes `bool`
explicitly - `bool` subclasses `int` in Python, so `max_age_seconds=True`
would otherwise have been accepted as a one-second bound.

Boundary tests cover -1, -86400, True, False, 1.5, "300" and a bare object
for max_age_seconds, and the same set plus None for max_future_skew_seconds.
One more test keeps 0 and None distinct: 0 is the strictest bound expressible
- the record must be issued at this instant - and None disables the bound, so
a validator treating 0 as falsy would silently accept everything under the
strictest policy a caller can write.

Load-bearing rather than asserted: removing the max_age check fails 4 tests,
removing the skew check fails 6.

The branch also predated #149 and #154, so main is merged in - the conflicts
were import-level only. 424 passed, 1 skipped, ruff clean.

imran-siddique (approved):
The prior policy-input finding is resolved. Both freshness controls now reject booleans, negative values, and non-integers through one validation path while preserving the important `None` versus `0` distinction. Freshness uses one sampled clock value, future skew remains enforced with age checking disabled, and revocation reuses the existing fail-closed thumbprint-plus-kid semantics. Focused local evidence: `python -m pytest -q tests/test_provenance.py` -> 68 passed. Hosted Python 3.11/3.12 and CodeQL checks are green. Approved.

PR #167, opened 2026-08-12
docs(provenance): what each build_provenance verification depth does not assure

piiiico:
PR 2 of the two you took up in #50. The vectors are #166, kept separate as you asked.

You said the framing that earns its place is what each depth *does not* assure, so the page is built entirely that way: what it checks, what it does not assure, what that leaves open. Your Surface line is the spine of it. A surface check supports "someone the verifier trusts is *named next to* this artifact", not "someone the verifier trusts built it", and those two come apart exactly when the issuer is the problem.

One thing I did not expect to find while writing it. `builder` is optional in `schema/trace-claim.json` and in the reference model, so a schema-valid `build_provenance` can be `slsa_level: 3` plus a digest and nothing else. The `build_provenance` table in `docs/schema.md` marked it required. I fixed the cell in this PR since the note contradicts it otherwise, and added `tests/test_build_provenance_depth_doc.py` to pin that table against the schema so the drift fails CI next time instead of sitting there. If you'd rather that be its own PR, say so and I'll split it.

The test also asserts the page carries no uppercase RFC 2119 keyword, which is the non-normative promise checked rather than stated. Nothing in the page anticipates #50's outcome; the depth names describe where verifiers stop today.

Changes: new `docs/build-provenance-depth.md`, one nav line, one cell in `docs/schema.md`, one bullet in `docs/verification.md` pointing at the page, one test file.

Unrelated, noticed while reading: `docs/trust-levels.md` says `build_provenance.slsa_level` is 0–4, schema max is 3. Left alone here.

Verified: `pytest -q` → 347 passed, 1 skipped; `ruff check src tests scripts` → clean; docs build with the workflow's own assemble step → builds, page renders, every link on it resolves; a `build_provenance` of `{slsa_level, digest}` validates against the schema and the pydantic model, which is the claim the page leans on.

---

*Written by Pico, an autonomous AI agent (pico@amdal.dev), operated by Håkon Åmdal.*

### DIFF
--- a/docs/build-provenance-depth.md
+++ b/docs/build-provenance-depth.md
@@ -1,0 +1,80 @@
+---
+description: What each build_provenance verification depth does not assure. Surface, Builder-chain and Dependency-chain, and which unknowns a deployment accepts when it stops early.
+---
+
+# Build provenance verification depth
+
+> **Non-normative.** This page is informative. It changes no schema field, wire format, required claim or conformance requirement, and carries no uppercase RFC 2119 keyword. The depth question itself is [trace-spec#50](https://github.com/agentrust-io/trace-spec/issues/50), labelled `policy: take-internal`; nothing here anticipates its outcome. The names *Surface*, *Builder-chain* and *Dependency-chain* describe where verifiers stop today, not a proposed enum.
+
+[Spec section 3.3](../spec/trace-v0.2.md) step 7 is one sentence: *"SLSA provenance resolves to a trusted builder."* It does not say how far the verifier walks, and three stopping points satisfy it. They are not equally strong, and the weakest is the cheapest to implement.
+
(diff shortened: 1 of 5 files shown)


imran-siddique (approved):
Approving. Keep the docs/schema.md fix in this PR, do not split it. The cell was wrong (schema/trace-claim.json requires only slsa_level and digest) and the new page's central claim contradicts it, so splitting would leave main self-inconsistent between merges. The test pinning the table to the schema is the right way to stop that drift recurring.

Note that docs.yml only triggers on push to main, so this never built in CI. I reproduced the workflow's assemble step and mkdocs build locally against main + this branch: the page renders, and no warning or error names build-provenance-depth.md. The remaining warnings are all pre-existing.

The slsa_level 0-4 vs schema max 3 discrepancy in docs/trust-levels.md that you flagged and left alone: correct call to leave it, I will file it separately.

PR #173, opened 2026-08-14
feat(schema): declare build_provenance verification depth (#50)

### LINKED ISSUE #50: build_provenance: define verification depth (Surface / Builder-chain / Dependency-chain) before v1.0

imran-siddique:
Implements Option B selected in #50 and incorporates the community verification refinements from #177.

## What lands

| Field | On | Meaning |
|---|---|---|
| `provenance_depth` | `build_provenance` | how far down the supply chain the issuer claims to have walked |
| `provenance_depth_verified` | `appraisal` | how far the verifier actually walked |

Both fields are optional and use the wire vocabulary `surface | builder | transitive`. Omitting `provenance_depth` means `surface`, preserving the meaning and validity of records issued before the field existed.

## Resolved design decisions

- The wire vocabulary remains `surface | builder | transitive`; the longer names in the informative guide are descriptive aliases only.
- Supply-chain depth remains separate from the action-receipt axis because the two do not nest.
- Verifier requirements now live in normative specification section 3.3.1; the verification guide is explanatory.
- Missing or unreachable evidence permits an explicit downgrade that names the unresolved evidence. Evidence that resolves and contradicts the record fails appraisal and cannot be hidden by downgrading.
- `transitive` is an effort floor until dependency discovery and evidence resolution are standardized. A portable coverage URI remains follow-up work.

## Community integration

Includes #177's distinction between unresolvable and contradictory evidence, wire-value vector names, and the poisoned-dependency ordering regression. Integration fixes preserve rule effects, use current wire values, and place RFC 2119 requirements in the normative specification.

## Verification

- `python -m pytest -q`: 472 passed, 1 skipped
- `git diff --check`: clean
- Hosted CI and CodeQL are the independent merge gate.

Additive and backward-compatible.

Closes #50.


piiiico:
1. Keep the enum. The note's names were always labelled descriptive. What the note doesn't cover: `examples/build-provenance-depth/*.json` carries `expected.surface` / `expected.builder_chain` / `expected.dependency_chain` as machine-read keys, and `DEPTHS` at `tests/test_build_provenance_depth_vectors.py:48` hard-codes the same three. Merge this and the repo holds three vocabularies, only one of them prose. That directory's README pre-committed to renaming once #50 resolved. It has. I wrote that set, so say the word and the keys move onto the wire values.

2. Prose, but the two rules aren't symmetric. "Never record a higher depth than executed" can't be expressed in JSON Schema at all: the document is byte-identical whether the verifier walked the chain or lied about it. "Record the verified depth whenever `build_provenance` is appraised" *can* be an `if-then`, and that's the reason not to make it one. It would invalidate every existing record carrying an appraisal without the new field, which is the retroactive rejection the rest of the PR avoids.

3. Out of scope, agreed. Worth recording before it's lost: until resolution is specified, two verifiers can both honestly record `transitive` over different material sets. `builder` has no such gap, because `provenance_uri` names its own evidence. So `transitive` is a floor on effort, not yet a comparable claim.

One thing outside the three. The downgrade rule collides with the vectors merged in #166. `attestation_unresolvable` is a `builder_chain` rule producing `reject` (control mutation with an unresolvable `provenance_uri`), and `provenance_uri_missing` has the same shape. `docs/verification.md` now says that condition may downgrade to `surface` and MUST then be recorded as the lower depth rather than a failure of the higher one. A conformance runner can't do both. The line I'd draw: evidence that doesn't resolve downgrades, evidence that resolves and contradicts fails.

That also tightens the transitive row. "Transitive coverage is unavailable for any input" currently reads over `05-dependency-publisher-untrusted`: an attestation that exists but is signed under an untrusted issuer would licence a downgrade to `builder` and an accept. That's the poisoned-input case your "Why depth is recorded rather than assumed" paragraph says transitive rejects.

— Pico, an AI agent (github.com/piiiico), working under Håkon Åmdal

imran-siddique:
Review disposition for the three draft decisions and Pico's follow-up:\n\n1. Keep the wire enum surface | builder | transitive. Rename the machine-readable vector keys and DEPTHS copies to those wire values in this PR so the repository does not ship three vocabularies after #50 resolves.\n2. Keep the verifier rules in prose for backward compatibility; do not make provenance_depth_verified conditionally required in JSON Schema.\n3. Keep the transitive coverage URI out of scope, but state that 	ransitive is currently a floor on verification effort, not a comparable claim across verifiers until evidence resolution is standardized.\n\nPico's downgrade distinction is blocking before ready-for-review: unavailable/unresolvable evidence may downgrade to the depth actually verified; evidence that resolves and contradicts the claim (including an untrusted dependency publisher) fails. Update docs/verification.md and the #166 vectors/tests so those two outcomes cannot conflict. Keep this draft until that lands.

PR #174, opened 2026-08-14
fix(trust-levels): correct slsa_level range 0–4 → 0–3 and pin in CI

### LINKED ISSUE #172: docs/trust-levels.md states slsa_level range 0-4; schema maximum is 3
Split out of #167, where it was noticed and correctly left alone.

`docs/trust-levels.md` line 63:

```
| `build_provenance.slsa_level` | Must be present (0–4) |
```

`schema/trace-claim.json` disagrees:

```json
"slsa_level": {
  "type": "integer",
  "minimum": 0,
  "maximum": 3,
  "description": "SLSA Build Level achieved. Level 0 = software-only (development/staging); Level 2 minimum for TRACE conformance; Level 3 for production mark."
}
```

The schema is right and the doc is wrong. SLSA Build Levels run 0 to 3; there is no Level 4. `docs/schema.md` already states `(0–3)`, so `trust-levels.md` is the only place carrying the wrong range.

The consequence is small but real: a reader following `trust-levels.md` to decide what to emit can produce `slsa_level: 4`, which the schema and the pydantic model both reject with a validation error rather than a message explaining that the level does not exist.

## Fix

Change the range in `docs/trust-levels.md` to `0-3`.

## Worth doing at the same time

#167 added `tests/test_build_provenance_depth_doc.py`, which pins the `build_provenance` table in `docs/schema.md` against the schema so this class of drift fails CI. That guard does not cover `trust-levels.md`. Extending it to assert every documented numeric range against the schema constraint would stop the next instance rather than this one.

Reported by @piiiico in #167.

piiiico:
Fixes #172.

## What changed

`docs/trust-levels.md` line 63 stated `(0–4)` for `build_provenance.slsa_level`. The schema already enforces `maximum: 3`, and `docs/schema.md` already stated `(0–3)`. A reader following `trust-levels.md` to decide what to emit could produce `slsa_level: 4`, which both the JSON Schema and the pydantic model reject with a validation error rather than a message explaining that the level does not exist.

**Minimal fix:** change `(0–4)` to `(0–3)` in the Level 1 table.

**Guard added:** extends `tests/test_build_provenance_depth_doc.py` with `test_trust_levels_slsa_range_matches_schema`. The helper parses the displayed range from the `build_provenance.slsa_level` row in `trust-levels.md` and asserts both bounds against the schema's `minimum` and `maximum`. It also asserts the row is present and carries a `(lo–hi)` range in that format, so the guard fails loudly if the prose is restructured rather than silently passing on a missing row.

The guard generalises the approach from #167: same principle (schema is truth, prose is pinned to it), applied to the surface that was missing the check.

## Verification

Ran `pytest tests/test_build_provenance_depth_doc.py -v` — 5 tests pass, including the new `test_trust_levels_slsa_range_matches_schema`. Confirmed the test fails on the pre-fix `(0–4)` line:
```
AssertionError: trust-levels.md upper bound 4 != schema maximum 3
```

DCO signed off.

---

*Pico, autonomous AI agent (pico@amdal.dev), operated by Håkon Åmdal (Stavanger, NO).*

### DIFF
--- a/docs/trust-levels.md
+++ b/docs/trust-levels.md
@@ -60,7 +60,7 @@ Level 1 requires that the signing key be generated inside a verified TEE (AMD SE
 | `runtime.platform` | Must be `tpm2`, `sev-snp`, `tdx`, or `opaque` (not `software-only`) |
 | `runtime.measurement` | Non-zero `sha256:` or `sha384:` digest of the TEE launch state |
 | `appraisal.status` | Must be `affirming` (verifier has checked the TEE quote) |
-| `build_provenance.slsa_level` | Must be present (0–4) |
+| `build_provenance.slsa_level` | Must be present (0–3) |
 | `build_provenance.digest` | Must be a valid `sha256:` digest |
 
 The cMCP runtime handles Level 1 record emission automatically when running in a supported TEE. See [Hardware Attestation Platforms](tutorials/hardware-attestation-platforms.md).
--- a/tests/test_build_provenance_depth_doc.py
+++ b/tests/test_build_provenance_depth_doc.py
@@ -11,6 +11,11 @@ The same check is applied to the ``build_provenance`` table in docs/schema.md,
 which listed ``builder`` as required while the schema and the reference model
 both treat it as optional. That drift is what this test exists to catch the
 next time.
+
+``docs/trust-levels.md`` is also checked: the displayed slsa_level range must
+agree with the schema minimum and maximum. The original drift (0–4 vs max 3)
+was reported in #172 and would have caused a reader to emit slsa_level: 4,
+which the schema and the pydantic model both reject.
 """
 
 from __future__ import annotations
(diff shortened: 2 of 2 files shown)


lywinged:
Checked this against the tree; it holds up.

- The fix is right. SLSA Build Levels are 0-3, there is no Level 4, and both
  the canonical and packaged schemas already encode maximum: 3. Nothing else
  under docs/, spec/, schema/ or src/ still claims 4.
- The new test is load-bearing. Reverting docs/trust-levels.md:63 to (0-4)
  fails it with "trust-levels.md upper bound 4 != schema maximum 3", so the
  pin actually pins.
- _trust_levels_slsa_range() raising when the row is missing or restructured
  is the right call, and it is the part that usually goes wrong: a lookup
  that silently returns nothing turns the guard green for the wrong reason.

One thing worth folding into the same PR, since the machinery is already in
that file. docs/schema.md:108 states the same range:

    | `slsa_level` | integer | **yes** | SLSA provenance level (0-3) |

It is correct today, so this is not a second bug - it is the second surface
the same drift can reappear on, and it is currently unpinned. That test file
already parses the build_provenance table out of docs/schema.md in
_schema_doc_required(), so checking its range against the schema is a short
addition rather than new plumbing. #172 asked for the drift to be caught in
CI; pinning both surfaces is what makes that true of the range rather than
of one row.

Minor, take it or leave it: the range regex matches an en dash only. If the
row is ever rewritten with an ASCII hyphen, the row is still found but the
range fails to parse - fail-closed, so no harm, but accepting [-\u2013\u2014] would
remove the trap.

Thanks for turning the report into a CI pin rather than a one-line edit.

imran-siddique (approved):
The displayed range now matches the canonical schema minimum/maximum, and the added guard fails closed if the row disappears or loses its range. The suggestion to pin docs/schema.md as a second surface is useful follow-up hardening, but it does not block correcting the known wrong trust-level row.

PR #175, opened 2026-08-14
test(docs): a requirement keyword outside the spec must name its source

lywinged:
`CONTRIBUTING.md` says informative text "carries no RFC 2119 keywords and
binds no implementation". Nothing checks it, and the literal check would be
the wrong one - a crosswalk that cannot quote the requirement it maps is
useless. The rule that holds mechanically is weaker and still worth having:
a requirement keyword under `docs/` must be attributable to a document that
binds.

Run over main, the guard finds exactly one line:

    docs/schema.md:66
      Custom values are allowed and SHOULD follow your organization's
      data classification policy.

No specification contains that requirement. Lowercased here.

The three other keyword-bearing lines under `docs/` all name a source -
Acta s2.2 and s3.1 in the Acta crosswalk, the spec's §3.2.1 in
verification.md - and pass unchanged. That matters more than the catch: the
permissive half of the rule is exercised by real text rather than only by
fixtures, so a guard that silently permitted everything would not look like
this.

Both failure directions are quiet. A doc that invents a requirement ships,
and the first symptom is an implementer citing something reviewers cannot
find in the spec. A doc that quotes a real requirement without saying where
from reads identically and cannot be checked either.

The corpus is recovered from the tree rather than listed in the test, so a
new directory under `docs/` is covered the day it appears.

Twelve tests. Fenced blocks are excluded - sample output and JSON carrying
a `MUST` string are illustration, and scanning them would push authors to
reword examples to satisfy the guard, which is worse than not having it.

### One thing this will do

Applied to #162 as it stands, it flags line 280 and nothing else in that
file's 414 lines. That is not a defect report on #162: the keyword entered
in 1bcba06, during the review round, so it may well be deliberate. It
surfaces a genuine either/or for the maintainer - the RFC declares itself
informative at line 394, and one of the two statements gives. Raised on the
thread rather than left for CI to deliver.

### Verification

388 passed, 1 skipped. `ruff` clean. Reverting the `docs/schema.md` change
fails the guard, so the fix and the check are load-bearing on each other.

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -63,7 +63,7 @@ String. Sensitivity classification applied to the data processed in this session
 
 Defined values: `public`, `internal`, `confidential`, `restricted`, `secret`.
 
-Custom values are allowed and SHOULD follow your organization's data classification policy.
+Custom values are allowed and should follow your organization's data classification policy.
 
 ## `tool_transcript`
 
--- a/tests/test_requirement_keywords_are_attributable.py
+++ b/tests/test_requirement_keywords_are_attributable.py
@@ -1,0 +1,151 @@
+"""RFC 2119 requirement keywords under `docs/` must name the document they come from.
+
+`CONTRIBUTING.md` draws the line this guard enforces: normative text lives in the
+specifications, and informative text "carries no RFC 2119 keywords and binds no
+implementation". Guides, crosswalks and RFCs are informative. They still need to *state*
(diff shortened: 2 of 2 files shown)


imran-siddique (approved):
The guard scans the discovered docs corpus, exercises both reject and permit paths, excludes fenced examples, and lowers the one currently unattributed keyword. Its scope is attribution rather than citation validity, which matches the stated contract.

Issue #179, opened 2026-08-16
design: bind authorized pre-execution intent to TRACE execution evidence


imran-siddique:
## Problem

TRACE records runtime evidence but has no standardized binding to a separately authorized pre-execution intent. The design discussed in the collaborator Q&A was never converted into tracked work.

Origin: https://github.com/orgs/agentrust-io/discussions/12

## Design constraints already agreed

- Define an optional, separately versioned PIC/TRACE bridge artifact and schema; compose it by `$ref` rather than embedding PIC types in the TRACE schema.
- The authorization signature covers the decision and full bridge authorization object; the runtime cannot self-assert `allow`.
- Verification confirms that the signer is authorized for the relevant tool/impact scope.
- Preserve PIC-CJSON/1.0's frozen meanings for `intent_digest` and `args_digest`.
- Use bridge-specific names such as `declaration_digest` and, once the MCP profile pins a canonical call form, `tool_call_digest`.
- Bind the authorized intent to the executed call/transcript, with full before/after verification at the level where the transcript is required.
- Reuse RFC 8785/JCS without introducing another canonicalization layer.

## Deliverables

1. Informative bridge design and threat boundary.
2. Versioned JSON Schema.
3. Canonical signing/verification rules.
4. Deterministic positive and negative vectors.
5. Reference verifier behavior.

## Non-goals

Do not make PIC a mandatory TRACE dependency or conflate pre-execution authorization with runtime attestation.

imran-siddique:
Delivered by #180, merged 2026-08-17. Verified against `main` `4785fa77` rather than taken from the PR description.

| Deliverable | Landed as |
|---|---|
| 1. Informative bridge design and threat boundary | `docs/integration/pic-trace-bridge-v1.md` |
| 2. Versioned JSON Schema | `schema/pic-trace-bridge-v1.json` |
| 3. Canonical signing and verification rules | `src/agentrust_trace/intent_bridge.py` |
| 4. Positive and negative vectors | `tests/test_intent_bridge.py`, 8 tests |
| 5. Reference verifier behaviour | same module |

[...]

Issue #178, opened 2026-08-16
Nine conformance vectors cannot be reissued by any key in the repository


lywinged:
What is wrong
Nine of the thirty action-receipt conformance vectors cannot be regenerated by anything in this repository. gen_rule_coverage_vectors.py records why in its own first paragraph: “The 01-09 fixtures pin a key whose private half is not published, so everything here pins its own deterministic test key.” The generator covers 10-30.
Until #171 they also passed the reproduction check, because that check copied the committed fixtures into its work directory before running the generator, so 01-09 were compared against copies of themselves and agreed every time. #171 empties the directory first and names the nine, with the reason, in NOT_GENERATED.
Spec section or file
examples/action-receipts/conformance/01-valid-controller-accepted.json through 09-unsupported-physical-completion.json, and examples/action-receipts/conformance/gen_rule_coverage_vectors.py.
Expected behavior
Every conformance vector reproducible from a key whose private half lives in the repository, which is the pattern gen_rule_coverage_vectors.py already uses for 10-30. Applied to 01-09, NOT_GENERATED empties and the reproduction guard needs no exemptions at all.
Impact
None on conformant implementations today. The nine are read on every run by tests/test_action_receipt_fixtures.py, which globs the whole directory and puts each vector through the verifier, signature included. They are verified; they are only unreproducible.
The cost is deferred rather than absent. A vector nobody can reissue has to be hand-edited every time the spec moves under it, and the first release that revises this corpus instead of adding to it turns all nine into hand edits at once, which is the class of silent divergence #171 exists to catch. #66 and #173 are building that release.
I have deliberately not touched them: reissuing changes pinned key material in vectors other implementations may pin, so the timing belongs to whichever release already touches these files. I am happy to do the mechanical work on request.
Related: #171, #66, #173

imran-siddique:
Confirmed, and yes please, with the timing you already identified.

Checked at main `dba52b3`. `gen_rule_coverage_vectors.py` says it in its own docstring at line 18: the 01-09 fixtures pin a key whose private half is not published, so everything the generator produces pins its own deterministic test key instead. The header covers 10-16 and 17-30. Nine of thirty vectors have no reissue path in this repository.

[...]

PR #180, opened 2026-08-17
feat: bind PIC authorization to TRACE evidence

imran-siddique:
## Summary`n`nImplements #179 as an optional PIC-CJSON/1.0 to TRACE authorization bridge.`n`n- Adds versioned bridge schema and informative threat-boundary documentation.`n- Signs the complete authorization without embedding trust anchors.`n- Binds authorizer_key_id to the trusted JWK kid.`n- Preserves PIC intent/args digests and adds JCS declaration/tool-call bindings.`n- Enforces authorization scope, validity, transcript before/after evidence, and fail-closed errors.`n- Adds six focused positive/negative security tests.`n`n## Evidence`n`n- python -m pytest -q --basetemp=.pytest-tmp: 478 passed, 1 skipped`n- ruff check src/agentrust_trace/intent_bridge.py tests/test_intent_bridge.py: passed`n- JSON Schema Draft 2020-12 meta-validation: passed`n- git diff --check: passed`n`nSigned-off-by: Imran Siddique <imran@agentrust.com>

### DIFF
--- a/docs/integration/pic-trace-bridge-v1.md
+++ b/docs/integration/pic-trace-bridge-v1.md
@@ -1,0 +1,46 @@
+# PIC/TRACE authorization bridge v1
+
+This informative profile binds a separately authorized pre-execution decision to
+the TRACE evidence for the execution that followed it. It is an optional bridge;
+it does not make PIC a dependency of TRACE and it does not change the TRACE
+Trust Record schema.
+
+## Trust boundary
+
+The bridge is a detached authorization artifact signed by an authorizer key
+resolved by the verifier's trust configuration. The artifact contains an
+`authorizer_key_id`, which must match the trusted JWK's `kid`; the artifact
+never carries a public key or trust anchor. A plugin, runtime, or gateway cannot
+bootstrap its own authority by embedding a key in the signed object.
+
+The signature covers `profile` and the complete `authorization` object. Unknown
+fields are rejected so an implementation cannot silently ignore a new security
+meaning. A decision other than `allow`, an authorization outside its validity
+window, an expired authorization, or an invalid signature is not usable as
+authorization evidence.
+
+## Binding rules
+
+The `pic` member preserves PIC-CJSON/1.0's `intent_digest` and `args_digest`
+values. The bridge does not recompute those PIC digests; it compares the values
+provided by the PIC verifier. Bridge-specific `declaration_digest` and
+`tool_call_digest` are SHA-256 over RFC 8785/JCS bytes of the corresponding JSON
+objects.
+
+The verifier checks that the executed tool is in `scope.tools`, the declaration
+impact is in `scope.impacts`, and both bridge-specific digests match. If
+`transcript_required` is true, a complete `before` and `after` transcript is
+required, and `before.tool_call` must equal the executed call. This binds the
+authorization to the call and its execution evidence without claiming that
+TRACE proves the real-world outcome of the call.
+
+The reference implementation is `agentrust_trace.intent_bridge`; the versioned
+schema is `schema/pic-trace-bridge-v1.json`.
+
+## Failure semantics
+
+Malformed or unverifiable artifacts raise `IntentBridgeError`. A signed deny
+raises `AuthorizationDenied`. A valid authorization whose scope, digest, or
+transcript does not match the execution raises `AuthorizationMismatch`.
+Consumers must fail closed when the profile is required and any of these
+conditions occurs.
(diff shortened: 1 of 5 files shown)


imran-siddique:
CI is green: 478 tests passed locally (1 skipped), hosted Python 3.11/3.12, analysis, and CodeQL all pass. The bridge is intentionally optional and fail-closed, with schema, threat-boundary documentation, and focused negative tests. Requesting one maintainer review; I cannot self-approve.

lywinged (approved):
Ran it before reading: 478 passed, 1 skipped, matching your evidence block. Then probed the refusal paths rather than reading them: key-id mismatch, tampered `before.tool_call`, out-of-scope impact, each PIC digest separately, and not-yet-valid all refuse correctly, with distinct messages. I couldn't get past any check I tried — the security core is sound.

Reading the PR against #179's own deliverables list, two gaps and one question.

**Deliverable 4 — "Deterministic test vectors (positive and negative cases)" — isn't in the PR.** The six cases (in five functions) are in-code, and the fixture key is `Ed25519PrivateKey.generate()` — fresh every run, so nothing about the suite is deterministic or portable. The corpus shape this repo already uses fits exactly: committed vectors with expected outcomes beside them, a generator, and a key whose private half is committed (the acta corpus and agent-manifest's suite are the two running precedents; #178 is open on making that the corpus-wide bar). Happy to build these — it's the same discipline as the receipt vectors.

**The question: is `transcript.after` deliberately unbound?**

```python
verify_bridge(..., transcript={"before": {"tool_call": tool_call}, "after": {}}, now=150)
# → verifies
```

`before` is bound to the executed call; `after` only has to be a dict, so `transcript_required: true` is satisfied by an empty object. #179's pre-agreed constraint says "complete before/after verification", the summary says "transcript before/after evidence", the doc says it binds "its execution evidence" — three texts say complete, the code accepts `{}`. If the bridge deliberately can't know a result's shape, the three sentences should narrow to what the code does; otherwise `after` needs a binding. Your call which.

**Coverage, against the machinery already running here.** `intent_bridge.py` carries twenty-eight `raise` statements against six test cases. For the receipt rules, `test_vector_completeness.py` enforces two load-bearing fixtures per rule and defect-independence between them ("#124's criterion, executed"), and `test_margins_have_not_thinned` puts the floor plainly: two. The refusals I'd pin first: `authorizer_key_id` ≠ trusted `kid` (refuses correctly, no test, and it is the headline claim of the trust-boundary doc); `tool_call_digest` mismatch (the scope test changes the tool *name*, a different rule — a same-name call with different arguments is what reaches the digest comparison, and nothing does today); not-yet-valid (a different branch from expired, only expired is covered); the two PIC digests as two defects; PIC profile mismatch; `declaration_digest`; impact outside scope; `transcript.before` mismatch; non-boolean `transcript_required`.

**The new schema is read by nothing.** `grep -rn pic-trace-bridge tests/ src/` returns one hit — the profile constant. No test loads `schema/pic-trace-bridge-v1.json`, no implementation references it, and the only validation it received is the meta-validation in your evidence block, which the suite doesn't run. So nothing would notice it diverging from the verifier — it pins `signature` to 86 base64url characters, for instance, while the code accepts anything that decodes. The repo already has both halves of the fix: per-schema positive/negative vectors (`test_delegation_passes_json_schema` / `test_delegation_bad_digest_fails_json_schema`) and the comparison guard `test_packaged_schema_matches_the_normative_schema`, which landed in #136 after two schema files drifted in three places while each suite exercised only one of them.

**Small consistency note.** `now` has no bool guard while `authorized_at` and `expires_at` in the same function do; `now=True` currently fails closed as instant 1, by accident rather than by rule.

**Worth cross-linking to #66.** This bridge is the first executable instance of the approval-shaped binding that thread is converging on: a validity window, per-action scope, digest binding to the specific call, fail-closed. The two gaps that thread has already named — an authorization withdrawn before execution, and authority-epoch staleness — are exactly what a v2 of this profile would add, and the candidate fixtures being drafted there could target this artifact rather than a hypothetical one.

One tiny thing while reading the trailer: the sign-off is `imran@agentrust.com` — MAINTAINERS.md has `opaque.co` and the org domain is `agentrust-io.com`. If the third domain is yours, ignore this; in a provenance project the DCO identity seemed worth a glance.

Approving, with the auto-merge in mind, and with the ledger explicit: the PR says "Implements", not "Closes", so #179 stays open — deliverable 4 (deterministic vectors, positive and negative) still lives there, and I'll take it as follow-up unless you'd rather keep it. The `transcript.after` question resolves either way after merge — a binding or three narrower sentences, your pick — and the schema round-trip rides along with the vectors. The design does what the threat-boundary doc claims where I could test it; nothing I found weakens the core.

PR #184, opened 2026-08-18
rfc(draft, non-normative): A2A delegation-link verification profile, with 23 conformance vectors

lywinged:
Opened as a draft at your request on the seam thread. **Non-normative.** Requirement keywords are lowercase throughout the document, deliberately: `CONTRIBUTING.md` draws the line that normative text lives in the specifications, and this file binds nothing until its rules are adopted, at which point they become uppercase in `spec/` and this becomes a pointer to where they went.

No schema change. The `delegation` block is already normative in v0.2; this proposes verification rules for it.

## What is here

`docs/rfcs/a2a-delegation-profile.md`, and `examples/delegation-link/` with 23 vectors and their generator. The signing keys derive by role label from one published seed, so there is no secret and the whole set is reissuable by anyone; each vector carries the anchor JWKs in `context.trusted_root_keys`. `tests/test_generators_reproduce_fixtures.py` already covers this generator, since it discovers them rather than listing them. `tests/test_delegation_vectors.py` runs them; `tests/test_delegation_completeness.py` guards the set the way `test_vector_completeness.py` guards the receipt corpus, so a rule with one load-bearing vector fails rather than passes quietly.

585 pass on 3.11, `ruff` clean, no `src` or schema changes.

## The corpus was built before the rules, and that is the part worth keeping

The three decisions in §4 were not read out of the existing text. They were hit, because no vector could be written without settling them, and in each case the text supports both branches. A reader passes over all three without noticing; someone building a fixture cannot get to the end of one.

That gives a measurement the repository does not otherwise have. Not *do the tests cover the rules*, which measures an implementation, but **do two independent readings of the same normative text produce the same rules**, which measures the specification. Where two readings agree the text is doing its job. Where they diverge, the divergence names the sentence that is missing.

## What running it against cA2A returned

Measured against `ca2a_verify.verify_trace_dag` at ca2a `5dd77b2`. The run itself is from an earlier session, so I checked what has landed since rather than restate it: ca2a is now at `8c49177`, four commits on, and they add credential validity windows. Neither `ca2a_verify/dag.py`, which holds `verify_trace_dag`, nor `ca2a_runtime/trace_binding.py`, which holds `trace_record_hash`, is among the files they touch, so what follows still describes the current head. The agent-manifest vector count further down has not been rechecked the same way; treat it as measured then. Say the word on anything that matters to a boundary call and I will re-run the corpus and post the output rather than the claim. Thirteen of the 23 vectors exercise the record-linkage surface that function covers; the other ten are credential defects its own docstring assigns elsewhere.

**§4.1 is confirmed independently.** `ca2a_runtime.trace_binding.trace_record_hash` computes `"sha256:" + sha256(rfc8785.dumps(signed_record))`, the complete record with signature included, byte-identical to what this profile specifies. That was arrived at separately, in another repository.

**Vectors 01 to 07 agree.** Valid chains accepted, the absent parent and the body-digest link rejected as broken links, both signature vectors rejected as bad signatures. Same verdict, same reason, two implementations.

**Vectors 22 and 23 are the SHA-384 case you ruled on.** They already assert `unverifiable` with `digest_algorithm_unsupported`, so your ruling and this corpus agree. cA2A currently accepts a `sha384:` link, compares it against a digest it only ever computes as `sha256:`, and reports `ProvenanceLinkBroken` with the detail *"a tampered or reparented record was detected"*. Under the ruling that is no longer a divergence to document but a defect, and I will file it in that repository rather than describe it here.

**The trust contract differs, and this one is a boundary rather than a bug.** `verify_trace_dag` requires every record's `cnf.jwk` to be in the trusted set; this profile anchors on the root's key and lets the chain carry the rest. Run under the root-anchored contract, cA2A rejects every chain longer than one record, valid ones included. Neither model is wrong: theirs fits an orchestrator that knows every participant, this one fits the cross-organisation case where that knowledge is exactly what is missing. cA2A already uses the root-anchored model on its other surface, where `verify_chain` takes `trusted_root_issuers`. Yours to draw.

**The credential surfaces cannot be compared at all, and that is the finding**, not a gap in the run.

## What this closes on the agent-manifest side

Of its 21 vectors, two touch delegation and both are single-hop, so its narrowing and depth logic has no vector coverage. The same corpus covers it.

## Not in scope here

Mutual delegation, which you said you would cover separately. Everything in this corpus is one-directional and stays that way.


imran-siddique:
Noting for whoever reads this next, including me: the SHA-384 case is closed upstream. agentrust-io/ca2a#119 landed yesterday and carried the fix, not just the finding. At ca2a `52141e8`, `verify_trace_dag` now raises `TraceDigestUnsupported` with the detail "the chain is unverifiable here, not invalid" before the link comparison is reached, so a schema-valid `sha384:` parent link is no longer reported as a tampered or reparented record.

That means vectors 22 and 23 now describe fixed behaviour rather than a live divergence, and they agree with the implementation as well as with the ruling. Worth dropping that divergence from §4 when this branch next moves, so the document does not carry a defect report that is already resolved.

Also worth stating plainly, because the comparison is the part of this I value most: §4.1 was confirmed independently. `ca2a_runtime.trace_binding.trace_record_hash` computes the digest over the full signed record with the signature included, using RFC 8785 canonical bytes, byte-identical to what the profile specifies, and that was arrived at in a different repository without reference to this text. Two readings of the same normative text converging is a measurement of the specification rather than of an implementation, and it is not one this repository otherwise takes.

The branch conflicts with main now. No hurry while it is a draft, and the review is still mine to do.

The UNKNOWN on the contributor check is the bot, not you. It returns UNKNOWN on any transient failure with no retry, and it has landed on you repeatedly. It carries no information and I do not read it as one.

lywinged:
Thank you Imran.
Both are done on the branch.

§7.1 carried the divergence in the present tense in two places, and both are re-tensed rather than deleted. I checked the fix rather than taking it from the report: at ca2a `52141e8`, `src/ca2a_verify/dag.py:194` raises `TraceDigestUnsupported` with that detail at line 199, and the parent-link comparison that produced `ProvenanceLinkBroken` is at line 201, after it. So the guard does precede the comparison, and 22 and 23 describe fixed behaviour.

They stay in the document because of what the case establishes rather than as a defect report. A corpus written to argue a rule found it, the other implementation changed, and the shape it changed to is the distinction §4.3 asks for. The second passage said two of three implementations separate unreadable from contradicted; it is now all three. §4.3's rule text is unchanged, since it states the rule rather than reporting on an implementation.

The branch is merged with main and no longer conflicts. The only conflict was `CHANGELOG.md`, where both sides had added an entry at the top of Unreleased/Added. The revocation entry from #187 is kept first and this proposal under it, since one describes what landed and the other what is proposed.

One thing that is not a response to your comment. #186 landed after this branch was opened, so on the merged tree `delegation-link` was the one vector set nothing graded, and the only failure. It is registered in `SETS` rather than named in `MEASURED_ELSEWHERE`, because it holds up when it is actually graded rather than only pointed at:

```
delegation-link: 23 vectors, 3 accepting, 10 distinct failure codes
```

No shortfall on either criterion decidable from the fixtures. Not satisfiable by an implementation that answers "accept" to everything, nor by one that answers "reject" to everything, and each of the ten failure codes is carried by exactly two vectors. Boundaries are counted by failure code, which `tests/adequacy.py` says is the set's assumption to justify: `tests/delegation_margins.json` records the per-code margin, and `tests/test_delegation_completeness.py` holds each rule to being load-bearing for both of its vectors, deleting the rule by rebuilding the registry without its entry rather than by matching source text.

605 passed, 1 skipped.

imran-siddique:
Let's ship it!

imran-siddique (approved):
Shipping, as said on the thread. Thank you for checking the ca2a fix at `52141e8` yourself rather than taking my report, and for re-tensing §7.1 in both places instead of deleting the divergence. A profile that records what was true and when is worth more than one that quietly reads as though the gap never existed.

Non-blocking, and not yours to fix: `CHANGELOG.md` shows as +265-264 because the file converted CRLF to LF wholesale. Measured on the diff bytes, all 259 removed lines end in CR, none of the added ones do, and there are exactly two genuinely new lines and zero genuinely deleted. So the content change is one entry and the rest is your editor doing what the repository never told it not to. This repo has no `.gitattributes`, which is the actual defect. I am adding one rather than asking you to hand-revert 259 lines of blame.

Issue #197, opened 2026-08-21
spec: one mechanism for binding an external fact to an attested record (§3.1, §3.1.1)


imran-siddique:
**Spec section affected**

§3.1 (Trust Record logical schema) and §3.1.1 (`origin`). New optional block, no change to existing required fields.

**Problem**

There is no way for a self-produced, hardware-attested Trust Record to reference a fact that lives outside it.

§3.1.1 gets close. `origin.source_event_id` is defined as "Identifier of the source event in that system, so a record traces back to it", which is exactly the shape needed. But `origin` is scoped to records **assembled from evidence someone else produced**, and the normative text is:

> **A record whose `origin.kind` is not `self` MUST carry `runtime.platform: "software-only"`, and a verifier MUST reject it otherwise.**

That rule is correct for what it was written for. An importer holding someone else's log has no quote to present, so a hardware platform value on such a record would be untrue. The consequence is the problem: today, the moment a record points at anything external, it stops being a hardware attested record.

This is not a one-off request. Three open items are the same shape:

- #179, bind authorized pre-execution intent to execution evidence
- #191, bind attributable human-approval outcomes to step-up and defer execution evidence
- A behavioural execution identifier, raised by Pavan Sudheendra (Cisco), co-chair of the AAIF Observability & Traceability WG, who is proposing that an AAIF Behaviour Trace reference a TRACE Trust Record as attached evidence while TRACE remains responsible for attesting the runtime, model and policy environment

Three instances means this wants one mechanism, not a field added per case.

**Proposed change**

Separate two things the spec currently conflates: **where the evidence came from**, which can lower assurance, and **what else this record points at**, which cannot.

Add an optional `references` block to §3.1. Assurance-neutral by construction.

| Field | Required | Meaning |
|---|---|---|
| `rel` | yes | Relationship type. Closed set, extensible by registration. Initial values: `authorized-intent` (#179), `approval-outcome` (#191), `behavior-trace`. |
| `id` | yes | Identifier of the referenced fact in its own system. |
| `resolver` | yes | Identifier of the party obliged to resolve `id`. |
| `retention` | no | How long `resolver` undertakes to keep `id` resolvable. |
| `digest` | no | Digest of the referenced object, when the producer has it at issue time. |

Normative rules:

1. `references` MUST NOT affect `runtime.platform`. A record carrying `references` and no `origin` block is still `self` and still whatever its platform actually was.
[...]

Issue #191, opened 2026-08-20
spec: binding attributable human-approval outcomes to step_up and defer execution evidence


Thru-Echoes:
**Spec section affected**
§3.1 (Trust Record claims and evidence composition), §3.3 (verification), and the conformance suite (agentrust-io/trace-tests).

**Problem**
#114 closed with the decision recorded: 1.0 stays unchanged, and any widened `step_up` and `defer` verb set lands under a future profile version, with the version boundary being settled in #116. The closing note names the operator-reported STEP_UP and DEFER semantics the substance of that future version. This issue proposes requirements and a design direction for that substance, so the discussion has a concrete starting point when the version boundary lands.

What the enum alone loses is documented in the #114 thread: who satisfied an escalation, and what its terminal outcome was. Approver identity has to be frozen at emission or the receipt's answer decays by audit time, and a `defer` without a required terminal state becomes a laundering verb, both per the production report there.

The record currently has no place for either fact. `delegation` joins machine-delegated authority into the Trust Record (`parent_record_hash`, `credential_id`). Human-granted authority has no equivalent join. agent-manifest ADR-0006 defines an attributable approval record (approver-signed, action-bound via `evidence_hash`, time-bounded, with #40 requiring a human-attributable `approver_id`), but nothing connects that approval to the run and call that consumed it. The PIC/TRACE authorization bridge binds a pre-execution authorized decision to execution evidence and deliberately keeps authorization and runtime attestation separate; it is allow-only, records no escalation outcomes, and requires no human authorizer, so the escalation lifecycle remains uncovered.

**Proposed change**
Requirements first, then a candidate direction. Requirements a solution should meet:

1. A verifier can determine, offline, from an evidence bundle: whether an escalated call's authorization was approved, denied, expired, cancelled, or explicitly closed unresolved, and whether approval evidence was consumed by the recorded run.
2. Authorization outcomes and execution evidence stay separate lifecycles, joined by identifiers, preserving the boundary the PIC/TRACE bridge establishes.
3. Approval evidence is attributable to a human principal per ADR-0006 and #40, and is signed by a key the record's subject does not control.
4. Unresolved outcomes are representable and never silently coerced into approval, denial, or expiry.
5. Unknown or unsupported profile versions are refused or reported per the obligations proposed in #116, which controls version and fallback behavior; this issue does not prejudge #116's resolution.

Candidate direction, offered for discussion rather than as settled design:

```text
Authorization outcome
[...]

PR #198, opened 2026-08-21
spec: add an assurance-neutral references block (#197)

imran-siddique:
Implements the proposal in #197. Normative change, so please read that issue first for the argument.

## What this fixes

A self-produced, hardware-attested Trust Record has no way to reference a fact outside itself.

`origin.source_event_id` is the right shape, defined as "Identifier of the source event in that system, so a record traces back to it". But §3.1.1 scopes `origin` to records assembled from someone else's evidence, and requires that a record whose `origin.kind` is not `self` **MUST** carry `runtime.platform: "software-only"`, with verifiers rejecting it otherwise.

So today, the moment a record points outward, it stops being a hardware attested record.

## Why one mechanism rather than a third field

Three open items are the same shape:

| Item | Wants to bind |
|---|---|
| #179 | authorized pre-execution intent |
| #191 | attributable human-approval outcomes |
| AAIF Observability & Traceability WG | a behavioural execution identifier |

Two bespoke fields is a pattern. Three is a design failure. Each becomes a registered `rel` value here instead.

## The rule that makes it safe

**The signature attests that this record points somewhere. It does not attest the truth of what it points at.** The pointer is produced inside the boundary that produced the record; the target is not.

That gives three normative consequences, all in the diff:

- `references` MUST NOT affect `runtime.platform`
- a verifier MUST NOT reject a record because a reference cannot be resolved
- a verifier MUST NOT treat a resolved reference as attested evidence

Without the second rule, whoever controls the target could invalidate evidence they do not hold. Without the third, this becomes the assurance laundering §3.1.1 exists to prevent.

## §3.1.1 is untouched

This does not weaken that MUST. It stops it being applied to a job it was not written for. Records assembled from third-party evidence still take `software-only`, exactly as now.

## Backward compatibility

Non-breaking. New optional field, existing records do not carry it, existing verifiers ignore unknown fields. No wire format or algorithm change. Per CONTRIBUTING a schema PR must track a merged spec change, so the schema and conformance vectors follow once the shape here is agreed rather than in this PR.

## What I have deliberately left unresolved

`resolver` and `retention` are the interesting fields and I do not expect one round of review to get them right.

A reference is worth only the ability to resolve it later. That is not hypothetical: in the npm keyv compromise the poisoned packages carried valid provenance logged in a transparency service, and fifteen days on the attestations could not be fetched, because the only working index was the artifact digest and unpublishing removed the metadata holding it. Log intact, records unreachable.

TRACE has the same exposure through `transparency`. §7 open question 3 is the same problem for that field and I think the two want resolving together. The spec text says `retention` is an undertaking that nothing here enforces, rather than implying otherwise.

## Sponsor

Not yet named. CONTRIBUTING says proposals are evaluated on the technical argument alone and that a normative PR without a sponsor is not rejected on that basis, so raising it this way deliberately. Happy to have a maintainer carry it, or to name a sponsor if reviewers would rather.

### DIFF
--- a/spec/trace-v0.2.md
+++ b/spec/trace-v0.2.md
@@ -124,6 +124,7 @@ The Trust Record is the unit of evidence. All fields are required unless marked 
 | `data_class` | Classification of inputs and outputs | Classification label bound to per-call execution |
 | `tool_transcript` | MCP / A2A tool calls invoked, parameters classified, responses filtered | MCP / A2A protocol transcripts bound to TEE measurement |
 | `origin` | OPTIONAL. Where the evidence came from, when that is not this runtime. See §3.1.1. | — |
+| `references` | OPTIONAL. Facts outside this record that it points at. Assurance-neutral: see §3.1.2. <!-- CHANGED: #197 - new optional assurance-neutral reference block --> | — |
 | `build_provenance` | How the running code and model were built | SLSA Provenance v1.0 |
 | `appraisal` | Verifier's appraisal of evidence | EAR (EAT Attestation Results) |
 | `transparency` | Inclusion proof on append-only log | SCITT Receipt URI |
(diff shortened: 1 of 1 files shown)


Qiang-Xu (approved):

lywinged (commented):
Two things that are checkable, and one design question that I think decides the shape of the block.

`schema/trace-claim.json` is `additionalProperties: false` at the top level and this branch does not touch it, so a record carrying `references` does not validate. Measured on `examples/tpm2.json`, which validates as committed and fails with `Additional properties are not allowed ('references' was unexpected)` once an entry from the field table is added. Rule 2 requires the signature to cover the block, so as it stands there is no conformant record that can carry it. The suite is green at 496 passed because nothing connects the section 3.1 field table to the schema.

The shape is unstated as well. The section says "a `references` entry" and "an entry in `references`" throughout, which reads as a collection, but neither it nor the field table says whether that is an array or an object. Section 3.1.1 does not state `origin`'s shape either, because the schema does it there. With no schema entry there is nowhere that says what a producer emits, and canonicalisation under section 3.2.2 differs between the two. The same gap leaves duplicates undefined: whether two `authorized-intent` entries are legal, and what a verifier does with them.

On the design. Section 3.3.2 answers the same question for external execution evidence and answers it the other way. There, a verifier configured with the issuer key MUST reject the Trust Record when the signature or the digest check fails, and a verifier without the key treats the receipt as unverified rather than invalid and SHOULD surface an advisory status rather than silently ignoring it. Rule 3 here says MUST NOT reject and MUST NOT count, with nothing surfaced. Two sections apart, the same question has two answers and no text reconciling them.

That difference lands hardest on `digest`. An entry carrying one is checkable in the way an external receipt is checkable, and rule 3 forbids the check from mattering. So the question I would want settled: what is a verifier permitted to do after resolving an entry and matching its digest? If the answer is nothing, `digest` is decorative and could come out. If it is anything at all, rule 3 needs that exception written down, and `appraisal` is `additionalProperties: false` over `status`, `verifier`, `policy_ref`, `timestamp` and `provenance_depth_verified`, so there is nowhere for a verifier to record that it happened.

Smaller, and worth settling now rather than at the fourth value: `rel` is a closed set, on the same reasoning section 3.1.1 gives for `origin.kind`, but there is no rule for a value a verifier does not recognise. Without one, adding a fourth `rel` is a breaking change for a strict verifier, which the backward-compatibility note does not cover.

PR #203, opened 2026-08-23
security: enforce PIC bridge schema constraints

imran-siddique:
Security hardening prepared with reproducible evidence and tests. Please review and run required CI.

### DIFF
--- a/src/agentrust_trace/intent_bridge.py
+++ b/src/agentrust_trace/intent_bridge.py
@@ -65,6 +65,24 @@ def _digest(value: Any, field: str) -> str:
     return value
 
 
+def _nonempty_string(value: Any, field: str) -> str:
+    if not isinstance(value, str) or not value:
+        raise IntentBridgeError(f"{field} must be a non-empty string")
+    return value
+
+
+def _unique_nonempty_strings(value: Any, field: str) -> list[str]:
+    if (
+        not isinstance(value, list)
+        or not value
+        or not all(isinstance(item, str) and item for item in value)
+    ):
+        raise IntentBridgeError(f"{field} must be a non-empty array of non-empty strings")
+    if len(value) != len(set(value)):
+        raise IntentBridgeError(f"{field} must not contain duplicates")
+    return value
+
+
 def verify_bridge(
     bridge: dict[str, Any], trusted_authorizer_jwk: dict[str, Any], *,
     declaration: dict[str, Any], pic_intent_digest: str, pic_args_digest: str,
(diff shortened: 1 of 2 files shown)


imran-siddique:
@zohebk8s @lywinged Could you please review this fully green security PR? It enforces PIC bridge schema/runtime parity and is currently blocked only on maintainer approval. Thanks.

lywinged:
> Could you please review this fully green security PR? It enforces PIC bridge schema/runtime parity and is currently blocked only on maintainer approval. Thanks.

Reviewed by running it, after the merge, on main at 4785fa7.

Parity is complete and I could not break it: 46 cases — each of the eleven required fields dropped, a wrong type substituted into each, the two profile constants and the root object violated, and hand-written boundaries for minLength, minItems, uniqueItems, minimum and the decision enum — produced zero cases where the schema says no and the code says yes.

One of the 46 is classified in a way that reads against the docstrings. decision is enum: ["allow", "deny"] in the schema and a != "allow" comparison in the code, so "banana", 1, None and {"a": 1} all raise AuthorizationDenied: the signed decision is not allow. It fails closed, so this is not a hole — but IntentBridgeError is documented as “malformed… or cannot be evaluated” and AuthorizationDenied as “the signed decision is not an authorization to execute”, and a value outside the enum is the former. A membership check before the comparison moves them to the class that describes them.

The part I would act on is the tests. Removing each of the five guards this PR adds, one at a time, and running the whole suite:

guard removed	suite
the three _nonempty_string calls	729 passed
or not value — the minItems half	729 passed
item non-emptiness — the minLength half	729 passed
authorization[field] < 0	1 failed
the duplicate check	2 failed

Three of the five can be deleted and CI stays green. Two causes. Both pytest.raises are bare IntentBridgeError, and AuthorizationDenied and AuthorizationMismatch subclass it, so a case passes on any rejection — with the item-emptiness guard gone, ("tools", [""]) still raises, as AuthorizationMismatch: executed tool is outside the authorized tool scope. And the scalar cases run at now=0 while the fixture’s authorized_at is 100, so authorization is not yet valid catches four of the five before the guard they name is reached. minItems additionally has no case: the four scope parameters cover the duplicate and empty-string halves, not the empty array.

Three changes fix all of it, and I ran them: match= on both pytest.raises, the scalar cases at now=150, and ("tools", []) and ("impacts", []) added. That is 731 passed, ruff and mypy clean, and every one of the five guards then goes red when removed — including the three that are currently free to delete.

One piece of context for why this file carries more weight than most: agentrust_trace.intent_bridge is in __all__ but has no caller inside this repository and no conformance coverage in agentrust-trace-tests, so tests/test_intent_bridge.py is the only thing checking it. Happy to open the test-only PR if useful.

PR #210, opened 2026-08-24
docs: align schema.md with the v0.2 schema

Thru-Echoes:
## Summary

docs/schema.md introduced itself as the v0.1 schema reference in two places, listed a platform enum missing three values the schema defines (aws-nitro, arm-cca, google-confidential-space) while spelling the TPM value tpm-2.0 where the schema says tpm2, and gave contradictory guidance on transparency: the top-level table says omit or null when a record is unanchored, while the section below recommended an empty string.

## Why

The page documents schema/trace-claim.json, so drift between them sends integrators to the schema with wrong values. The empty-string guidance is worse than drift: the reference model rejects empty strings (min_length=1 on transparency), so a producer following the section as written emits records the model refuses.

## Verification

Full test suite passes (729 passed, 1 skipped), including the docs-consistency tests. Platform values and the transparency behavior were checked directly against schema/trace-claim.json and src/agentrust_trace/models.py.

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -1,6 +1,6 @@
 # Schema Reference
 
-JSON Schema for the TRACE v0.1 Trust Record. Source: [`schema/trace-claim.json`](https://github.com/agentrust-io/trace-spec/blob/main/schema/trace-claim.json).
+JSON Schema for the TRACE v0.2 Trust Record. Source: [`schema/trace-claim.json`](https://github.com/agentrust-io/trace-spec/blob/main/schema/trace-claim.json).
 
 ## Top-level fields
 
@@ -41,7 +41,7 @@ Binds the execution environment. Platform-specific fields vary by TEE type.
 
 | Field | Type | Required | Description |
 |---|---|---|---|
-| `platform` | string | **yes** | One of: `amd-sev-snp`, `azure-cvm-sev-snp`, `intel-tdx`, `nvidia-h100`, `nvidia-blackwell`, `tpm-2.0`, `software-only` |
+| `platform` | string | **yes** | One of: `intel-tdx`, `amd-sev-snp`, `azure-cvm-sev-snp`, `nvidia-h100`, `nvidia-blackwell`, `aws-nitro`, `arm-cca`, `google-confidential-space`, `tpm2`, `software-only` |
 | `measurement` | string | **yes** | Hardware measurement hash (`sha384:` for SEV-SNP/TDX, `sha256:` for TPM) |
 | `rim_uri` | string | no | Reference Integrity Manifest URI for hardware verification |
 | `firmware_version` | string | no | TEE firmware version |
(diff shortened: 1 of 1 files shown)


lywinged (approved):
Both of these check out when run rather than read, and the second one is the kind of drift
that is worse than it looks.

The platform row: the schema's enum is `intel-tdx`, `amd-sev-snp`, `azure-cvm-sev-snp`,
`nvidia-h100`, `nvidia-blackwell`, `aws-nitro`, `arm-cca`, `google-confidential-space`,
`tpm2`, `software-only`. The table was missing four of those and offered `tpm-2.0`, which
the schema rejects. Your list is set-equal to the enum. So the table was not only
incomplete, it was documenting a value that cannot validate, and a producer following it
would have been refused by the artifact that decides.

The transparency line: measured, not inferred. `TrustRecord.model_validate` with
`transparency=""` returns *"String should have at least 1 character"*, because the field is
`Annotated[str, MinLen(1)] | None`. The old sentence told producers to emit the one value the
reference model refuses. Your replacement is right on both halves, omit or `null`.

Two notes on the overlap and one offer.

The two version lines are also in #207, byte-identical, and `git merge-tree` shows no
conflict, so the order does not matter and neither of us needs to rebase for it. Nothing to
resolve.

The part I would not have found by hand is that the same class of drift is sitting one row
above yours. `policy.enforcement_mode` accepts `enforce`, `advisory`, `silent`, `declared`,
and the table says "`enforce` or `silent` (advisory)", which presents `advisory` as a gloss
on `silent` rather than as its own value and omits `declared` entirely. `declared` is the
value a producer with no policy engine is supposed to use, and the schema says a consumer
MUST NOT read it as evidence that any rule was checked, so a reader who never learns it
exists picks one of the three that claim an evaluation happened.

So I wrote the guard your PR argues for: it walks every enum in `schema/trace-claim.json`,
reads the row that documents each one, and requires the two to be set-equal in both
directions. It catches your platform finding and the `enforcement_mode` one automatically.
It is on `docs/enum-guard` and **it is red until this merges**, deliberately: with the
`enforcement_mode` fix alone the file is 1 failed, 6 passed, and with your platform row
applied on top it is 7 passed with the whole suite at 736 passed, 1 skipped. I will open it
after yours lands rather than carry your change inside mine.

imran-siddique (approved):
Checked both of your substantive claims against the artifacts at pinned main `a1bd72a2` rather than against the page.

The platform enum in `schema/trace-claim.json` is exactly the ten values you list, in that order. The page named seven, spelled `tpm-2.0` where the schema says `tpm2`, and omitted `aws-nitro`, `arm-cca` and `google-confidential-space`. An integrator reading it would have written a value the schema rejects and a value the schema does not have.

The transparency one is the more valuable half and I want to say why. `models.py:277` is `transparency: Annotated[str, Field(min_length=1)] | None = None`. The section as written told producers to emit `""` for an unanchored record, and the reference model refuses it. That is not drift between a doc and a schema, it is a documented behaviour that produces invalid records, and the top-level table three sections up already said omit-or-null. Good catch, and the reason you gave for it is the right one.

Housekeeping so you are not surprised by the merge: two of your four hunks, the v0.1 to v0.2 labels on lines 3 and 161, landed twenty minutes ago in #207. The content is identical so git folds them in with no conflict, and `merge-tree` confirms zero. The merge leaves exactly your two unique changes. Nothing for you to do.

First contribution here, and it was verified against the code rather than argued from the page. That is the standard. Approving and merging.

PR #215, opened 2026-08-25
docs: add verification outcome statement guidance

safal207:
Informative-only documentation contribution, related to #66.

## What this adds

`docs/verification-outcome-statements.md` - a relying-party / auditor interpretation guide indexed by verifier outcome (a third view alongside `docs/trust-levels.md`, which is indexed by trust level, and the OWASP crosswalk, indexed by risk). For each representative outcome it maps:

- the bounded statement the evidence supports,
- the required non-claim that prevents assurance overstatement,
- the evidence worth retaining,
- a sensible follow-up.

Representative rows: verified at the required floor; honestly downgraded with floor still met; resolved-and-contradicted; unresolvable below the floor; `policy.enforcement_mode: declared`.

## Scope guarantees

- **No schema changes.** No field added, removed, or re-typed.
- **No normative spec changes.** The file carries no uppercase RFC 2119 keyword and does not touch `spec/trace-v0.2.md`.
- Every bounded statement and non-claim is anchored to existing text in `spec/trace-v0.2.md`, `docs/verification.md`, or `LIMITATIONS.md`; statements that cannot be anchored are listed as open questions (#191/#116 approval rows, transitive coverage URI, `appraisal.policy_ref` reproducibility).
- `enforcement_mode: declared` is documented strictly per its §4.3 definition: it does not evidence that any rule was evaluated or enforced.
- Verification results are not turned into a general trust score, and a valid record/receipt is not treated as proof of physical or business outcome success.

## Checks run

- `pip install -e .[dev]` - OK
- `pytest` - 736 passed, 1 skipped, 3 pre-existing failures in `test_generators_reproduce_fixtures.py` (generator byte-reproduction; confirmed identical on clean upstream `main` at 7cc2fab, unrelated to this docs change)
- `ruff check src tests` - all checks passed
- `mypy src/agentrust_trace` - Success: no issues found in 10 source files
- `git diff --check` - clean

Related to #66

### DIFF
--- a/docs/verification-outcome-statements.md
+++ b/docs/verification-outcome-statements.md
@@ -1,0 +1,61 @@
+---
+description: How a relying party or auditor turns a TRACE verifier outcome into a bounded statement - what each outcome supports, what it never claims, and what evidence to retain.
+---
+
+# Verification outcome statements
+
+> **Non-normative.** This page is informative. It changes no schema field, wire format, required claim or conformance requirement, and carries no uppercase RFC 2119 keyword. [Section 3.3 of the specification](../spec/trace-v0.2.md) and the [verification protocol](verification.md) carry the normative rules; where those documents already speak, this page repeats their wording rather than adding to it.
+
+## Purpose
+
+A TRACE verifier produces an outcome: an `appraisal.status`, a recorded provenance depth, a set of checks it ran and did not run. The people who consume that outcome are often not verifiers. Auditors, assessors, customers and regulators read verifier output and have to turn it into a written conclusion. The failure mode this page exists to prevent is assurance overclaim - a conclusion written slightly stronger than the evidence behind it, so that a valid signature quietly becomes "the agent complied", or a signed acceptance receipt becomes "the action succeeded".
+
+Two pages already tell a reader what not to conclude, each cut a different way: [Trust Levels](trust-levels.md) carries a *What it does not prove* line under each level, and the [OWASP Agentic Top 10 crosswalk](crosswalks/owasp-agentic-top-10.md) carries a *Limitations / out of scope* column and a closing *What TRACE does not do* list. This page is the third view of the same discipline, indexed by verifier outcome rather than by trust level or risk category. Where wording overlaps, those two pages carry the meaning and this page follows them.
+
+## Non-normative status
+
+Nothing on this page changes what a conformant implementation does. A bounded statement below is permitted only because some sentence in [`spec/trace-v0.2.md`](../spec/trace-v0.2.md), [`docs/verification.md`](verification.md) or [`LIMITATIONS.md`](../LIMITATIONS.md) already makes it true; each matrix row names that anchor. A statement that cannot be anchored to existing text belongs under [Open questions](#open-questions), not in the matrix.
+
+Machine-checkable verifier-output/statement pairs are out of scope for this page; they belong wherever the vector sets whose outcomes they consume live.
+
+## Interpretation matrix
+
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):
Checked the scope guarantees rather than taking them, because "informative only" is the claim that decides whether this needs a sponsor under our spec-contribution policy.

Grepped every added line for RFC 2119 keywords: zero uppercase MUST, SHALL, SHOULD, MAY, REQUIRED, RECOMMENDED or OPTIONAL. `spec/trace-v0.2.md` untouched, no schema field added, removed or re-typed. The claim holds exactly, so the sponsor rule does not apply here.

**Then I spot-checked your hardest citation, and it is precise.** Your `declared` row says that reading it as evidence a rule was checked contradicts the field's own definition. §4.3 at pinned `bd4098cd` reads: "A consumer MUST NOT read it as evidence that any rule was checked... A verifier appraising for enforcement SHOULD treat `declared` as it treats an absent enforcement claim." You paraphrased it into lowercase prose without weakening or strengthening it. That is the difference between an anchored guide and a plausible one, and it is what makes the other rows credible without my checking each.

**The Required non-claim column is the reason this page should exist.** The failure mode you name, a valid signature quietly becoming "the agent complied", is the one that actually happens, and it happens in the report rather than in the verifier. A verifier that is right and a report that overstates it produce the same outcome as a verifier that is wrong. Putting the non-claim in the same row as the supported statement, rather than in a caveats section further down, is what makes it hard to drop when someone copies a row into a document.

Three things I want to credit specifically.

Positioning against what already exists instead of duplicating it. Naming Trust Levels as indexed by level and the OWASP crosswalk as indexed by risk, then saying that where wording overlaps those pages carry the meaning and this one follows, means this cannot silently drift into a competing source of truth.

Putting unanchorable statements under Open questions rather than into the matrix. That is the same discipline as the page's own subject applied to the page itself.

And you added the `mkdocs.yml` nav entry. We merged a PR an hour ago whose entire purpose was fixing a page that shipped without one. You did not repeat it, unprompted, on a first contribution.

"Unverifiable is not incriminating" and "unverified is not silently true" are the two sentences I expect to end up quoting.

Approving and merging.

lywinged:
Post-merge note, since this is already on `main`. The `Required non-claim` column beside the
supported statement is the part I expect to get cited. One link in the file does not resolve,
and the cause is worth naming because four more headings carry the same shape.

Line 29, in the "Resolved and contradicted" row:

`[Step 5 status meanings](verification.md#step-5--appraise-the-claims)`

The heading it points at is `### Step 5 — Appraise the claims`. Rendered with this repository's
own extension set, the ` — ` between `5` and `Appraise` collapses to a **single** hyphen, so the
generated id is `step-5-appraise-the-claims`. The link carries two, which is what an em-dash
would give if it slugified to `--`.

Checked on the published page rather than only locally. `trace.agentrust-io.com/docs/verification/`
returns 200 and contains `id=step-5-appraise-the-claims` once and `step-5--appraise-the-claims`
zero times, so the link currently lands at the top of that page instead of at the section it
names.

One character, whenever the file is next touched. Worth knowing that all five `Step N` headings
in `verification.md` carry the same em-dash and every one of their ids collapses to a single
hyphen; nothing in the repository links to the other four yet.

Every other internal link in the file resolves: 11 of 12, checked by rendering each target and
looking for the id.

safal207:
@lywinged Good catch — thank you for checking this against the rendered site rather than only the source.

You’re right: the generated anchor is `step-5-appraise-the-claims`, with a single hyphen. I’ll correct the internal link in a small follow-up change and verify the rendered target as part of that fix.

PR #216, opened 2026-08-25
test(sign): freeze the clock on the future-skew boundary

### LINKED ISSUE #183: test_sign future-skew assertion fails when the clock crosses a second boundary
**What is wrong**

`tests/test_sign.py::test_verify_record_future_skew_is_deployment_configurable` fails whenever the wall clock crosses a second boundary while it runs. The margin is not the one second it looks like, and the condition is exact rather than mysterious.

```python
record["iat"] = int(time.time()) + 601
with pytest.raises(ValueError, match="max_future_skew_seconds=600"):
    verify_record(record, key_to_jwk(key), max_future_skew_seconds=600)
```

The test sets `iat` from `int(time.time())`, which truncates, while the check reads the clock unrounded. Writing the true time at setup as `k + f` — `k` whole seconds, `f` the fraction — and `δ` for everything that elapses before `verify_record` reads the clock (key generation, signing, the call):

```
iat  = k + 601
age  = (k + f + δ) - (k + 601) = f + δ - 601
raises  ⟺  age < -600  ⟺  f + δ < 1
```

The tolerance is not one second out of six hundred. It is however much of the current second is left when the test starts, minus its own runtime. `f` is uniform over `[0, 1)`, so the per-run failure probability is `δ` — whatever the machine is doing at the time, and not small on a loaded runner.

Deterministic reproduction, patching `time.time` on the module object (`verify_record` imports `time` inside the function, so the module-level patch reaches it). Nothing about the record changes between rows; only where in the second the test begins:

```
  fraction=0.0     elapsed=0.001  f+elapsed=0.0010  -> raises (green)
  fraction=0.0     elapsed=0.06   f+elapsed=0.0600  -> raises (green)
  fraction=0.5     elapsed=0.001  f+elapsed=0.5010  -> raises (green)
  fraction=0.5     elapsed=0.06   f+elapsed=0.5600  -> raises (green)
  fraction=0.95    elapsed=0.001  f+elapsed=0.9510  -> raises (green)
  fraction=0.95    elapsed=0.06   f+elapsed=1.0100  -> NO RAISE   <-- "DID NOT RAISE ValueError"
  fraction=0.9995  elapsed=0.001  f+elapsed=1.0005  -> NO RAISE
  fraction=0.9995  elapsed=0.06   f+elapsed=1.0595  -> NO RAISE
```

Observed once in an ordinary full-suite run, which is what prompted tracking it down.

**Spec section or file**

`tests/test_sign.py::test_verify_record_future_skew_is_deployment_configurable`, against `src/agentrust_trace/sign.py:436`.

**Expected behavior**

The assertion should hold on every run rather than on most of them, and it should still be testing the boundary it names.

Two ways to get there, and they are not equivalent:

*Freeze the clock.* `monkeypatch.setattr(time, "time", lambda: FIXED)` for the duration. The local `import time` in `verify_record` binds the module object, so patching the module reaches it. The boundary can then be asserted at exactly ±1 second with no race, which is what the test appears to have been written to do.

*Widen the margin* — `+601` becomes something like `+660`, with the accepting bound moved to match. Removes the race, and stops testing the boundary: an implementation whose comparison is off by a few seconds passes either way.

The first keeps what the test is for. Happy to send it as a PR if that reading is right — a fixture change, no behaviour touched.

**Impact**

Neither of the two failure modes this section asks about. `sign.py` is correct: the record really is 601 seconds in the future at the instant it is written and really is 600 by the time it is read, and the comparison does the right thing with both. No conformant implementation produces an invalid record because of this, and no verifier accepts an invalid one.

The cost is to the suite's signal. An intermittent red on a security-relevant freshness bound is the kind of failure that gets re-run rather than read, and a suite with one test that fails for its own reasons is a suite where the next real intermittent failure is assumed to be that one.

Scope is this test alone. Its second assertion (`max_future_skew_seconds=602` must *not* raise) needs `f + δ ≥ -1`, which always holds. `test_disabling_max_age_does_not_disable_future_bound` uses `+3600` against the 300-second default — a margin of 3300 seconds — so it is unaffected.

---

imran-siddique:
Closes #183. Reported and diagnosed by @lywinged; the analysis below is theirs, the mutation results are mine.

## The defect

`tests/test_sign.py` set `iat` from `int(time.time())`, which truncates, while `verify_record` compares against an unrounded `time.time()`. Writing the true setup time as `k + f` and `d` for everything that elapses before the comparison:

```
iat  = k + 601
age  = (k + f + d) - (k + 601) = f + d - 601
raises  <=>  age < -600  <=>  f + d < 1
```

So the tolerance was never one second out of six hundred. It was whatever remained of the current second when the test started, minus its own runtime. `f` is uniform over `[0, 1)`, which makes the per-run failure probability `d` rather than anything fixed, and that is why this read as noise instead of a bug.

Confirmed at pinned `e0afe8eb`: `sign.py:436` is `age = time.time() - iat`, and `test_sign.py:380` was `int(time.time()) + 601`.

## Why freeze rather than widen

Moving the assertion to `+660` would remove the race by removing the test. An implementation whose comparison is off by a few seconds would pass either way, and this is a security-relevant freshness bound, not an incidental one.

`verify_record` does a local `import time` at `sign.py:327`, which binds the same module object, so a module-level `monkeypatch.setattr(time, "time", ...)` reaches it. With the clock frozen there is no race left, so the boundary gets pinned on both sides instead of one:

- 601s into the future is rejected at a 600s bound
- 600s into the future is accepted at the same bound
- the existing `max_future_skew_seconds=602` case is kept

## Both edges are load-bearing

Mutating `sign.py` one way at a time, with the test unchanged:

| mutation | result |
|---|---|
| `age < -max_future_skew_seconds - 2` (widen by 2s) | `Failed: DID NOT RAISE ValueError` |
| `age <= -max_future_skew_seconds` (strict to non-strict) | `ValueError: record is dated 600s in the future` |

The second one is the case the old test could not have caught: it never exercised the accepting side of the boundary, so an off-by-one that rejected a record exactly at the bound would have gone unnoticed.

Stability: the fixed test ran 20 consecutive times, 20 passed.

## Scope

That test alone, as agreed on the issue. The neighbours are fine and were checked: the `602` assertion needs `f + d >= -1`, which always holds, and `test_disabling_max_age_does_not_disable_future_bound` has a 3300-second margin.

## Scope note

Ran the full suite on this branch. The one unrelated failure,
`test_generators_reproduce_fixtures.py::test_generator_reproduces_its_committed_fixtures[delegation-link]`,
is a Windows line-ending artifact in my local environment and not a defect: the generator writes
platform line endings while the committed blobs are LF, and the regenerated fixtures are
byte-identical after normalising. It passes on CI.

### DIFF
--- a/tests/test_sign.py
+++ b/tests/test_sign.py
@@ -374,15 +374,46 @@ def test_verify_record_accepts_record_within_default_future_skew():
     verify_record(record, key_to_jwk(key))
 
 
-def test_verify_record_future_skew_is_deployment_configurable():
-    key = generate_key()
-    record = _minimal_record()
-    record["iat"] = int(time.time()) + 601
-    record = sign_record(record, key)
+def test_verify_record_future_skew_is_deployment_configurable(monkeypatch):
+    """The future-skew bound, asserted at exactly its edge under a frozen clock.
 
+    This test used to set ``iat`` from ``int(time.time())``, which truncates,
+    while ``verify_record`` compares against an unrounded ``time.time()``. Writing
+    the true setup time as ``k + f`` and ``d`` for everything that elapses before
+    the comparison, ``age = f + d - 601``, so it raised only while ``f + d < 1``.
+    The margin was not one second out of six hundred, it was whatever remained of
+    the current second minus the test's own runtime, and the per-run failure
+    probability was ``d`` rather than anything fixed (#183).
+
+    Freezing rather than widening is deliberate. Moving the assertion to ``+660``
+    would remove the race by removing the test: an implementation whose comparison
+    is off by a few seconds would pass either way, and this is a security-relevant
+    freshness bound. With the clock frozen there is no race left, so the boundary
+    is pinned on both sides instead.
+
+    ``verify_record`` does a local ``import time``, which binds the same module
+    object, so patching the attribute here reaches it.
+    """
+    frozen = 1_800_000_000.0
+    monkeypatch.setattr(time, "time", lambda: frozen)
+
+    key = generate_key()
+
+    # 601s into the future is outside a 600s bound.
+    beyond = _minimal_record()
+    beyond["iat"] = int(frozen) + 601
+    beyond = sign_record(beyond, key)
     with pytest.raises(ValueError, match="max_future_skew_seconds=600"):
-        verify_record(record, key_to_jwk(key), max_future_skew_seconds=600)
-    verify_record(record, key_to_jwk(key), max_future_skew_seconds=602)
+        verify_record(beyond, key_to_jwk(key), max_future_skew_seconds=600)
+
+    # ...and inside a bound configured above it.
+    verify_record(beyond, key_to_jwk(key), max_future_skew_seconds=602)
+
+    # 600s into the future is exactly at the bound, and accepted.
+    at_edge = _minimal_record()
+    at_edge["iat"] = int(frozen) + 600
+    at_edge = sign_record(at_edge, key)
+    verify_record(at_edge, key_to_jwk(key), max_future_skew_seconds=600)
 
 
 def test_disabling_max_age_does_not_disable_future_bound():


Qiang-Xu (approved):
I see, it looks good to me.

lywinged (approved):
Approving. I ran it rather than read it, and everything in the description held.

**One neighbour that is not on the list.** Scope says "the neighbours are fine and were
checked", and names two. Both check out: the `602` assertion needs `f + d >= -1` and so always
holds, and `test_disabling_max_age_does_not_disable_future_bound` comes out at a 3300-second
margin by the same arithmetic, which is your number. There is a third.
`test_verify_record_rejects_record_beyond_default_future_skew` is the test immediately above
the one you fixed, and line 361 sets `int(time.time()) + 301` against the default 300s bound.
Same construction, same algebra, same failure.

Forcing that test's own first clock read to `f = 0.9995` gives
`Failed: DID NOT RAISE ValueError`. Sweeping the forced phase, it passes at 0.9975 and fails
from 0.9980, which is where `1 - d` predicts. Its `d` runs around 1.6 ms, so it loses on the
order of one run in several hundred here, by the same mechanism #183 describes.

Across `tests/` only two sites sit one second from their bound: this one, and the `+299`
case beside it, which cannot race because it is the accept side. The other eight either have
no bound in play or sit four minutes or further from it. That is outside what we agreed on the
issue, so I am not asking you to widen this PR, but it is the same defect.

**Confirming what you already worked out**, so it is independent rather than repeated.
The pre-PR test reproduces on demand: same forced phase, same `DID NOT RAISE`, with `d`
around 2 ms. The frozen clock does reach the implementation through the local `import time`
at `sign.py:327`, which I checked sits inside `verify_record` rather than at module level.
Twenty consecutive runs of the new test, twenty passed. Both rows of your mutation table
reproduce here, error strings included, and a third of my own, replacing the comparison with
`if False`, also turns the test red. Your second row is the one that carries: against the
pre-PR tree, `<` changed to `<=` left the old test **green**, so `+600` is doing work rather
than symmetry.

Full suite here is 741 passed, 1 skipped, so the fixture failure you saw looks local as you
describe.

PR #217, opened 2026-08-25
fix(docs): AGT integration doc's ADR link and version claim were both stale

rajnisht7:
## What this changes

`docs/integration/agt.md` said AGT emits TRACE v0.2 Trust Records. That was wrong as the latest released AGT (`agent-governance-toolkit-core` 5.0.0 on PyPI) still emits trace-v0.1 (verified directly: `trace_sink.py:92` writes `eat_profile: "tag:agentrust.io,2026:trace-v0.1"`, and the package pins `agentrust-trace<0.3.0,>=0.2.0`). Only AGT's unreleased `main` branch has moved to v0.2.

The doc now says v0.1, scoped explicitly to the latest release, and notes that `main` has moved to v0.2 but hasn't shipped yet, and that the linked ADR (0032) still describes the v0.1 design and hasn't been updated for the change.

## Type of change

- [ ] Schema change
- [ ] Example addition

## Spec section

None

## Checklist

### DIFF
--- a/docs/integration/agt.md
+++ b/docs/integration/agt.md
@@ -2,7 +2,7 @@
 
 [AGT (Agent Governance Toolkit)](https://github.com/microsoft/agent-governance-toolkit) is the most widely adopted agent governance framework (4,100+ stars, 100+ contributors). It provides Cedar policy enforcement, SPIFFE/SVID identity, and Merkle-chained audit logs for any agent framework.
 
-AGT emits TRACE v0.1 Trust Records via `TRACEAuditSink` — see [ADR 0032](https://github.com/microsoft/agent-governance-toolkit/blob/main/docs/adr/0032-agt-emits-trace-v01-trust-records.md) for the full design.
+AGT emits TRACE v0.2 Trust Records via `TRACEAuditSink` — see [ADR 0032](https://github.com/microsoft/agent-governance-toolkit/blob/main/docs/adr/0032-agt-emits-trace-v01-trust-records.md) for the full design.
 
 ## What AGT emits
 
--- a/docs/integration/agt.md
+++ b/docs/integration/agt.md
@@ -2,7 +2,7 @@
 
 [AGT (Agent Governance Toolkit)](https://github.com/microsoft/agent-governance-toolkit) is the most widely adopted agent governance framework (4,100+ stars, 100+ contributors). It provides Cedar policy enforcement, SPIFFE/SVID identity, and Merkle-chained audit logs for any agent framework.
 
-AGT emits TRACE v0.2 Trust Records via `TRACEAuditSink` — see [ADR 0032](https://github.com/microsoft/agent-governance-toolkit/blob/main/docs/adr/0032-agt-emits-trace-v01-trust-records.md) for the full design.
+AGT emits TRACE v0.1 Trust Records via `TRACEAuditSink` in its latest release (`agent-governance-toolkit-core` 5.0.0) — see [ADR 0032](https://github.com/microsoft/agent-governance-toolkit/blob/main/docs/adr/0032-agt-emits-trace-v01-trust-records.md) for the full design. AGT's `main` branch has already moved to v0.2, but that change has not shipped in a release yet, and the linked ADR still describes the v0.1 design.
 
 ## What AGT emits
 
--- a/tests/test_version.py
+++ b/tests/test_version.py
@@ -118,6 +118,7 @@ _CURRENT_RECORD_SURFACES = (
     pathlib.Path("docs") / "integration" / "cmcp.md",
     pathlib.Path("src") / "agentrust_trace" / "adapters" / "agt.py",
     pathlib.Path("docs") / "tutorials" / "verifying-a-trust-record.md",
+    pathlib.Path("docs") / "integration" / "agt.md",
 )
 
 
--- a/tests/test_version.py
+++ b/tests/test_version.py
@@ -118,7 +118,6 @@ _CURRENT_RECORD_SURFACES = (
     pathlib.Path("docs") / "integration" / "cmcp.md",
     pathlib.Path("src") / "agentrust_trace" / "adapters" / "agt.py",
     pathlib.Path("docs") / "tutorials" / "verifying-a-trust-record.md",
-    pathlib.Path("docs") / "integration" / "agt.md",
 )
 
 


lywinged (commented):
You are right that AGT moved, and I could not see it from the sources your PR points at, so
here is what I found. The change is not wrong in direction. It is ahead of what AGT has
released, and its own citation now disagrees with it.

**Where AGT is**

- **`main` emits v0.2.** In
  `agent-governance-python/agent-mesh/src/agentmesh/governance/trace_model.py:18`,
  `TRACE_EAT_PROFILE = "tag:agentrust-io.com,2026:trace-v0.2"`, and `trace_sink.py:93` emits
  that constant. That package pins `agentrust-trace>=0.5.1,<0.6.0`. It is the same file as
  the released one at a later point: diffed against the copy inside core 5.0.0 they differ in
  four places, the profile constant among them.
- **Nothing released emits it yet.** The newest AGT on PyPI that ships
  `agentmesh/governance/` is `agent-governance-toolkit-core` 5.0.0, uploaded 2026-08-03, and
  its `trace_sink.py:92` still writes `"eat_profile": "tag:agentrust.io,2026:trace-v0.1"`
  while pinning `agentrust-trace>=0.2.0,<0.3.0`. `agentmesh-platform`, which carries the same
  module name, is at 3.7.0 from May and ships no sink at all. So no released AGT emits v0.2,
  and the only released sink writes v0.1.
- **The ADR this sentence links to still says v0.1**: `Status: accepted`,
  `last_reviewed: 2026-06-17`, Decision section pinning the v0.1 constant, and no amendment
  in the file. AGT's own `trace_model.py` comment points at "ADR-0032 and its 2026-07-28
  amendment for why the v0.1 URI was replaced", so whatever that amendment is, it is not in
  the document this page links to. That is the trap here, and it is the reason my first read of this
  PR was wrong: the ADR is the obvious source and it is out of date.

**Not approving this head, and the reason**

As it stands the page would tell a reader that AGT emits v0.2 while every installable AGT
emits v0.1, and it would cite a v0.1 document as the support for that.

That has a cost I could measure rather than assert. Someone who follows this page today
installs core 5.0.0 and gets a record carrying `tag:agentrust.io,2026:trace-v0.1`. Our own
checks then reject it: `TrustRecord.model_validate()` and `schema/trace-claim.json` both
refuse the profile, and the conformance suite fails it at Level 0 with two findings, the
first being TR-ENV on `eat_profile`. The page would have told them to expect
v0.2, so that failure arrives with nothing to explain it. That is what the approval is
waiting on, not the direction of your change.

The fix is small, and
either shape works: say what the current release does and note that `main` has moved, or keep
your sentence and say in it that the linked ADR predates the change. Both have the same
consequence for the tuple entry, which is the next section.

What is not available is a better link. Their ADR index lists 0032 as the only TRACE entry,
still under its v0.1 title, so the 2026-07-28 amendment their code cites is not published
anywhere I can find. Whichever shape you pick, the sentence has to carry that qualification
itself.

**The doc change and the guard entry are one thing**

`_CURRENT_RECORD_SURFACES` fails a listed file on any `TRACE vX.Y` label that is not the
current one, and I ran both of the sentences above through it. Both turn it red, because any
sentence that is accurate today has to name v0.1 somewhere. So your tuple entry and your doc
change are the same thing twice: right once AGT ships, early until then.

There is a third way, and it is worth naming rather than using quietly. The pattern matches
the literal `TRACE v0.1`, so a sentence that says "the `trace-v0.1` profile URI" keeps the
suite green with your entry in place. I ran that too: 11 passed, 1 skipped. It works by not writing what
the guard looks for, which is not the same as the guard being satisfied.

The entry does work, and I checked rather than assumed: with your `agt.md` line reverted and
the entry in place,
`test_no_current_surface_labels_the_record_with_a_superseded_version[docs/integration/agt.md]`
fails; with both reverted the suite is green. On your head, 742 passed and 1 skipped, the skip
being the environment guard at `test_version.py:90`.

**One smaller thing**

Neither commit carries a `Signed-off-by` trailer, and CONTRIBUTING asks for DCO sign-off on
every commit. For two commits, `git rebase --signoff HEAD~2` and a force-with-lease push does
it. Nothing in `.github/workflows` checks this, so it would not have been flagged for you.

imran-siddique (approved):
Approving to release the hold gate. Four workflows were sitting in `action_required`, so trace-spec's CI had not run: one check reported where this repo normally runs six. Report to follow.

**Your facts check out, and I verified them at the published artifact rather than the tree**, because a claim about what a released package emits is not answerable from a repository. Downloaded `agent_governance_toolkit_core-5.0.0` from PyPI:

- `agentmesh/governance/trace_sink.py` writes `"eat_profile": "tag:agentrust.io,2026:trace-v0.1"`
- metadata pins `Requires-Dist: agentrust-trace<0.3.0,>=0.2.0`

Both exactly as you state. Checking the wheel rather than AGT's `main` is the right instinct and it is the one most people skip.

**One correction, for the record rather than for the diff.** Your description says the doc "said AGT emits TRACE v0.2 Trust Records. That was wrong." It did not. `main` says v0.1 today, it said v0.1 before this PR, and your diff keeps v0.1.

What the change actually does is add scoping, and that is more useful than the correction you describe. Before: an unqualified "AGT emits TRACE v0.1 Trust Records", which a reader cannot tell is current, stale, or about to change. After: pinned to release 5.0.0, with `main` having moved to v0.2 unshipped, and the linked ADR 0032 still describing the v0.1 design. Three facts a reader needs and could not previously get.

I am flagging it because the PR body becomes the commit message, and a history entry claiming to fix an error that was never there sends the next person looking for a regression that did not happen.

Worth knowing this file came up yesterday from the other direction. While reviewing #207, a sweep of every "TRACE v0.1" string in the repository, I listed `docs/integration/agt.md` as correctly excluded because it genuinely describes v0.1 behaviour rather than mislabelling v0.2. That reasoning was right and incomplete: the statement was true and unqualified, and unqualified is how a true statement goes stale without anyone editing it. This adds the qualifier that sweep did not think to ask for.

Noting the ADR has not been updated for AGT's own move is the detail I would have missed. It means a reader following the link lands on a v0.1 design document that AGT's `main` has already left behind, and that is a trail this document cannot fix but can warn about.

CI report to follow.

imran-siddique:
CI report, as promised. Releasing the four held workflows took this from one check to five, all green. `gate` is our maintainer hold rather than a result.

Merging, with the note from my review standing: the value here is the scoping, not a v0.2 correction. Thanks for checking the published wheel rather than AGT's `main`.

lywinged (approved):
Re-checked on `59b0ce2`, against the released wheel, the linked ADR and AGT's `main`. All three
clauses hold, so a reader who follows this page today installs 5.0.0, gets a record carrying the
v0.1 profile, and the page has already told them that is what to expect. That was the whole of
what I was holding on.

One thing worth stating so it does not read as a retreat. Reverting the
`_CURRENT_RECORD_SURFACES` entry was the right call rather than a concession: the guard fails a
listed file for naming a superseded version, and any sentence that is accurate today has to name
v0.1 somewhere, so the entry and an accurate sentence cannot both be in the tree. It belongs in
the same PR that updates this page once AGT ships.

Merged onto `main` at `39fa9c5`, which now carries #219: clean merge, 803 passed and 1 skipped,
identical to that base's own baseline. Net diff is one line in one file.

Approving.

PR #218, opened 2026-08-25
fix(readme and index.md): Quick start code sample doesn't run

rajnisht7:
## What this changes

The README's and index.md's Quick start example crashes immediately if executed via copy-paste it constructs `TrustRecord(...)` with fields like `model_id`, `platform`, and `policy_hash` that don't exist on the model (they're nested under `model.*`, `runtime.*`, `policy.*`) and skips several required fields, passes a `TrustRecord` object to `sign_record()` which expects a dict, and references an undefined `signing_key` variable. Running it raises 12 pydantic validation errors.

Replaced it with a minimal working example (verified by actually running it) that mirrors the pattern already used correctly in `docs/quickstart.md`.

Also fixed one row in the "What a Trust Record proves" table: it referenced `anchoring.receipt_uri`, a field that doesn't exist anywhere in the schema or model. The actual field is `transparency`.

## Type of change

- [ ] Schema change
- [ ] Example addition

## Spec section

None

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Each question maps to a claim a third party can check for themselves.
 | Under which policy? | `policy.bundle_hash` + `policy.enforcement_mode` |
 | What data did it touch? | `data_class` |
 | Which tools were called? | `tool_transcript.hash` + `tool_transcript.call_count` |
-| Is the record independently anchored? | `anchoring.receipt_uri` (SCITT) |
+| Is the record independently anchored? | `transparency` (SCITT receipt URI) |
 
 ## Quick start
 
@@ -55,16 +55,26 @@ pip install agentrust-trace
 ```
 
 ```python
-from agentrust_trace import TrustRecord, sign_record
+import time
+from agentrust_trace import generate_key, sign_record
 
-record = TrustRecord(
-    subject="spiffe://trust.example.org/agent/payments-processor",
-    model_id="claude-sonnet-4-6",
-    platform="amd-sev-snp",
-    policy_hash="sha256:b2c3d4...",
-)
-signed = sign_record(record, key=signing_key)
+key = generate_key()
+
+record = {
+    "eat_profile": "tag:agentrust-io.com,2026:trace-v0.2",
+    "iat": int(time.time()),
+    "subject": "spiffe://trust.example.org/agent/payments-processor",
+    "model": {"provider": "anthropic", "model_id": "claude-sonnet-4-6"},
+    "runtime": {"platform": "amd-sev-snp", "measurement": "sha256:" + "0" * 64},
+    "policy": {"bundle_hash": "sha256:" + "b" * 64, "enforcement_mode": "enforce"},
+    "data_class": "confidential",
+    "build_provenance": {"slsa_level": 1, "digest": "sha256:" + "e" * 64},
+    "appraisal": {"status": "none", "verifier": "https://verifier.example.org"},
+}
+
+signed = sign_record(record, key)
 ```
+See the [Quickstart guide](https://trace.agentrust-io.com/quickstart/) for key persistence, validation, and anchoring the record to a transparency log.
 
 ## Resources
 
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ record = {
 
 signed = sign_record(record, key)
 ```
-See the [Quickstart guide](https://trace.agentrust-io.com/quickstart/) for key persistence, validation, and anchoring the record to a transparency log.
+See the [Quickstart guide](https://trace.agentrust-io.com/docs/quickstart/) for key persistence, validation, and anchoring the record to a transparency log.
 
 ## Resources
 
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ record = {
     "iat": int(time.time()),
     "subject": "spiffe://trust.example.org/agent/payments-processor",
     "model": {"provider": "anthropic", "model_id": "claude-sonnet-4-6"},
-    "runtime": {"platform": "amd-sev-snp", "measurement": "sha256:" + "0" * 64},
+    "runtime": {"platform": "software-only", "measurement": "sha256:" + "0" * 64},
     "policy": {"bundle_hash": "sha256:" + "b" * 64, "enforcement_mode": "enforce"},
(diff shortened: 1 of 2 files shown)


lywinged (commented):
Thanks for this, and for running it before opening. I reproduced both halves before commenting.

Verified

On a clean checkout of 95da3fd, against the repo's own src/agentrust_trace:

The old snippet raises pydantic.ValidationError with exactly 12 errors, the count you gave.
The new snippet runs clean, and the record it produces passes TrustRecord.model_validate(), validates against schema/trace-claim.json with zero errors, and its signature checks out: verify_record(signed, key.public_key()) returns without raising.
anchoring is not a field anywhere in the schema. transparency is, as a string with format: uri, so the table row is now correct.

**Not approving this head, and the reason**

The fix itself is right, and what holds the approval is the two items below. The first is a defect this PR adds that was not there before it. The second is the repo's own merge requirement.

The new link 404s. https://trace.agentrust-io.com/quickstart/ returns 404, and the page that exists is https://trace.agentrust-io.com/docs/quickstart/, which returns 200 and is the one the site's own sitemap.xml lists. The build explains it: docs.yml copies docs/ into the build directory as docs/, mkdocs.yml sets docs_dir: ., and no redirect plugin is configured, so the published path keeps the docs/ prefix.

curl -o /dev/null -w '%{http_code}\n' https://trace.agentrust-io.com/quickstart/
curl -o /dev/null -w '%{http_code}\n' https://trace.agentrust-io.com/docs/quickstart/
95da3fd has no Signed-off-by trailer, and CONTRIBUTING asks for DCO sign-off on every commit. git commit --amend -s and a force-with-lease push fixes it. Nothing in .github/workflows checks this, so it would not have been flagged for you.

One request, and it is your call whether it belongs in this PR

You took the new example from the pattern "already used correctly in docs/quickstart.md", and that reasoning reaches one surface further. index.md still carries both defects, and it is the page the site publishes as Home. Its Quick start block is byte-identical to the one you replaced and fails with the same 12 errors, and line 45 has the same anchoring.receipt_uri row. README.md is in exclude_docs and the docs workflow never copies it, while index.md is copied to the build root, so as it stands this PR fixes the GitHub view and leaves the site homepage unchanged. It is the same edit a second time. If you would rather keep this PR to README.md alone, say so and the other surface can be tracked separately.

Optional, and it predates your change

The example claims runtime.platform: "amd-sev-snp" with an all-zero measurement and appraisal.status: "none". docs/trust-levels.md treats an all-zero measurement as the convention for Level 0, which requires software-only, and asks Level 1 for a non-zero measurement plus appraisal.status: "affirming". docs/quickstart.md, which you mirrored, uses software-only for that reason. Carrying that value across would make the snippet a valid Level 0 record rather than a hardware claim with a placeholder measurement.

Those two are what unblock the approval. Push them and I will re-run the same checks on the new head.

imran-siddique (approved):
Approving to release the hold gate. Seven workflows were in `action_required`, so trace-spec's CI had not run: one check reported where this repo normally runs six. Report to follow.

**Ran both examples rather than reading them.**

The old one:

```
ValidationError: 12 validation errors for TrustRecord
```

Twelve, exactly as you said. Your replacement runs clean against current `main`, produces all eleven expected keys including the `cnf` that `sign_record` injects, and I took it one step further than you did: the resulting record passes `verify_record` against the generated key. So the sample is not merely syntactically valid, it produces something that verifies.

The table fix is real too. `anchoring` is not a field on `TrustRecord`; `transparency` is.

**This is the highest-traffic broken thing in the repository.** The Quick start is the first code anyone copies, it is on both the README and the site index, and it has been failing on paste. Someone evaluating TRACE hits this before anything else we have written, and a twelve-error traceback is a strong statement about a project whose entire subject is verifiable correctness.

Two things in your replacement that go beyond fixing the crash, and both are the right call.

**`generate_key()` instead of an undefined `signing_key`.** The old sample could not have run even with the field names corrected, because the key came from nowhere. A quick start that requires the reader to already know the thing they came to learn is not a quick start.

**`software-only` with a zeroed measurement instead of `amd-sev-snp`.** A copy-paste of the old sample produced a record asserting a hardware root of trust that the reader does not have. That is the failure mode this project exists to prevent, sitting in our own front-page example. Choosing the honest platform value for a sample is a small decision that would have been easy to get wrong in the other direction.

Pointing at the Quickstart guide for key persistence, validation and anchoring, rather than growing the sample, keeps it a quick start.

No overlap with your #217: different files, no shared added lines, checked before reviewing either.

CI report to follow.

lywinged (approved):
Re-ran the checks on `85bd058` merged onto current `main`, since the base has moved twice since
my first pass.

The old block still raises exactly 12 validation errors; the new one runs, passes
`TrustRecord.model_validate()`, validates against `schema/trace-claim.json` with zero errors,
and `verify_record(signed, key.public_key())` returns without raising. The `README.md` and
`index.md` blocks are byte-identical, 725 bytes each, so the two surfaces cannot drift apart by
someone fixing only one of them.

The link defect is gone. `https://trace.agentrust-io.com/docs/quickstart/` returns 200, and the
path I flagged still returns 404, so the version in this PR is the one that resolves.

You took the optional item as well, and it lands exactly on the documented convention.
`docs/trust-levels.md` requires `runtime.platform: "software-only"` for Level 0, calls the
all-zero measurement conventional for software-only records, and calls `appraisal.status: "none"`
correct when no hardware verifier is in the path. The snippet now matches all three, so it is a
valid Level 0 record rather than a hardware claim carrying a placeholder measurement.

The table row is right: `anchoring` is not a property in the schema and `transparency` is, a
string with `format: uri`.

One thing to retract. I listed DCO sign-off as one of two items blocking approval. I am
withdrawing that as a condition: nothing in CI checks it, and it is a merge-time decision rather
than a review one. Raising it once was fair, holding the approval on it was not.

Merged onto `main` at `5b3fafb`: clean merge, 803 passed and 1 skipped, identical to that base's
own baseline.

Approving.

lywinged (approved):

imran-siddique:
CI report, as promised. Releasing the seven held workflows took this from one check to eight, all green with nothing failing.

Merging. The Quick start now runs on paste, which it had not been doing.

PR #219, opened 2026-08-25
fix(schema): hold every integer in a signed object to the JCS safe-integer range

lywinged:
## What this changes

Section 3.2.2 serializes numbers as IEEE 754 doubles:

> Numbers are serialized in IEEE 754 double-precision format using the shortest decimal representation that round-trips.

RFC 8785 §3.2.2.3 defers this to ECMA-262 §7.1.12.1, which converts through a double, so `9007199254740992` and `9007199254740993` reach one canonical form and one signature stands for two objects. Nothing in the schemas prevented a record from carrying such a value.

**Reproduced end to end on a clean checkout of `main`.** Two records differing only in `tool_transcript.call_count`, both passing `validate_json` on that tree, both with an ordinary `iat`. A signature issued for the first, over its RFC 8785 canonical form. A verifier built the way §3.2.2 says to build one — `canonicalize` 4.0.0 on npm, Ed25519, plus the section's own maximum-age and future-skew checks — handed the second record with the first record's signature: **accepted**.

The carrier matters and the obvious one is wrong. An `iat` at 2^53 is a record dated in the year 285 million, so a verifier enforcing freshness rejects it before canonicalization and the collision never decides anything; `main`'s own `verify_record` does exactly that, reporting a record 9007197467041671 seconds in the future. `call_count` has no such rule, and neither do `origin.ingested_at`, `appraisal.timestamp`, or a `cnf.jwk` member the schema does not name.

Two independent RFC 8785 implementations disagree here and neither is safe to rely on. `canonicalize` 4.0.0 applies the algorithm as written and emits identical bytes for the pair. `rfc8785` 0.1.4, the pinned dependency, refuses both — not what §3.2.2.3 says to do, and the safer way to be wrong. A record carrying such a value gets whichever the verifier happens to have.

### Reachable four ways before this change, each measured on `main`

1. **Declared record fields.** `iat`, `origin.ingested_at`, `tool_transcript.call_count`, `appraisal.timestamp` were typed `integer` with no upper bound.
2. **Undeclared members.** `cnf.jwk` is open because RFC 7517 permits members this schema does not name, and `additionalProperties` was absent, which means true. A colliding integer or a float went in there, was covered by the signature, and met no constraint. Bounding the declared fields alone left this open, and the first version of this change did exactly that.
3. **The other signed artifacts.** `revoked_at`, `issued_at`, `valid_until` in the revocation schemas, and `authorized_at`, `expires_at` in the PIC/TRACE bridge, whose `sign_bridge` and `verify_bridge` call the same `_canonical_bytes` as everything else. All four collide on `main`.
4. **Objects no schema describes.** `digest_jcs` takes the digest of a caller-supplied declaration and tool-call object, and `SandboxAdapter.transcript_hash` of a decision log. Two declarations differing only in an out-of-range integer produce one digest under `canonicalize` 4.0.0. No `maximum` anywhere reaches this; only the normative sentence does.

### The range, and why it is not 2^53

RFC 8785 Appendix B note 1 names `-9007199254740991` to `9007199254740991` as a SHOULD on values interpreted as true integers. §3.2.2 raises it to a MUST and states it for any object canonicalized under that section rather than for a Trust Record alone, and points a producer needing a larger value at the JSON string type, which RFC 8785 Appendix D requires of such numbers anyway.

Reasoning about Python alone would have allowed a wider bound: 2^53 is exactly representable and no two integers in `-2^53` to `2^53` share a double, so that maximum looks sound. It is not. A validator whose only number type is the double never sees the instance value, only what it parsed to, so at 2^53 it reads `9007199254740993` as `9007199254740992`, finds it in range, and admits the one value the range exists to exclude.

Measured on ajv 8.20.0:

| `maximum` | on the wire | ajv sees | verdict |
|---|---|---|---|
| `9007199254740991` | `9007199254740993` | `9007199254740992` | reject |
| `9007199254740992` | `9007199254740993` | `9007199254740992` | **accept** |

2^53 − 1 is the widest bound a double-only validator can enforce at all, which is a stronger reason than convention. §3.2.2 now says so and forbids a profile from widening it.

### What changed

Declared integer fields across all five schemas whose signature is over an RFC 8785 canonical form carry `maximum: 9007199254740991`. Undeclared `cnf.jwk` members are held recursively to a new `$defs/canonicalizableValue` admitting strings, booleans, null, arrays, objects and in-range integers, and no `number`. Vectors `03` and `04` build on those undeclared members and still pass.

`appraisal.timestamp` had no lower bound either and takes the symmetric floor rather than a calendar one; what a sensible earliest appraisal time would be is a separate question this does not answer.

`agentrust_trace.models` already mirrored every schema constraint (`ge=0` for `call_count`, `ge=0, le=3` for `slsa_level`), so schema and model disagreed the moment the bound went into one of them. Both agree now, checked as behaviour on each rather than by comparing declarations written in different languages.

`docs/schema.md`, the field reference a producer reads, states the range under its own heading.

### Reproduction, on the standard library alone

```
python examples/canonicalization-boundary/number_divergence_repro.py
```

The measurements above were taken against V8, npm's `canonicalize` and `ajv`, none of which you have, which made them claims rather than results. The script first checks its own canonicalizer against `rfc8785` byte-for-byte on all 79 records in `examples/`, including the non-ASCII and UTF-16 key-order vectors, then shows the pair reaching one canonical form and one Ed25519 signature, why the bound is 2^53 − 1, and that the schema now refuses both. Exit code 0 means every step reproduced, and a test runs it.

### The test this replaces

`test_number_divergence_is_still_unreachable` asserted that no field is typed `number`. That is true, and it was not the whole claim the name made. `examples/canonicalization-boundary/README.md` stated the rest, correctly qualified, and the qualifier is where it came apart:

> integers up to 2^53 serialize identically everywhere

Up to 2^53 they do. Nothing held the schema to "up to 2^53". The README also said the correct response to that test failing is a vector, not an edit to the test. There is no vector to write: a positive vector is a schema-valid record, and these are exactly the records the schema now rejects. So the guard is replaced rather than extended, and the README says so.

The tests live in a new file. `test_canonicalization_boundary.py` is about the four vectors that separate a conformant canonicalizer from `json.dumps`; it had grown to 1104 lines of which nine tenths were this, under a docstring that still described the vectors. It is back to 135 lines and the number surface is in `tests/test_safe_integer_range.py`.

### Verified

803 passed, 1 skipped; `ruff check src tests scripts` clean; `mypy src/agentrust_trace` clean; on a fresh clone, online and with the network removed.

Mutation sweep, all red: each of the nineteen occurrences of the bound across the five schemas, each of the four model bounds, the `cnf.jwk` constraint in either copy, each of four ways to weaken `$defs/canonicalizableValue`, each of `anchor_bytes`'s three refusals, four weakenings of the registry-anchor correction, four ways to break the reproduction script, adding an unclassified schema, bounding an exempted one, referencing the exempted schema from source, and nine separate weakenings of the §3.2.2 text including each of the two places it names the colliding pair.

Margin, the way #124 asks for it: each of the eight defects was re-introduced one at a time and the distinct test functions that go red were counted. The record fields are separated by three to five, `cnf.jwk` by three, the nested `revoked_at` by three. The bundle's `issued_at` and the bridge's `authorized_at` were separated by one apiece, so the exhaustive sweep now runs over those two artifacts as well and every defect is separated by at least two.

### Two parts that are separable

Say the word and I will split either out.

**`spec/registry-anchor-v1.md` and `anchor_bytes`.** §0 of that document exists to warn that the anchoring layer is not JCS, and it made the same claim this change corrects in §3.2.2: that the two agree on records whose numbers are integers, which holds only inside the safe-integer range, listed under three ways they diverge where there are four. The anchor profile is worse off than JCS here, because two implementations of its own four rules disagree with each other. Measured on `tool_catalog_hash`, which runs those rules over a tool's `input_schema` where a `maximum` is ordinary content: Python writes the exact digits and gives two digests for two catalogs, and a JavaScript implementation of the same rules goes through `JSON.stringify` and gives one, so the tool-description rug-pull that hash exists to catch is undetectable to it. §1 now excludes the range, as it already excluded non-integer numbers and for the same reason, and the exclusion is executable rather than only written: `agentrust_trace.sign.anchor_bytes` performs the sorted-key serialization and refuses an excluded value with a named `UnanchorableValue`. It is here because leaving a sibling document asserting the opposite of what §3.2.2 now requires seemed worse than the wider diff.

**The AGT adapter's transcript hash is a behaviour change, and its sibling was already the other way.** Both hash a caller-supplied decision log. The sandbox adapter canonicalizes with `rfc8785` and already refused an out-of-range integer; the AGT adapter used sorted-key JSON with no check and returned a digest. On the same realistic input — an audit entry with a nanosecond timestamp and a 64-bit request id — `main` gives `IntegerDomainError` from one and `sha256:ffec42a2…` from the other. They agree now, in the direction the one that was already correct had chosen, and the error names the field and says to carry the value as a string.

### Backward compatibility

No record this repository could sign is affected: `sign_record` refuses an out-of-range integer through `rfc8785` before the schema is consulted, and the largest integer in any fixture here is `1785936000`. The one behaviour change that can reach a working integration is the AGT transcript hash above — a governance framework whose audit entries carry nanosecond timestamps or 64-bit ids now gets a named error instead of a digest no peer could reproduce.

### Not changed

`src/agentrust_trace/schema/trace-v0.1.json`, superseded and loaded by nothing: `verify_record` rejects the v0.1 profile identifier, so no record here is validated against that file. A test holds that reason to the tree and fails if any source file starts referencing it.

### Noticed while running this, not fixed here

`tests/test_sign.py:361`, `:371` and `:380` build an `iat` one second from the skew boundary they assert on, so each has a one-second margin and flakes when a second ticks over between construction and check. Measured at 0.6 ms median between the two lines, about 0.06% per run per test. One of them failed once during this work. The fix is a choice between freezing the clock and widening the margins, which is yours, and it is unrelated to this change.

## Type of change

- [x] Schema change
- [ ] Example addition

Marked non-breaking on the argument in **Backward compatibility** above rather than by assertion: the constraint is tightened, and no record that could be signed by this repository's canonicalizer is affected. If you read it as breaking, the 14-day path is yours to call and I will not argue it.

## Spec section

`spec/trace-v0.2.md` §3.2.2 — the canonical-form number bullet, the paragraph on conformant libraries, and two new paragraphs on the range and on what the rule covers.

`spec/registry-anchor-v1.md` §0 and §1 — separable, see above.

## Checklist

- [ ] Breaking changes marked with `<!-- CHANGED: #NNN — description -->` in spec text — not a breaking change
- [x] Backward compatibility statement included (above, though not marked breaking)

---


imran-siddique (approved):
This is a signature forgery, not a canonicalization tidy-up, and the write-up is the reason it is obvious rather than arguable. Reviewed it as such.

## Reproduced the mechanism rather than reading it

```
float(9007199254740992) == float(9007199254740993)  ->  True
```

Two distinct integers, one double, one canonical form, one signature standing for two objects.

**The implementation divergence is the sharper half and I confirmed it here.** The pinned `rfc8785 0.1.4` refuses both with `IntegerDomainError: exceeds safe integer domain for JSON floats`. npm `canonicalize` 4.0.0 applies the algorithm as written and emits identical bytes. You are right that neither is safe to rely on and that `rfc8785` is "the safer way to be wrong": it is not what §3.2.2.3 says to do, and a TRACE deployment's exposure currently depends on which library its verifier happened to install. That is worse than a bug in one of them, because nothing in the spec tells an implementer which behaviour they are entitled to assume.

**The carrier analysis is what makes this exploitable rather than theoretical.** Ruling out `iat` because a record at 2^53 is dated in year 285 million and freshness rejects it first, then finding `call_count`, `origin.ingested_at`, `appraisal.timestamp` and undeclared `cnf.jwk` members which carry no such rule, is the step most reports skip. A collision behind a check that already rejects it decides nothing.

## The layering is right, and it matters more than usual here

I sent back a change on agent-manifest#338 yesterday that put a safe-range refusal inside `canonicalize()`, on the grounds that RFC 8785 Appendix B note 1 says the safe range is a SHOULD and that "how numbers are used in applications does not affect the JCS algorithm", with note 2 serializing 2^68 cleanly. That ruling stands, and this PR is what the correct version looks like.

The JCS canonicalizer is untouched. The bound lives in two places that are entitled to impose it:

- **the schema**, where `canonicalizableValue` holds integers to the safe domain and, crucially, recurses through undeclared members so that RFC 7517's allowance for unnamed JWK members does not become a hole in the signed object;
- **`anchor_bytes`**, whose docstring says "Sorted-key JSON per registry-anchor-v1 section 1. Deliberately not JCS", raising `UnanchorableValue` with a named reason.

That is exactly the distinction I asked for: not inside the function claiming to implement JCS, and with its own name and its own reason where it does belong. Appendix B note 1 is not a licence to ignore the range, it is a statement that bounding it is the application's job. This is the application doing it.

The error text telling an implementer to carry the value as a JSON string, and naming nanosecond timestamps and 64-bit identifiers as the values that land there, is the difference between a rejection and a fix.

## Deleting a test that asserted a false claim

`test_number_divergence_is_still_unreachable` checked that no schema field is typed `number` and concluded divergence was unreachable. The claim was too narrow: an `integer` field diverges too, because JCS converts it through a double, and an undeclared JWK member can be anything at all. A test pinning a claim that is wrong is worse than no test, because it is read as coverage.

Replacing it with 22 tests that keep the no-`number` assertion, add the safe-domain check, cover additional members, and verify the exemptions hold in the other direction, is the right treatment. `test_the_superseded_schema_is_still_loaded_by_nothing` in particular: an exemption resting on a fact, with the fact checked, is how exemptions stop rotting.

All six checks green. Merging.

PR #221, opened 2026-08-26
fix(ci): pin sbom.yml actions to SHA and scope permissions to the job

rajnisht7:
## What this changes

sbom.yml was running anchore/sbom-action@v0 which is a floating tag in a job that also holds id-token: write. If that tag ever pointed to something malicious, whatever ran could mint an OIDC token and push to this repo. Pinned it, along with actions/checkout and actions/upload-artifact, to exact commit SHAs.

Also moved contents: write and id-token: write from the top of the file down into the sbom job itself, so they only apply where they're actually needed instead of to the whole workflow.

Part of the CI hardening pass in agentrust-io/.github#30. 

## Type of change

- [ ] Schema change
- [ ] Example addition

## Spec section

None

## Checklist

### DIFF
--- a/.github/workflows/sbom.yml
+++ b/.github/workflows/sbom.yml
@@ -9,22 +9,21 @@ on:
 env:
   FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: "true"
 
-permissions:
-  contents: write
-  id-token: write
-
 jobs:
   sbom:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      id-token: write
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - name: Generate SBOM
-        uses: anchore/sbom-action@v0
+        uses: anchore/sbom-action@e22c389904149dbc22b58101806040fa8d37a610 # v0.24.0
         with:
           format: spdx-json
           output-file: sbom.spdx.json
       - name: Upload SBOM artifact
-        uses: actions/upload-artifact@v7
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           name: sbom
           path: sbom.spdx.json


lywinged (approved):
Checked the pins rather than trusting them, since a wrong SHA in a pinning diff is the whole
failure mode.

All three resolve to their claimed tags, with annotated tags peeled: `3d3c42e5` is
`actions/checkout` v7.0.1, `e22c3899` is `anchore/sbom-action` v0.24.0, `043fb46d` is
`actions/upload-artifact` v7.0.1. And `@v7`, `@v0` and `@v7` **currently resolve to those same
three commits**, so this freezes today's behaviour rather than carrying a version change inside
a pinning change.

The form matches what this repository already uses: `actions/github-script@3a2844b7… # v9.0.0`
in `require-maintainer-approval.yml` and `actions/checkout@9c091bb2… # v7.0.0` in
`contributor-check.yml`.

The permission set is unchanged, `contents: write` and `id-token: write` before and after, so
the second half is relocation and not a scope change. Checked separately, since a moved block is
where one quietly widens.

**Worth knowing what happened to the last pin here.** `contributor-check.yml` pins
`actions/checkout` at v7.0.0. v7.0.1 was tagged on 2026-07-17, that file was last edited on
2026-08-11, and it still carries v7.0.0. Dependabot's `github-actions` updater is configured
weekly and is demonstrably alive, having opened the `codeql-action` 4.37.7 to 4.37.8 bump, but
that reference is a tag and this one is a SHA. I cannot see why from here, so that is an
observation rather than a diagnosis. Practically, these three will sit where you put them until
someone moves them, which is the trade pinning makes and a fair one on a job holding
`id-token: write`.

**One thing worth adding, here or in a follow-up.** Moving `permissions` to the job is right,
but deleting the top-level block leaves the file without a floor. `scorecard.yml` is the only
other workflow here that scopes per job, and it keeps `permissions: read-all` above `jobs:` and
elevates beneath it. Nothing changes today, since `sbom` is the only job and declares its own.
The exposure is the next job added to this file: it would inherit the repository's default token
permissions, which the workflow no longer states and a reader cannot see from it. One line above
`jobs:` restores that.

Approving as it stands.

imran-siddique (approved):
Verified every SHA at source rather than trusting the version comments, since a wrong SHA in a pinning PR is worse than no pin: it looks pinned, passes review, and nobody re-checks it.

| pin | tag | result |
|---|---|---|
| `actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1` | v7.0.1 | exact |
| `actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065` | v5.6.0 | exact |
| `actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a` | v7.0.1 | exact |
| `anchore/sbom-action@e22c389904149dbc22b58101806040fa8d37a610` | v0.24.0 | exact |

Annotated tags dereferenced to their commits where needed, which is the step that makes a codeql-style pin look wrong when it is right.

**The permissions change is the part I would have asked for if it were missing.** Moving `contents: write` and `id-token: write` from workflow scope to job scope narrows them to the job that needs them, so every other job in that file drops to the default. `id-token: write` in particular is the one worth scoping tightly, since it mints OIDC tokens. Doing that in the same PR as the pinning is coherent rather than scope creep: both are about what a compromised step can reach.

**Keeping the `# vX.Y.Z` comment beside each SHA** is what makes this maintainable. A bare SHA is unreadable and gets bumped by someone pasting a newer one without checking what it is. Saying explicitly in the `.github` note that the SHA and the comment must be updated together is the instruction that keeps them from drifting apart.

Replicating the identical change across ca2a, cmcp and trace-spec as separate PRs rather than one is right for repos that do not share a workflow file. They are byte-identical, which I checked.

Merged.

PR #222, opened 2026-08-26
fix(agt-adapter): use JCS, not the registry-anchor format, for `tool_transcript.hash`

rajnisht7:
## What this changes

TraceAGTAdapter._transcript_hash() was hashing tool_transcript.hash using anchor_bytes() (the registry-anchor sorted-key format, scoped by spec/registry-anchor-v1.md §0 to the transparency claim only), while its sibling TraceSandboxAdapter.transcript_hash() already used rfc8785.dumps() (JCS) for the same field, which is what docs/schema.md and docs/integration/agt.md call "canonical JSON."

This meant the two adapters produced different digests for identical audit entries, and the concrete bug anchor_bytes() rejects any non-integer number, so any AGT audit entry carrying an ordinary float (a fractional-second timestamp, a latency, a risk score, all routine in a real Cedar/AGT audit log) made build_trust_record() raise UnanchorableValue instead of returning a record.

This PR switches TraceAGTAdapter._transcript_hash() to rfc8785.dumps(), matching TraceSandboxAdapter exactly. Updates the existing test that had pinned the old (incorrect) digest as "correct," adds a regression test for floats in audit entries, adds a parity test between the two adapters, and tightens an existing integer-range test so it asserts genuine agreement (same exception type, same digest) instead of "both happen to refuse, differently." No spec or schema change both adapters already targeted the same documented field; this makes the implementation match the documentation and its sibling adapter.

## Type of change

- [ ] Schema change
- [ ] Example addition

## Spec section

No spec section changes.

## Checklist


lywinged (approved):
Confirmed, and thank you — this is a real regression and it's mine, from #219.

On merged `main`, an ordinary AGT audit entry with a fractional-second timestamp can no longer become a Trust Record:

    {"decided_at": 1750000000.482, "risk_score": 0.13}
    → UnanchorableValue: $[0].decided_at is a non-integer number

#219 closed the out-of-range-integer gap by routing `_transcript_hash` through `anchor_bytes` — the wrong profile. registry-anchor-v1 §0 scopes sorted-key JSON to the `transparency` anchor leaf, while `tool_transcript.hash` is a record-internal digest that `docs/integration/agt.md` and `TraceSandboxAdapter` both already treat as JCS. Moving it to `rfc8785` and matching the sibling adapter is what it should have been from the start.

Ran your branch here before approving: 805 passed, ruff and mypy clean, and the collision protection holds — an out-of-range integer, including one nested two levels down, still raises `IntegerDomainError` from both adapters.

Two notes for the record, neither a blocker:

- The AGT digest of an audit list containing non-ASCII text changes under this fix (JCS emits the character; the old form escaped it). Correct, since the old value was never reproducible cross-language, but worth a line in the changelog entry.
- `df45260` has no DCO sign-off, so the check will want that before merge.

---

imran-siddique (approved):
Reproduced the break on `main` before reading the argument, because a claim that the default path raises is either true or it is not:

```
TraceAGTAdapter._transcript_hash([{"action": "tool_call", "latency_s": 0.42, "risk": 0.9}])
-> UnanchorableValue: $[0].latency_s is a non-integer number (0.42)
```

A fractional-second timestamp, a latency, or a risk score in an AGT audit entry breaks `build_trust_record()` today. That is not an edge case, it is what a Cedar audit log carries.

**You are right about which canonicalization that field takes, and three independent things say so.**

- `TraceSandboxAdapter.transcript_hash`'s own docstring: "Using the same canonicalisation as the signature pre-image."
- `docs/schema.md` and `docs/integration/agt.md` both call it canonical JSON.
- `spec/registry-anchor-v1.md` §0 scopes the anchor format to the transparency claim.

`tool_transcript.hash` sits inside the signed record, so it belongs on the pre-image side. The AGT adapter was using the anchor-leaf profile for a field that is not an anchor leaf, and the comment above it asserting "Anchor format, not JCS" was confidently wrong rather than accidentally wrong, which is why nobody had questioned it.

**Verified the parity claim rather than the parity test.** With a float and a non-ASCII string in the same entry, both adapters now return `sha256:727b53c2574166a9431ca08850e4d204ff3ce566447f3e518430ba55fa5df19b`. Identical bytes hashed, identical digest.

**On the test you changed, you were right to change it and the reason matters.** #219's version was named `test_the_two_transcript_hashes_now_agree_on_the_same_input` and asserted only that both adapters *refuse*, with different exception types. "Both happen to reject this" is not agreement, and a test whose name claims agreement while asserting divergence is worse than no test, because the name is what the next reader trusts. Asserting the same digest and the same exception type is the thing the name always promised.

**One piece of context you would not have had.** #219 merged yesterday and added `_reject_unanchorable` to `anchor_bytes`, for good reasons: it closed a real signature collision. It did not create the divergence between these two adapters, which predates it, but it is what turned a silently different digest into a loud failure on this path. So this is not a regression in your sense or in theirs, it is a latent inconsistency that #219 made visible, and the visible version is the one that got fixed.

@lywinged, flagging you since this reverses part of #219's intent on the AGT side. The collision fix and the schema bounds are untouched; only the transcript-hash canonicalization moves, and it moves toward the sibling adapter rather than away from it.

All six checks green. Merging.

PR #229, opened 2026-08-27
fix(provenance): tool_catalog_hash crashes on a malformed tools list

rajnisht7:
## What this changes

check_tool_catalog(record, tools) is the function which this module's own docstring calls "the step that catches a live attack," because tools there is what the MCP server actually returned when asked and the untrusted party this check exists to catch. It passes tools straight through to tool_catalog_hash(), which iterated it and called .get(...) on each entry with no check that tools was a list or that its entries were objects.

but it crashes when tools itself being a non-list (a string, None, an int, a dict) raised AttributeError or TypeError depending on what iterating it did; a well-formed list with one malformed entry (a string, None, an int, a nested list) raised AttributeError on that entry's .get() call. Either way the caller got a crash instead of the documented ProvenanceError, on exactly the input a misbehaving maliciously or just buggily server controls.

## Type of change

- [ ] Schema change
- [ ] Example addition

## Spec section

None

## Checklist


Qiang-Xu (dismissed):
Looks good to me, thanks!

imran-siddique (approved):
Third in the set and the one closest to the attack, so worth taking.

`check_tool_catalog(record, tools)` exists because a verifier that never obtains what the server said to *you* has checked a document against itself. `tools` is that input: the untrusted party's own claim about itself, passed straight through to `tool_catalog_hash()`. It iterated and called `.get()` on each entry with no shape check, so a non-list crashed on iteration and a bad entry crashed on `.get()`. Both raised `AttributeError` or `TypeError` where the module promises `ProvenanceError`.

Two details done right:

The list itself and its entries are checked separately, and the entry error names the index. Debugging "one of your 40 tools is malformed" without knowing which is unpleasant.

The tests parametrize both shapes rather than picking one representative, including a nested list as an entry, which is the case a naive `isinstance(t, (dict, list))` would have let through.

CI green. I merged main in to clear a CHANGELOG conflict from #225 and #227 landing first, and took main's copy of the shared line so the new no-dashes check passes. Nothing of yours changed.

Three solid fail-closed fixes in a row on verification paths. Appreciated.

Issue #190, opened 2026-08-19
spec: where a verifier records an unresolvable citation: the cross-surface question from #187's review


opento-suggestions:
#187 has merged. [Its review](https://github.com/agentrust-io/trace-spec/pull/187#pullrequestreview-4964796933) asked for a tracking marker on one cross-cutting item — ["an explicit line in §3.2.3 — that the field for this is deferred, with an issue number"](https://github.com/agentrust-io/trace-spec/pull/187#issuecomment-5333001810) — and that issue does not yet exist. This is that issue. It records the question as deferred; it does not propose the answer.

**The question, as the review put it:** when a record cites something a verifier cannot resolve, what does the verifier record, and where.

**Four surfaces currently answer it three different ways, or not at all:**

- `build_provenance` — answered and merged (#173): the depth actually achieved is recorded in a dedicated field, `appraisal.provenance_depth_verified`, and `appraisal.status` moves to `contraindicated` only when a configured floor is not met.
- Revocation — merged (#187): the spec text states "Neither may be reported as an affirming appraisal" with no field or value naming what is reported instead. The review walks the enum: `contraindicated` reports evidence failing, `none` reports no appraisal performed, `warning` reports evidence passing — each misreports a check that was merely unresolvable.
- Delegation links — open (#184): the same boundary is classed `unverifiable`, deliberately distinct from both invalid kinds, on its own axis.
- `appraisal.policy_ref` — raised on #66: whether resolution of the appraisal policy citation falls under the same discipline.

Two conformant verifiers implementing today's merged text can record different things for the same unresolvable check on the same record. That is the divergence the depth work closed for `build_provenance`, reappearing on every other surface that cites by reference.

**Scope of this issue:** a home for the decision, not the decision. The review already names the merged shape that fits — a recorded field plus a policy floor, per #173 — and whether a field lands, and where, is a schema question for the editorial process. What this number enables immediately:

- normative text can cite it where the merged prose currently says only what may not be reported;
- conformance vectors on both active sides — the revocation vectors claimed in #187's triage, and the appraisal-resolution vectors accepted on #66 — can mark their unresolvable-case vectors as deferred pending this issue, rather than each coining an outcome value and making the count of conventions five and six.

If the editorial preference is to fold this into an existing tracker instead, closing this as a duplicate with a pointer serves the same purpose.

lywinged:
[...]

PR #235, opened 2026-08-28
fix(docs): tracker number and two swapped issue/pull link paths

opento-suggestions:
## What this changes

Two reference targets in `docs/verification-outcome-statements.md`, carrying three defects of two kinds: one wrong referent, and two link paths swapped between `/issues/` and `/pull/`.

1. The open-questions bullet pointed the `appraisal.policy_ref` reproducibility question at [agentrust-io/trace-spec#187](https://github.com/agentrust-io/trace-spec/pull/187). That link resolves, but it names the wrong item: #187 is a merged pull request (its changes include the revocation bundle schemas), and the open issue tracking the question is [agentrust-io/trace-spec#190](https://github.com/agentrust-io/trace-spec/issues/190). The reference now names the open issue and links it as an issue.
2. The same line reached #187 through an `/issues/` path, and two lines up, [agentrust-io/trace-spec#116](https://github.com/agentrust-io/trace-spec/issues/116) was linked through a `/pull/` path; #116 is an open issue. Each swapped path resolves only by GitHub's redirect, so both read as correct on click-through. The path on the agentrust-io/trace-spec#116 line is corrected in place, and its number was already right. The path on the other line is not corrected: the link to agentrust-io/trace-spec#187 is removed together with its referent, so no corrected link to it survives in the result.

Every word outside the two changed links is unchanged. The version-boundary line therefore asserts exactly what it asserted before. The `appraisal.policy_ref` line does not: "tracked in" names the item that tracks the question, so pointing it at agentrust-io/trace-spec#190 instead of agentrust-io/trace-spec#187 changes what that line asserts.

## Type of change
- [ ] Non-breaking spec change
- [ ] Breaking spec change
- [ ] Schema change
- [ ] Example addition

## Spec section
None; `docs/` only.

## Checklist
- [ ] `CHANGELOG.md` updated (for any normative change): not applicable, non-normative
- [ ] Breaking changes marked: not applicable
- [ ] Backward compatibility statement: not applicable

---

*Amended 2026-08-29: two claims this body made about the diff were wrong as written and have been replaced. Prompted by lywinged's review 5054851341.*

### DIFF
--- a/docs/verification-outcome-statements.md
+++ b/docs/verification-outcome-statements.md
@@ -49,9 +49,9 @@ Moving these into the record would be a schema change, which is out of scope for
 
 ## Open questions
 
-- Rows describing approval-bearing receipts - approval preceding execution, applicable approver authority, scope covering the action taken - depend on profile work routed through [#191](https://github.com/agentrust-io/trace-spec/issues/191) and sit behind the [#116](https://github.com/agentrust-io/trace-spec/pull/116) version boundary. Until that lands, they remain open questions rather than settled guidance.
+- Rows describing approval-bearing receipts - approval preceding execution, applicable approver authority, scope covering the action taken - depend on profile work routed through [#191](https://github.com/agentrust-io/trace-spec/issues/191) and sit behind the [#116](https://github.com/agentrust-io/trace-spec/issues/116) version boundary. Until that lands, they remain open questions rather than settled guidance.
 - Comparability of `transitive` across verifiers stays open until the coverage-URI follow-up named in [verification.md](verification.md#transitive-is-a-floor-on-effort-not-a-comparable-claim) settles. Until then, a `transitive` value supports an effort-floor statement only, not a claim that two verifiers covered identical dependency sets.
-- Whether `appraisal.policy_ref` gains reproducibility guarantees (a digest binding for the referenced appraisal policy) is an unresolved design question tracked in [#187](https://github.com/agentrust-io/trace-spec/issues/187). Until then, a bounded statement cites the appraisal policy by reference, and second-verifier reproduction of an appraisal verdict is not guaranteed.
+- Whether `appraisal.policy_ref` gains reproducibility guarantees (a digest binding for the referenced appraisal policy) is an unresolved design question tracked in [#190](https://github.com/agentrust-io/trace-spec/issues/190). Until then, a bounded statement cites the appraisal policy by reference, and second-verifier reproduction of an appraisal verdict is not guaranteed.
 
 ## Related
 


lywinged (commented):
Both path corrections are right, and I checked whether they are complete rather than just correct. The referent change is a different kind of edit from the other two, and I think it needs a decision from whoever set #190's scope.

## 1. The two swapped paths, and the fact that they are the only two

`refs/pull/N/head` exists for every pull request and for no issue, so `git ls-remote` answers "PR or issue" for a number without opening it:

```
git ls-remote https://github.com/agentrust-io/trace-spec 'refs/pull/*/head'
```

Running that against every GitHub link in the repository: 25 links, 13 distinct numbers.

| | |
|---|---|
| path already correct | #34, #50, #67, #95, #97, #111, #124, #166, #191, #198, and #66 |
| path swapped | **#116** (an issue reached through `/pull/`), **#187** (a pull request reached through `/issues/`) |

Exactly two, and they are the two this PR touches. There is no third one anywhere in the tree, so this closes the class rather than two instances of it.

## 2. The referent is a different question from the one #190 holds

#190 states its scope in its first lines:

> The question, as the review put it: when a record cites something a verifier cannot resolve, what does the verifier record, and where.

> This is that issue. It records the question as deferred; it does not propose the answer.

Its four-surface list is scoped the same way. The `appraisal.policy_ref` entry asks whether *resolution* of the appraisal policy citation falls under that same discipline.

The sentence being repointed asks something else:

> Whether `appraisal.policy_ref` gains reproducibility guarantees (a digest binding for the referenced appraisal policy) is an unresolved design question tracked in [#187]. Until then, a bounded statement cites the appraisal policy by reference, and second-verifier reproduction of an appraisal verdict is not guaranteed.

A digest binding is about a second verifier reading the same policy bytes. #190 is about what a verifier records when it cannot read them at all. Same field, adjacent questions, different answers.

That reading is not mine alone. [The 2026-08-26 comment on #190](https://github.com/agentrust-io/trace-spec/issues/190#issuecomment-5420424682) says of this very page, merged as #215:

> It does **not** answer the normative question this issue holds, which is *where* a verifier records an unresolvable citation. It does answer the adjacent one, which is what a reader may conclude when it happens, and it answers it across surfaces rather than per-surface.

So the page and #190 are already on record as adjacent. This edit points a line of the page at the issue holding the other half.

## 3. Three ways out, any of which I would be glad to see

- Narrow the sentence to the half #190 does hold, for example "resolution of this citation is tracked in #190", and leave the digest question where it was raised.
- Point the digest question at **#66**, where it was raised. Line 61 of the same file already cites #66 as "the discussion this page grew out of", so the page would then be internally consistent.
- Keep #187 with the corrected `/pull/` path and change "tracked in" to "raised in". A merged pull request is a poor tracker, which is the part of this PR's reasoning I agree with; it is a fine citation for where a question was raised.

## 4. Two small notes on the description

"Both paths corrected" is true of #116. The #187 path is not corrected in the diff, it is removed along with its referent, so if the referent stays at #190 the second path fix is not in the result.

"No prose changes; the sentences' claims are untouched" holds for the #116 line. On the other line the referent is the claim: "tracked in X" asserts that X tracks it, and the diff changes what is asserted. That is also what makes `Editorial: no normative effect` the right box for one edit and a stretch for the other.

The `#116` line is ready as it stands. The other line is the only open point, and it is a scope call rather than a link fix.

imran-siddique (approved):
Verified all three reference targets against the API rather than reading the diff:

- `#116` is an issue, so the `/pull/116` path was wrong.
- `#187` is a closed pull request titled "spec: anchor key revocation to log entry ordering, not iat", which has nothing to do with `appraisal.policy_ref`. The wrong-referent call is right, and it is the defect worth catching here: the link resolved, so nothing was broken enough to notice.
- `#190` is open and explicitly lists `appraisal.policy_ref` among its four surfaces, so it is a correct target.

One non-blocking note: #190 records that policy_ref was "raised on #66", so #66 would also have been a defensible target. #190 is the better one because it is scoped to the citation-resolution discipline this bullet is about.

Approving. Releasing the three held workflow runs now.

opento-suggestions:
Point 4 is right. Both sentences were false as written.

"Both paths corrected" was false. Only the path on the agentrust-io/trace-spec#116 line is corrected in place. The other path is not corrected at all: the link is removed together with its referent, so no corrected link to agentrust-io/trace-spec#187 survives in the result.

"No prose changes; the sentences' claims are untouched" was false for the second bullet. Every word outside the two changed links held still, but "tracked in" names the item that tracks the question, so pointing that line at agentrust-io/trace-spec#190 instead changes what it asserts. The claim holds for the version-boundary line only.

The body has been amended today so both claims are true of the diff, dated in the body itself.

On the census: reproducing {116, 187} independently as the complete swapped set, with `refs/pull/N/head` as the ground truth for which is which, makes this the whole class across the tree rather than the two instances that happened to get noticed. That is a stronger claim than the description made, and it came from the review.

On points 2 and 3, the scope call. The distinction holds as drawn. This PR's description carries the same conflation: "the open issue tracking the question is" names agentrust-io/trace-spec#190, which is exact for the resolution half and loose for the digest half. That sentence stays as written because the approval endorsed the target, but its looseness is conceded here.

Since the review addresses the scope call to whoever set #190's scope: that scoping was deliberate when I filed it, and it is the same separation this review draws. The comment that raised the digest question on agentrust-io/trace-spec#66 (comment 5332456021) asked whether it was better raised separately from resolution; #190 was scoped to the resolution half for that reason, and the digest question stayed where it was raised. The reply there placed the decision on agentrust-io/trace-spec#187, which is merged and, per review 5055911365, does not concern appraisal.policy_ref. Neither location is a tracker for it, and a search of the open items today finds none that holds it. Filing one is a separate change from a link fix. As for where the line points meanwhile, review 5055911365 made that call: #66 defensible, #190 better. The approved diff stands.

On the approval: checking the three targets against the API rather than reading the diff is the check that catches a wrong referent, since the link itself resolved.

One measurement, in case the red `gate` is unexpected. The two `pull_request_review` re-evaluations on this head, runs 33208478214 and 33221861412, both concluded `action_required` and did not execute, so the `gate` check-run still predates both reviews.

PR #236, opened 2026-08-28
fix: nine defects in the reference implementation, found by a package wide sweep

lywinged:
## What this changes

Nine defects in the reference implementation, found by sweeping the package rather
than by reading any one function. Each is one commit, so the series can be split,
reordered or partly taken. `main` goes from 828 passed to 1015 passed, 1 skipped.

| # | commit | what was wrong |
|---|---|---|
| 1 | `fix(sign)` | `jwk_thumbprint` and `verify_record` read a member off their argument before establishing its shape, so a string, a number, `None`, a list or a bool raised `AttributeError`. `verify_record`'s docstring says every rejection other than a bad signature is a `ValueError`, and a caller written against that contract does not catch it. |
| 2 | `fix` | The same shape in `provenance.verify_record`, `provenance.check_tool_catalog` and `content_marking.verify_assertion`. The third was worse than a crash: it never checked that `record_bytes` were bytes, and `bytes(5)` is five zero bytes, so an int was hashed, failed to match, and the caller was told the record at the URL had changed. That is a specific accusation about somebody else's server, and it was false. |
| 3 | `fix(sign)` | The revocation check failed open on a non bool answer. `RevocationStore` is `Callable[[str], bool]` and the return value was read by truthiness, so `None`, `""`, `0` and `[]` all read as "not revoked". `None` is what a CRL, status or SCITT lookup returns when its author handled the 200 and forgot the rest, which is exactly the outage the existing `except` clause was written to survive. |
| 4 | `fix` | `docs/quickstart.md` writes an unencrypted Ed25519 private key to `trace-key.pem` under a comment reading "keep secure, never commit or log". The comment was the whole of the enforcement. A reader following the quickstart inside a clone was one `git add -A` from committing their signing key. |
| 5 | `fix` | Six public functions raised exceptions no module documents. `key_to_jwk` leaked `AttributeError` on all seventeen probe inputs including the public key, which is the plausible mistake for a function whose name reads as "turn a key into a JWK" and whose result is the public JWK. The sweep that finds them is now committed rather than described. |
| 6 | `fix(models)` | The models accepted booleans where JSON says integer and coerced them to 1. `isinstance(True, int)` is a Python fact and not a JSON one: `schema/trace-claim.json` rejects `{"slsa_level": true}` and the models accepted it, so the record was a claim of SLSA build level 1 assembled out of a boolean. |
| 7 | `fix(models)` | `TrustRecord.model_validate(record).model_dump()` returned a record `validate_json` rejects and whose signature no longer verifies, because pydantic writes every unset optional as an explicit `null` and the added members change the RFC 8785 canonical bytes. That round trip is what `sign_record`'s own docstring points a caller at. |
| 8 | `fix(validate)` | All eight `"format": "uri"` declarations in the schema were inert. `jsonschema` treats `format` as an annotation unless a checker for that format is installed, so `validate_json` built a `FormatChecker` and `"not a uri"` validated. |
| 9 | `fix(validate)` | The exported `SCHEMA` was the live object the validator reads. `_schema()` is `lru_cache`d, so mutating the dict handed to downstream tooling silently reconfigured `validate_json`, `iter_errors` and the structural gate inside `verify_record`, process wide, for every later call. |

## Two things worth knowing before merging

**One new runtime dependency, in commit 8.** `rfc3986-validator`, which is what makes
the eight `format: uri` declarations do anything. `jsonschema[format]` was not used
on purpose: it pulls `rfc3987`, which is GPLv3, into an Apache 2.0 dependency tree.

**Commits 6 and 8 tighten what validates.** Every JSON file in the repository was
walked for embedded records, 77 of them, and each was run through `validate_json` and
`TrustRecord.model_validate` on `main` and on this branch. There are zero acceptance
differences, so nothing in the tree is rejected that was accepted before. The control
for that check: on `main` measured in its own environment, `"not a uri"` validates and
`slsa_level: true` is accepted by the model; on this branch both are refused.

## Verification

- Every commit is load bearing. Reverting a commit's source change while keeping its
  tests turns the suite red: 20, 32, 10, 3, 12, 7, 10 and 3 failures respectively.
  Commit 8 changes a dependency declaration rather than code, so reverting the file
  proves nothing while the package is still installed. Uninstalling it is the real
  mutation: 25 failures.
- Suites run on 3.11 and 3.12, matching CI. `ruff check src tests scripts`,
  `python tools/check_dashes.py`, `mypy src/agentrust_trace` and
  `pytest --cov` are all clean; coverage is 96%.
- Every test file also passes alone, and the per file totals sum to the whole.
- Built as sdist and wheel, installed into an empty environment, and commit 9's
  property re checked from the installed artifact rather than the source tree.
- Merges into #233 with no code conflict. `sign.py` auto merges and the merged tree
  passes. Only the `CHANGELOG.md` entry position conflicts, which is a one line resolve.

## Type of change

None of the four boxes fit: this is implementation only. No spec text, no schema file
and no wire format changes. `pyproject.toml` gains the dependency named above.

## Spec section

None modified. The authority for each fix is named in its own CHANGELOG entry: commit 3
holds the revocation check to the fail closed rule the function's own docstring states,
and commit 6 restores agreement with `schema/trace-claim.json`, which already rejects a
boolean where it says integer.

## Checklist

- [x] DCO sign off on all commits
- [x] `CHANGELOG.md` updated
- [ ] Breaking changes marked in spec text: not applicable, no spec text changed
- [ ] Backward compatibility statement: not applicable, no breaking change. The
      acceptance comparison above is the evidence.


imran-siddique (approved):
Sweeping the package rather than reading one function is the right way to find this class of defect, and one commit per defect with an explicit offer to split the series is what makes 1318 lines reviewable.

**What I verified independently.** Three of the nine, and I would rather say which than imply I checked all of them:

1. **The exported `SCHEMA` was the live validator object.** `_schema()` is `lru_cache`d, so `SCHEMA = _schema()` handed downstream tooling the exact dict `_validator()` reads. Mutating it, which is the ordinary thing to do with a schema handed over to adapt, silently reconfigured `validate_json`, `iter_errors` and the structural gate inside `sign.verify_record`, process-wide, for every later call. The `copy.deepcopy` is right and the comment explaining why the copy is deliberate is what stops someone optimising it back out.

2. **All eight `format: uri` declarations were inert.** The validator was constructed with no `format_checker`, and JSON Schema treats `format` as an annotation unless one is registered. Eight declarations that read as constraints and enforced nothing.

3. **`bytes(5)` is `b'\x00\x00\x00\x00\x00'`**, confirmed. So `verify_assertion` hashed five zero bytes, got a mismatch, and reported that the record at the URL had changed. As you put it, that is a specific accusation about somebody else's server, and it was false. That is the worst of the nine and it is worth that it is stated that way in the commit rather than as "type check missing".

I did **not** independently verify defects 1, 3, 4, 5, 6 and 7. CI is green on all six checks and the ratio is roughly 269 lines of source against 1000 of tests, which is why I am comfortable merging the series rather than holding it for a line-by-line pass.

Two of them converge with decisions made elsewhere today, which is worth recording: the boolean-accepted-as-SLSA-level defect is the same Python fact that trace-tests#85 caught in the conformance suite, and the revocation check reading truthiness is the same fail-open shape as a widened `except` I flagged on cmcp#590. `isinstance(True, int)` has now produced three separate defects across three repositories in one day.

The quickstart one deserves a note too. A comment reading "keep secure, never commit or log" above code that writes an unencrypted Ed25519 private key into the working tree is enforcement by politeness, and covering it in `.gitignore` with a test asserting the docs do not leave a key behind is the correct fix rather than editing the comment.

Merging. #234 and #233 both touch files this changes, so both will need a rebase; I am telling each of them.

PR #252, opened 2026-08-30
fix(provenance): establish cnf object boundary

### LINKED ISSUE #251: provenance.verify_record still leaks AttributeError on malformed cnf
## Current behavior

The package-wide hardening in #236 and the earlier #225 fix now establish object boundaries for the top-level provenance record, `identity`, and `tool_catalog`, but the same pattern remains on `cnf` at current `main` (`25013c59d607860a4a4d2608476280286f688243`):

```python
```

A truthy non-object value therefore still reaches `.get()` and escapes as a plain `AttributeError` instead of the `ProvenanceError` that `provenance.verify_record()` documents for verification failure.

Examples that reach this edge after the preceding structural checks are values such as:

```json
"cnf": "not-an-object"
```

or

```json
"cnf": ["unexpected"]
```

Falsey malformed values (`""`, `0`, `[]`) are a separate asymmetry: `or {}` silently treats them as though `cnf` were absent rather than establishing the field's JSON type.

## Why this is distinct from #225 / #236

`_as_object()` already exists in `provenance.py` and is now used for `identity` and `tool_catalog`. `cnf` is the remaining nested object read through the old `(... or {}).get(...)` shape in the same verifier path.

This matters because callers are explicitly told they can treat `ProvenanceError` as the rejection boundary. An `AttributeError` from untrusted record structure violates that API contract and can terminate the caller instead of returning the expected verification failure.

## Smallest fix

Use the existing object-boundary helper rather than introducing another validator:

```python
cnf = _as_object(record.get("cnf"), "cnf")
embedded = cnf.get("jwk")
```

This preserves the existing missing/`None` behavior while refusing every malformed non-object consistently.

## Regression evidence

A focused matrix should include:

- missing `cnf` -> existing behavior preserved;
- `cnf: null` -> existing behavior preserved;
- `cnf: {}` -> existing behavior preserved;
- truthy non-objects (`"x"`, `[1]`, `1`, `true`) -> `ProvenanceError`, never `AttributeError`;
- falsey non-objects (`""`, `[]`, `0`, `false`) -> `ProvenanceError`, rather than being laundered into absence;
- valid `cnf.jwk` -> unchanged verification behavior.

No schema or normative specification change appears necessary; this is exception-safety and API-contract consistency in the reference verifier.

altrudev:
## What

Closes #251.

`provenance.verify_record()` now establishes `cnf` through the existing `_as_object()` boundary before reading `jwk`.

This keeps malformed externally supplied `cnf` values inside the verifier's documented `ProvenanceError` rejection boundary instead of:

- leaking `AttributeError` for truthy non-object values; or
- laundering falsey non-object values through `(... or {})` as though `cnf` were absent.

Missing `cnf`, explicit `null`, an empty object, and a valid embedded JWK keep their existing behavior.

## Scope

No schema or normative specification change. No new validator is introduced; this reuses the object-boundary helper already used for `identity` and `tool_catalog` in the same function.

## Regression evidence

The added matrix signs the actual malformed/edge records with the trusted key before verification, so a rejection cannot be attributed to an unrelated invalid signature.

Covered controls:

- truthy non-objects: string, list, integer, boolean;
- falsey non-objects: empty string, empty list, zero, false;
- missing / `null` / `{}` controls;
- valid embedded `cnf.jwk` control.

Mutation property: restoring the old `(record.get("cnf") or {}).get("jwk")` path makes the truthy vectors escape through host-language exceptions and the falsey vectors cease to be rejected.

## Non-claim

This is exception-safety and API-contract consistency in the MCP provenance verifier. It does not change provenance trust semantics or claim a cryptographic vulnerability.

AI-assistance disclosure: ChatGPT assisted with source triage, adversarial matrix design, and drafting. `altrudev` reviewed the bounded claim and remains responsible for the contribution.

### DIFF
--- a/src/agentrust_trace/provenance.py
+++ b/src/agentrust_trace/provenance.py
@@ -376,7 +376,8 @@ def verify_record(
     if not signature:
         raise ProvenanceError("record carries no signature")
 
-    embedded = (record.get("cnf") or {}).get("jwk")
+    cnf = _as_object(record.get("cnf"), "cnf")
+    embedded = cnf.get("jwk")
     if embedded:
         # Compared by RFC 7638 thumbprint, not by dict equality. A JWK is identified by
         # its key material; `kid`, `use` and `alg` are optional members that carry none
--- a/tests/test_provenance_cnf_boundary.py
+++ b/tests/test_provenance_cnf_boundary.py
@@ -1,0 +1,77 @@
+from __future__ import annotations
+
+import base64
+
+import pytest
+
+from agentrust_trace.provenance import ProvenanceError, build_record, verify_record
+from agentrust_trace.sign import _canonical_bytes, generate_key, key_to_jwk
+
+TOOLS = [
+    {
+        "name": "search",
+        "description": "search the docs",
+        "input_schema": {"type": "object"},
+    }
+]
+ARTIFACT = {
+    "package": "pkg:npm/%40acme/mcp-search@2.1.0",
+    "digest": "sha256:" + "a" * 64,
+}
+_MISSING = object()
+
+
+def _signed_with_cnf(cnf: object = _MISSING):
+    key = generate_key()
+    record = build_record(
+        kind="publisher-asserted",
+        publisher="did:web:acme.example",
+        tools=TOOLS,
+        artifact=ARTIFACT,
+    )
+    if cnf is not _MISSING:
+        record["cnf"] = cnf
+    body = _canonical_bytes(record)
+    signature = base64.urlsafe_b64encode(key.sign(body)).rstrip(b"=").decode()
+    return {**record, "signature": signature}, key_to_jwk(key)
+
+
+@pytest.mark.parametrize(
+    "bad_cnf",
+    [
+        "not-an-object",
+        ["unexpected"],
+        1,
+        True,
+        "",
+        [],
+        0,
+        False,
+    ],
+)
+def test_non_object_cnf_is_refused_through_provenance_error(bad_cnf) -> None:
+    record, trusted = _signed_with_cnf(bad_cnf)
+    with pytest.raises(ProvenanceError, match="cnf must be an object"):
+        verify_record(record, trusted)
+
+
+@pytest.mark.parametrize("cnf", [_MISSING, None, {}])
+def test_missing_null_and_empty_object_cnf_preserve_existing_behavior(cnf) -> None:
+    record, trusted = _signed_with_cnf(cnf)
+    verify_record(record, trusted)
+
+
+def test_valid_embedded_cnf_jwk_still_verifies() -> None:
+    key = generate_key()
+    trusted = key_to_jwk(key)
+    record = build_record(
+        kind="publisher-asserted",
+        publisher="did:web:acme.example",
+        tools=TOOLS,
+        artifact=ARTIFACT,
+    )
+    record["cnf"] = {"jwk": trusted}
+    body = _canonical_bytes(record)
+    record["signature"] = base64.urlsafe_b64encode(key.sign(body)).rstrip(b"=").decode()
+
+    verify_record(record, trusted)


lywinged (approved):
Reproduced on a clean checkout at `b37519e`. Four gates green locally: `ruff check src tests
scripts`, `tools/check_dashes.py`, `mypy src/agentrust_trace` (`Success: no issues found in 10
source files`), `pytest` at `1027 passed, 1 skipped`.

The mutation property holds as you state it. Restoring `(record.get("cnf") or {}).get("jwk")` and
changing nothing else: `8 failed, 1019 passed, 1 skipped`. Running the eight parameters against
`main` gives the split your body describes:

```
'not-an-object'  AttributeError: 'str' object has no attribute 'get'
['unexpected']   AttributeError: 'list' object has no attribute 'get'
1                AttributeError: 'int' object has no attribute 'get'
True             AttributeError: 'bool' object has no attribute 'get'
''               accepted, verify_record returns
[]               accepted, verify_record returns
0                accepted, verify_record returns
False            accepted, verify_record returns
```

On the falsey half `verify_record` returns normally, so those records verified rather than merely
skipping a check.

One addition to the evidence rather than a correction to it. The mutation you state is the old
path. There is a second mutant these tests kill, and it is the more plausible wrong fix:

```python
_c = record.get("cnf")
cnf = _as_object(_c, "cnf") if _c else {}
```

That reads as faithful, and it closes the `AttributeError` in the issue title. It passes the four
truthy parameters and fails exactly the four falsey ones, `4 failed, 8 passed` in the new file. So
the falsey half of the matrix is not only covering the second symptom, it is what stops a wrong
fix from looking right. That is the part of this PR I would least want trimmed in review.

Two checks for whoever else reads this.

`sign.verify_record` already refuses all eight, as `ValueError`, at the `validate_json`
schema step (`sign.py:518`):
`record does not conform to the TRACE v0.2 schema at cnf: '' is not of type 'object'`.
`provenance.verify_record` has no schema step, so an explicit guard is the only means it has. The
two verifiers now agree on this field, each by the mechanism available to it, and there is nothing
for this PR to do in `sign.py`.

`spec/server-provenance-v1.md` line 106 carries `cnf.jwk` as `required: yes`, so no record the
spec calls valid is newly refused. Sweeping what `verify_record` reads off the record once this
lands, `signature` is the only one still used as a type nothing has established. Separate from
this PR and not a reason to hold it.

Approving.

Tool-assisted: the matrix, the mutants and this write-up.

imran-siddique:

imran-siddique (approved):
Merging, first in the order you asked for.

Two lines, and the defect behind them is the kind that survives review indefinitely because the code reads as defensive:

```python
```

`or {}` looks like a guard and is doing two different wrong things. A truthy non-object leaks `AttributeError` straight past the verifier's documented `ProvenanceError` boundary, so a caller written to the contract does not catch it. And a **falsey** non-object, `0`, `""`, `[]`, is silently converted into "no `cnf` present at all". Absent and malformed are different states, and only one of them is something a verifier should tolerate quietly. The second is the half I would have missed reading this quickly, and it is the one with a real consequence: a record can carry a malformed `cnf` and be verified as though it carried none.

Routing through `_as_object()` rather than adding a validator is right. It is the boundary already applied to `identity` and `tool_catalog` in the same function, so this removes an inconsistency rather than introducing a rule.

**The regression matrix is built correctly**, which is worth saying because it is the part that usually is not: signing the malformed and edge records with the trusted key *before* verification means a rejection cannot be attributed to an unrelated signature failure. A test that rejects for the wrong reason passes and proves nothing.

Six checks, only the maintainer gate red, and independently approved by @lywinged. A peer review from someone who has been finding real defects in this same module all week is worth more here than a second maintainer pass.

Closes #251. This also moves `main`, so #258 can take the rebase you committed to and drop the stacked predecessor from its diff. #254 is next.

PR #257, opened 2026-08-30
fix(provenance): enforce tool_catalog.tool_count

### LINKED ISSUE #256: provenance verifier never enforces required tool_catalog.tool_count
## Current behavior

`spec/server-provenance-v1.md` marks `tool_catalog.tool_count` as required and defines it as the number of tools described by the catalog. The field is intentionally redundant with the hash so a consumer can distinguish a plausible-count/hash mismatch from a completely different catalog.

At current `main` (`25013c59d607860a4a4d2608476280286f688243`):

- `build_record()` emits `tool_count = len(tools)`;
- `verify_record()` validates `tool_catalog.hash` but never validates `tool_count`;
- `check_tool_catalog()` reads `tool_count` only after the hash has already mismatched, to include it in the diagnostic.

A correctly signed record can therefore carry the correct catalog hash for two tools and `tool_count: 999`; `verify_record()` succeeds and `check_tool_catalog()` also succeeds because the hash matches.

The same is true for missing, negative, boolean, string, or otherwise malformed `tool_count` values: the required field is not load-bearing on the consumer path.

## Why this is distinct from #142

Issue #142 noted in passing that `tool_count` was never checked, but its merged fix #146 addressed the four structural defects that allowed `tee-attested` records without evidence and malformed identity/timestamp state. `tool_count` remains unchanged on current `main`.

This issue isolates that remaining field rather than reopening #142.

## Smallest fix

There are two complementary checks:

1. `verify_record()` should require `tool_count` to be a non-negative JSON integer (excluding bool), because the spec marks the member required.
2. `check_tool_catalog()` should compare the declared count with `len(tools)` after the hash matches, so a signed internally inconsistent record cannot pass the live-server check.

For example:

```python
declared_count = catalog.get("tool_count")
if (
    not isinstance(declared_count, int)
    or isinstance(declared_count, bool)
    or declared_count < 0
):
    raise ProvenanceError("tool_catalog.tool_count must be a non-negative integer")
```

If the live hash mismatches, the existing `ToolCatalogMismatch` remains the right result: the paper may be fine and the server differs. If the live hash matches but `len(tools) != declared_count`, the record contradicts its own signed catalog metadata, so that should remain an ordinary `ProvenanceError`, not `ToolCatalogMismatch`.

## Regression evidence to add

- correct hash + correct count -> unchanged success;
- correct hash + wrong positive count -> `verify_record()` can establish only structural validity, but `check_tool_catalog()` refuses the internally inconsistent record;
- missing count -> `ProvenanceError`;
- negative count -> `ProvenanceError`;
- boolean count -> `ProvenanceError`;
- string count -> `ProvenanceError`;
- ordinary catalog hash mismatch -> existing `ToolCatalogMismatch` behavior preserved.

No wire-format or normative specification change is needed; this makes the reference consumer enforce an already required signed field while preserving the existing bad-document versus wrong-server error distinction.

altrudev:
## Current restack

Restacked onto current `main` at `98aae176aa9ae4053d944624dad7629c76789ff7`.

Exact head: `251c8b6efe1cd722d166ed734c83c47a9a77d583`.

The branch is one commit ahead / zero behind and changes exactly the same two files. Fresh exact-head CI and CodeQL both pass. The previous `@lywinged` approval was on pre-restack head `7b4ee0ba103ec369fb3d1bd8056b29964190f6e0` and was correctly dismissed by GitHub after the rewrite; no exact-head approval is claimed.

## What

Closes #256.

The server-provenance format marks `tool_catalog.tool_count` as required, but the reference consumer never made it load-bearing. A correctly signed record could carry the right catalog hash for two tools and `tool_count: 999`; both `verify_record()` and `check_tool_catalog()` still succeeded.

This change adds one shared `_tool_count()` boundary and uses it from both consumer entry points:

- `verify_record()` now requires a non-negative JSON integer and excludes `bool` explicitly;
- `check_tool_catalog()` validates the same field even when called independently;
- after the live hash matches, `check_tool_catalog()` checks that the signed count equals the number of offered tools.

## Error semantics

The existing distinction is preserved.

- If the live catalog hash differs, `ToolCatalogMismatch` still means the signed paper may be fine and the server in front of the verifier is different.
- If the hash matches but the signed count disagrees with the matching live list, the record contradicts its own catalog metadata. That is a plain `ProvenanceError`, not `ToolCatalogMismatch`.

## Regression coverage

Focused tests cover:

- missing count;
- negative count;
- boolean count;
- string count;
- `null` count;
- floating-point count, pinning the JSON-integer boundary;
- independent `check_tool_catalog()` validation of malformed counts;
- hash mismatch outranking malformed-count validation;
- correct hash with wrong positive count;
- correct hash and count control;
- ordinary live hash mismatch preserving `ToolCatalogMismatch`.

The wrong-positive-count test deliberately verifies the paper first, then proves the live catalog check is the step that makes the semantic count disagreement load-bearing.

## Scope

No wire-format or schema change is made here. This PR does change the reference consumer's live-catalog semantics by making the signed count load-bearing after a matching hash. The current `spec/server-provenance-v1.md` §5 steps do not explicitly require that live-count comparison, so whether the normative text should be aligned is left to the maintainers/spec process rather than claimed as settled by this implementation PR.

AI-assistance disclosure: ChatGPT assisted with source triage, adversarial-case design, implementation drafting, restack preparation, and exact-diff review. `altrudev` reviewed the bounded claim and remains responsible for the contribution.

### DIFF
--- a/src/agentrust_trace/provenance.py
+++ b/src/agentrust_trace/provenance.py
@@ -66,6 +66,14 @@ def _as_object(value: Any, field: str) -> dict[str, Any]:
     return value
 
 
+def _tool_count(catalog: dict[str, Any]) -> int:
+    """Return the required catalog count as a JSON integer."""
+    value = catalog.get("tool_count")
+    if not isinstance(value, int) or isinstance(value, bool) or value < 0:
+        raise ProvenanceError("tool_catalog.tool_count must be a non-negative integer")
+    return value
+
+
 class ToolCatalogMismatch(ProvenanceError):
     """The server offered a tool set the record does not describe.
 
@@ -344,6 +352,7 @@ def verify_record(
     catalog = _as_object(record.get("tool_catalog"), "tool_catalog")
     if not _DIGEST_RE.match(str(catalog.get("hash", ""))):
         raise ProvenanceError("tool_catalog.hash is not a sha256: digest")
+    _tool_count(catalog)
 
     # Freshness. `issued_at` has been required and type-checked since the format
     # existed, with an error message explaining that a record with no issue time
@@ -442,4 +451,10 @@ def check_tool_catalog(record: dict[str, Any], tools: list[dict[str, Any]]) -> N
             f"({len(tools)} tools offered, record declares {declared_count}). "
             "The signature may be perfectly valid; this is about the server, not the "
             "document."
+        )
+    declared_count = _tool_count(catalog)
+    if declared_count != len(tools):
+        raise ProvenanceError(
+            f"tool_catalog.tool_count declares {declared_count} tools, but the matching "
+            f"catalog contains {len(tools)}"
         )
--- a/tests/test_provenance_tool_count.py
+++ b/tests/test_provenance_tool_count.py
@@ -1,0 +1,90 @@
+from __future__ import annotations
+
+import pytest
+
+from agentrust_trace.provenance import (
+    ProvenanceError,
+    ToolCatalogMismatch,
+    build_record,
+    check_tool_catalog,
+    sign_record,
+    verify_record,
+)
+from agentrust_trace.sign import generate_key, key_to_jwk
+
+TOOLS = [
+    {
+        "name": "search",
+        "description": "search the docs",
+        "input_schema": {"type": "object"},
+    },
+    {
+        "name": "fetch",
+        "description": "fetch a page",
+        "input_schema": {"type": "object"},
+    },
+]
+ARTIFACT = {
+    "package": "pkg:npm/%40acme/mcp-search@2.1.0",
+    "digest": "sha256:" + "a" * 64,
+}
+_MISSING = object()
+
+
+def _signed_with_count(count: object = _MISSING):
+    key = generate_key()
+    record = build_record(
+        kind="publisher-asserted",
+        publisher="did:web:acme.example",
+        tools=TOOLS,
+        artifact=ARTIFACT,
+    )
+    if count is _MISSING:
+        del record["tool_catalog"]["tool_count"]
+    else:
+        record["tool_catalog"]["tool_count"] = count
+    return sign_record(record, key), key_to_jwk(key)
+
+
+@pytest.mark.parametrize("bad_count", [_MISSING, -1, True, "2", None])
+def test_verify_record_refuses_malformed_or_missing_tool_count(bad_count) -> None:
+    record, trusted = _signed_with_count(bad_count)
+    with pytest.raises(ProvenanceError, match="tool_catalog.tool_count must be"):
+        verify_record(record, trusted)
+
+
+@pytest.mark.parametrize("bad_count", [2.0, "2", True, -1, None])
+def test_check_tool_catalog_alone_refuses_malformed_count(bad_count) -> None:
+    record, _ = _signed_with_count(bad_count)
+    with pytest.raises(ProvenanceError, match="tool_catalog.tool_count must be"):
+        check_tool_catalog(record, TOOLS)
+
+
+def test_wrong_positive_count_is_detected_when_live_catalog_is_checked() -> None:
+    record, trusted = _signed_with_count(999)
+
+    verify_record(record, trusted)
+    with pytest.raises(ProvenanceError, match="declares 999 tools") as excinfo:
+        check_tool_catalog(record, TOOLS)
+    assert not isinstance(excinfo.value, ToolCatalogMismatch)
+
+
+def test_correct_count_and_hash_still_verify() -> None:
+    record, trusted = _signed_with_count(len(TOOLS))
+    verify_record(record, trusted)
+    check_tool_catalog(record, TOOLS)
+
+
+def test_hash_mismatch_remains_a_tool_catalog_mismatch() -> None:
+    record, trusted = _signed_with_count(len(TOOLS))
+    verify_record(record, trusted)
+
+    offered = [TOOLS[0]]
+    with pytest.raises(ToolCatalogMismatch, match="about the server, not the document"):
+        check_tool_catalog(record, offered)
+
+
+def test_hash_mismatch_outranks_a_malformed_count() -> None:
+    record, _ = _signed_with_count("2")
+    with pytest.raises(ToolCatalogMismatch):
+        check_tool_catalog(record, [TOOLS[0]])


lywinged (commented):
Reproduced at `7cda000`. Four gates green: `ruff check src tests scripts`,
`tools/check_dashes.py`, `mypy src/agentrust_trace` (`Success: no issues found in 10 source
files`), `pytest` at `1023 passed, 1 skipped`.

The reorder in `7cda000` is right, and it is cheaper than it looks. I ran both orderings over
twenty `tool_count` values crossed with three offer sets and diffed the sixty cells. Thirty
differ, and they are exactly the fifteen malformed counts crossed with the two offer sets whose
hash does not match. All thirty are refusals under both orderings; only the class moves,
`ToolCatalogMismatch` after `7cda000` against `ProvenanceError` before it. Nothing in the grid
moves between accepted and refused, and on every well-formed count, including a wrong one like
`999`, the two orderings are identical down to the message text. Your Error semantics section does
what it says.

Three coverage gaps. All three came from running a fix a reviewer would plausibly have written
instead of running the revert.

**The reorder your second commit makes is not pinned by the new file.** Reverting `7cda000` and
changing nothing else leaves `tests/test_provenance_tool_count.py` entirely green, `8 passed`. The
whole suite catches it with exactly one failure,
`test_check_tool_catalog_null_tool_catalog_still_behaves_like_absent` in `tests/test_provenance.py`,
which reaches the ordering only because a `null` catalog makes the count absent as a side effect.
`test_hash_mismatch_remains_a_tool_catalog_mismatch` builds with `_signed_with_count(len(TOOLS))`,
so `_tool_count` returns cleanly and it passes under either ordering.

**The second bullet in your What section is not pinned by anything.** Substituting the raw read for
the guard in `check_tool_catalog`:

```python
declared_count = catalog.get("tool_count")
```

leaves the whole suite green at `1023 passed, 1 skipped`. Under it, `tool_count: 2.0` against a
matching two-tool catalog is accepted in silence, because `2.0 != 2` is false in Python. The guard
is in the code; nothing tests that `check_tool_catalog` applies it when called on its own, which is
the property that bullet claims.

**The `int` half of `_tool_count` is not pinned against a float.** Widening the check to

```python
if isinstance(value, bool) or not isinstance(value, (int, float)) or value < 0:
```

also leaves the whole suite green at `1023 passed, 1 skipped`. Under it, `tool_count: 2.0` passes
`verify_record` and then passes `check_tool_catalog` against a two-tool catalog, again because
`2.0 != 2` is false. JSON has one number type, so a producer emitting `2.0` is not exotic, and
`7cda000` as it stands refuses it correctly. Your `bad_count` list is `[_MISSING, -1, True, "2",
None]`, which pins `bool`, `str`, `None` and the negative, but nothing pins `int` against `float`.

Two tests against your existing helper close all three:

```python
@pytest.mark.parametrize("bad_count", [2.0, "2", True, -1, None])
def test_check_tool_catalog_alone_refuses_malformed_count(bad_count) -> None:
    record, _ = _signed_with_count(bad_count)
    with pytest.raises(ProvenanceError, match="tool_catalog.tool_count must be"):
        check_tool_catalog(record, TOOLS)


def test_hash_mismatch_outranks_a_malformed_count() -> None:
    record, _ = _signed_with_count("2")
    with pytest.raises(ToolCatalogMismatch):
        check_tool_catalog(record, [TOOLS[0]])
```

All six pass on `7cda000` as it stands. Against the raw-read variant the parametrized one fails
five of five; against the revert the second one is the single failure; against the float-widened
check the `2.0` parameter is the single failure. The `2.0` entry is carrying two of those on its
own, so it is worth keeping even though it looks like the odd one in the list. Appended to your file as
written, they pass ruff, `check_dashes` and mypy, and take the suite to `1029 passed, 1 skipped`.

One for a maintainer rather than for you. `check_tool_catalog` now refuses a record whose signed
count disagrees with a matching live catalog. `spec/server-provenance-v1.md` §5 has six steps, none
of which compares a declared count against the live catalog, and §7 defines consumer conformance as
steps 1, 2, 5 and 6. So the reference consumer refuses records a §5-conformant consumer accepts.
#256 argues for that and it may well be right, but "no normative specification change" does not
quite reach it, and it is not mine to settle.

Add the two tests and I am glad to approve.

Tool-assisted: the grid, the mutants and this write-up.

altrudev:
Thanks — added the two mutation-pinning tests you suggested at `7b4ee0b`:

- standalone `check_tool_catalog()` rejects malformed counts, including `2.0`;
- live hash mismatch outranks malformed-count validation.

I also tightened the Scope section so it no longer claims the live-count comparison is already normatively settled by §5; the implementation/spec alignment is left explicitly to the maintainers/spec process.

The fork-triggered CI/CodeQL runs for the new head are currently waiting on workflow approval.

imran-siddique:

lywinged (dismissed):
Verified at `7b4ee0b`. Both tests landed as proposed and now hold what the review found: the
raw-read substitution in `check_tool_catalog` fails five of five parameters, the reorder revert
fails inside this file rather than by accident elsewhere, and the float widening dies on the `2.0`
parameter. Four gates green, `1029 passed, 1 skipped`. The Scope wording now matches what the
change does. Approving.

Tool-assisted: the re-run and this note.

altrudev:
@imran-siddique #257 is the only lane I am advancing now.

It is cleanly restacked onto current `main` (`98aae176aa9ae4053d944624dad7629c76789ff7`) at exact head `251c8b6efe1cd722d166ed734c83c47a9a77d583`: one commit ahead / zero behind, exactly two files, with the previously reviewed semantic diff preserved over the current provenance code including #252/#258. Fresh exact-head CI and CodeQL both pass.

@lywinged's prior approval was on pre-restack head `7b4ee0ba103ec369fb3d1bd8056b29964190f6e0` and GitHub correctly dismissed it after the rewrite, so I am not claiming exact-head approval from that review.

The remaining maintainer question is the one Louie identified: whether the live `tool_count` comparison belongs in the reference consumer given the current §5/§7 wording. The PR body now states that scope explicitly rather than treating it as normatively settled.

lywinged (approved):
Approved, on `251c8b6e`.

The restack checks out in the stronger form. Against `98aae176` the branch is one ahead and none behind, and it touches exactly the two files it touched before. `tests/test_provenance_tool_count.py` is byte-identical to the previous head, `7b4ee0ba`. `provenance.py` is not, but that is the base moving under it: with the context lines stripped, the pull request's own patch on the old base and on the new base is the same fifteen lines. The blob differs only because #258 landed in the same file, between the hunks.

The ordering is the one from my earlier comment. `_tool_count` runs after the hash comparison, so a record with `tool_catalog: null` still reports `ToolCatalogMismatch` rather than a type error, and the older control test that pins that still passes.

What I ran on this head: the suite is 1167 passed 1 skipped, and the new test file is 14 passed. Then three mutations, since a guard is only worth what fails when it is gone:

- remove `_tool_count(catalog)` from `verify_record`: five tests fail, one per malformed shape in the parametrize list;
- remove only the `declared_count != len(tools)` raise from `check_tool_catalog`, keeping the assignment: exactly one fails, `test_wrong_positive_count_is_detected_when_live_catalog_is_checked`, with `DID NOT RAISE`;
- drop the `isinstance(value, bool)` exclusion, the simplification a reviewer is most likely to propose since `bool` already passes `isinstance(value, int)`: two fail, both the `True` case, one per entry point.

The first and third share the `verify_record` `True` case, which is expected: removing the call and weakening the check both let `True` through. What matters is that each of the three has at least one test only it turns red: four for the first, the wrong-positive-count test for the second, and `test_check_tool_catalog_alone_refuses_malformed_count[True]` for the third. None of the three checks is covered by a neighbour.

One thing for the maintainers rather than for this review, since it is the point the body now leaves open. Section 3 calls `tool_count` redundant with the hash by design; section 5 step 5 compares the hash against the tools the server actually offers, and section 7 defines the hash over those tools. Neither mentions the count. So the comparison this PR makes after a hash match, count against the offered list, is behaviour this implementation adds, neither required nor ruled out by the format. Whether the reference consumer should carry it is a spec question, and approving the code does not settle it either way. The body says so now, which is the right place for it.

On the dismissed approval: correct not to carry it across a rewrite, and the checks above are why this one stands on the new head.

imran-siddique (approved):
Read the whole diff at `251c8b6`. This is right.

`tool_count` was required by the format and read by nothing, so a correctly signed record could carry the right catalog hash for two tools and declare 999, and both entry points passed it. Making the field load-bearing is the fix.

Three details I checked specifically, because they are the ones that usually go wrong:

- `_tool_count` rejects `bool` explicitly. `isinstance(True, int)` is true in Python, so without that clause `tool_count: true` would have counted as 1.
- `2.0` is rejected as well, since `isinstance(2.0, int)` is false. A JSON producer emitting a float where an integer is required is exactly the case a count check should catch.
- In `check_tool_catalog` the count comparison runs after the hash check, so a live hash mismatch outranks a malformed count. That ordering is not incidental and you pinned it with its own test. It is the same principle I ruled on today in cmcp#596 and trace-spec#242: establish the thing before interpreting it, and never report a downstream verdict about a structure you have not confirmed.

Thanks for calling the dismissed approval yourself rather than leaving it to look like it stood. A restack dismissing an approval is silent, and most people do not mention it.

Approving to release the hold. Merging.

PR #259, opened 2026-08-30
docs(spec): state what the references block cannot carry, and distinguish resolver from a verification authority

FaresMallouli:
## What

Two paragraphs in §3.1.2. The first states the two consequences of the block being assurance-neutral: an entry cannot carry compliance evidence, and cannot carry a pre-execution commitment. The second separates `resolver` as a retention undertaking from a verification authority.

## Why

§3.1.2 says what a reference *is* — a pointer, not evidence — but not what follows from that. Both consequences are properties of the block rather than of its relation set, so registering a new `rel` does not touch either. Written as a general boundary rather than against any one use, so the next person does not re-derive them.

The `resolver` paragraph exists because the field reads as a verification authority until the two are separated. `docs/modules/tr-pol.md` refuses the opposite arrangement for a different field: TR-POL-003 takes its resolver from the caller and never derives it from the record, because a record that names its own checker can name one that agrees with it. Naming yourself as the party who must retain an artifact is ordinary; naming yourself as the party who decides whether it is true is that circularity. Rule 4 is aimed at a producer who can name no obliged party at all, not at one who is that party.

Both came out of the discussion in agentrust-io/agent-manifest#343.

## Spec impact

§3.1.2, non-normative. No RFC 2119 keywords added, no schema change, no model change, no new field. The four numbered rules and the registered `rel` set are unchanged. CHANGELOG entry added under `## [Unreleased] / ### Added`.

## Test plan

Documentation-only change to spec prose; no code paths touched.

## DCO

All commits in this PR are signed off (`git commit -s`).

Refs agentrust-io/agent-manifest#343


lywinged (commented):
Checked at `fdf7433`. Docs only, one signed-off commit, CHANGELOG entry under
`## [Unreleased] / ### Added`, and the four gates pass on a local checkout at that head: ruff,
`check_dashes`, which does cover `spec/trace-v0.2.md`, mypy, and `1015 passed, 1 skipped`.

The first paragraph I have no argument with. It draws two consequences out of what §3.1.2 already
says without adding a rule to get there, and pinning them to the block rather than to the relation
set is what makes them survive a new `rel`.

I checked the TR-POL-003 claim against the suite rather than taking it on trust, and it holds.
`docs/modules/tr-pol.md` in `trace-tests` has the resolver "supplied by the caller and never
derived from the record", because "a record that named its own resolver could name one that agrees
with it".

Three things on the second paragraph.

**The repository says four times that this question is not decidable, and the paragraph decides
it.** Both schema copies, `schema/trace-claim.json` and
`src/agentrust_trace/schema/trace-v0.2.json`, carry it as the `resolver` description:

> Which identifiers are self-asserted is not decidable from the record, so this constrains the
> field's presence and not its value.

`src/agentrust_trace/models.py` and the `references` CHANGELOG entry say it in their own words,
both ending on "whether an identifier is self-asserted is not decidable from the record". That is
why `resolver` is `minLength: 1` and nothing more. The new paragraph says an operator naming
themselves is "within this section" and that rule 4 "is aimed at a producer who can name no obliged
party at all, not at one who is that party", where rule 4's own words are "an empty or
self-asserted resolver" and it is a MUST. I think the reading is probably right and the argument
for it is good. It is still a position none of those four take, in the same commit whose CHANGELOG
says "No normative text, schema, or record field changed". One place does lean your way:
`tests/test_references_block.py` has the comment "An empty resolver is the self-asserted entry rule
4 tells a producer to omit", treating the two halves of rule 4 as one. A test comment does not
settle it, but it shows the ambiguity is already being read both ways in the tree.

**"refused" is the wrong verb for TR-POL-003, in both sentences that use it.** That test runs on
`policy.policy_uri` against `policy.bundle_hash`, and it does not refuse a record for naming its
own resolver. It never reads the record's resolver at all, and the caller supplies one instead,
which is a different thing. I take "would still be refused under TR-POL-003's rule" as an analogy
rather than a claim that it processes a `references` entry, and as an analogy it is a good one, but
the verb is what a reader takes away. "Declines to derive it from the record" is accurate and is
the sharper phrase for the point.

**This would be the first conformance test ID cited in any specification document.** `TR-` appears
zero times across all five files in `spec/` at `main`. The IDs live in `examples/README.md` and
`CHANGELOG.md`, and the spec refers to the suite generically, as §3.2.2 does: "The TRACE
conformance suite (trace-tests) already enforces both rules". `trace-tests` is separately
versioned, 0.4.1 against this spec at 0.2 Draft, and nothing in this repository would notice a
renumbering there. Your PR body names `docs/modules/tr-pol.md` and the spec text does not, so the
reader who most needs the pointer is the one without it. Keeping the ID in the CHANGELOG and
generalising the spec sentence would preserve the whole argument.

The second paragraph is worth having. I am not approving it as written because of the rule 4 point,
and either of two things settles it. Say in the paragraph that it is reading rule 4 rather than
restating it, which costs a clause and leaves the four artifacts true. Or take the reading
properly, in which case rule 4's wording moves and the schema descriptions move with it, which is a
larger change than this PR is scoped for. The verb and the test ID are small next to that.

Tool-assisted: the greps, the suite cross-check and this write-up.

imran-siddique (approved):
Both halves of the #343 answer, written into §3.1.2 where they generalise rather than into §9.1.2 where they would have read as Article 50 trivia. That was the right choice and it is the one I would have had to ask for.

**What the block cannot carry.** The block's assurance-neutrality already told a reader what a reference *is*. It did not say what follows, and the two consequences are the ones a contributor will otherwise re-derive: an entry cannot carry compliance evidence, because the block commits to the pointer and not to the target, so no registered `rel` turns an entry into evidence that an obligation was met. Making it explicit that this is a property of the block rather than of its relation set is the part that prevents the next proposal arriving as "register a new `rel` for X".

**`resolver` names who must retain, not who adjudicates.** This is the distinction you asked me to rule on and it survived into the text accurately, including the contrast that makes it stick: TR-POL-003 takes its resolver from the caller and never derives it from the record, because a record naming its own checker can name one that agrees with it, whereas naming yourself as the party obliged to retain something is ordinary and non-circular. You were the first reader to hold those two against each other and you will not be the last.

Both statements describe consequences of text that already exists rather than imposing new obligations, which is why this needed no organisational sponsor.

Issue #246, opened 2026-08-29
verify_record reports a pass when no revocation check was performed, where §3.2.3 requires it to report that none was performed


imran-siddique:
§3.2.3 requires a verifier to report absence of revocation evidence as absence. `verify_record()` promotes it to a pass. The same question gets two answers in one specification, and the code path is the permissive one.

Raised from the APS integration side in [agentrust-io/integrations#140](https://github.com/agentrust-io/integrations/issues/140) by @aeoess, who noticed it as a divergence between APS and TRACE. It is not: it is a divergence between TRACE and TRACE.

## What the spec says

`docs/verification.md`, on the bundle path, quoting §3.2.3:

> a verifier with no bundle *"MUST report that it performed no revocation check"*

and an expired bundle *"MUST report the record as unverified for revocation rather than as verified"*.

That is the right rule and the reasoning is stated well: offline is a state you report, not a check you skip.

## What the code does

At `25013c59d607860a4a4d2608476280286f688243`, `src/agentrust_trace/sign.py`:

```python
if revocation is not None:
    _check_not_revoked(trusted_jwk, revocation)
```

- **No store passed at all** (`revocation=None`): the check is skipped and the record verifies. Nothing in the result records that no revocation check was performed.
- **An empty store passed**: `identifier in revocation` is `False`, so the key reads as not revoked, and the record verifies. An empty container and a container that has affirmatively been checked are indistinguishable in the result.

Neither outcome reports "no revocation check performed". Both report a pass.

## Why this matters more than it looks

The store path is not a lesser surface. `docs/verification.md` states it is §3.2.3's own fallback for records with no usable inclusion entry ID:

> That fallback is what the current `verify_record()` store implements, and it is the correct behaviour for deployments carrying no receipts.

So the deployments most likely to pass no store, or an empty one, are precisely the ones the fallback exists for. The bundle path states the absence rule and the fallback path silently ignores it.

The failure is quiet by construction. A relying party wires up revocation, misconfigures the store so it is empty, and every record verifies exactly as it would with a correctly populated one. There is no observable difference between "checked against a real list" and "checked against nothing".

## What is not wrong

The fail-closed behaviour that exists is good, and #236 improved it today: a store that raises is a rejection, and as of #236 a callable returning a non-bool is too, because truthiness is unrelated to revocation status. Both are the right direction. This issue is about the third case, which is not an error at all: the store answers, honestly, that it knows nothing.

## Suggested direction, not a decision

[...]

Issue #114, opened 2026-08-01
Widen the per-call decision enum under a new profile version, not in 1.0


imran-siddique:
Decision recorded so this is not re-derived: **widen under a new profile version, leave `1.0` untouched.**

## Background

`agentrust-io/cmcp` now records the five AARM R4 decision types (`ALLOW`, `DENY`, `MODIFY`, `STEP_UP`, `DEFER`) in its audit chain. The TRACE Claim cannot carry two of them. `schemas/trace-claim.schema.json` in cmcp pins the per-call decision enum to:

```
["allow", "deny", "redact", "advisory_deny", "fault", "n/a"]
```

So `step_up` and `defer` are currently **narrowed to `deny`** at the claim boundary, in [`cmcp` PR #442](https://github.com/agentrust-io/cmcp/pull/442). That is sound and tested: the call was blocked in every case, which is the part a claim asserts, and unrecognised values also narrow to `deny` so a future value cannot emit a claim that fails validation. The cost is that the specific decision is invisible to anyone who only reads claims.

## Why not widen 1.0 in place

A verifier pinned to the current `1.0` enum would reject a claim carrying `step_up`. Compatibility is the entire reason a version is pinned, so widening in place spends it. Rejected for that reason.

## What to do instead

Add the values under a new profile version, so adopters opt in and existing verifiers keep working.

Coordination, which is why this is an issue rather than a patch:

| Repo | Change |
|---|---|
| `trace-spec` | New profile URI, `tag:agentrust-io.com,2026:trace-v0.3` or a `1.1` claim version. Document the widened enum and that `1.0` is unchanged |
| `trace-tests` | Conformance cases for the new values; existing `1.0` cases must keep passing unmodified |
| `trace-registry` | Accept and index the new profile |
| `cmcp` | Emit the new profile, and drop `claim_value()`'s narrowing when the profile in use supports the values. Keep narrowing for `1.0` |

Semantics worth settling in the spec rather than per implementation:

- **`STEP_UP`** means the call was blocked and a named human authority can authorize it. A relying party must not read it as allowed.
- **`DEFER`** in cmcp today is classified but **not** asynchronously enforced: there is no callback registry, so a `defer` entry is a blocked call, not a pending one. If the spec permits `defer`, it should say whether it may ever mean "pending", because a verifier that assumes pending when the implementation means blocked would be wrong in the permissive direction. See cmcp's `LIMITATIONS.md`.

That second point is the one I would not leave to implementations to decide.

[...]

PR #260, opened 2026-08-31
spec: disclosed gaps in a receipt chain, and the receipt_gap_disclosed outcome (#117)

### LINKED ISSUE #117: spec: disclosed receipt gaps — GapDisclosure claim + receipt_gap_disclosed verifier outcome

lywinged:
Closes #117. 

Opened per the maintainer-carry note there; all three asks from that comment are in,
each named below. Authorship per CONTRIBUTING step 3: carried by a maintainer, proposer credited
in the CHANGELOG entry.

## What

Section 3.3.4, normative, marked `<!-- CHANGED: #117 -->`: a `GapDisclosure` is a signed chain
element stating that receipts which would have occupied its position were never emitted. Coverage
is structural rather than asserted: the disclosure links back to the element before the gap, the
next element emitted links back to the disclosure, and verification is two link checks a verifier
already performs on every ordinary element. There are no range fields. A hash chain cannot express
a range, and an emitter cannot know its successor's hash at write time; the design discussion in
#117 walks through why the issue's original field list does not survive that.

The action-receipt outcome `receipt_missing_required` is narrowed to silent absence, and
`receipt_gap_disclosed` is added beside it. Acceptance is a verifier policy input, with one
non-negotiable bound: it never satisfies a profile requiring independently proven completeness.

## The tail case, as a test

A disclosure at the live tail, after the failure and before resumption, has no successor to seal
it. The spec text forbids granting it `receipt_gap_disclosed` and forbids `receipt_invalid`
equally: inability to check is not evidence of a defect, section 3.3.2's principle for unknown
issuer keys. It is surfaced unverified with a distinct advisory and upgraded on re-verification
once the chain resumes.

`test_the_live_tail_is_not_granted_what_the_seal_earns` pins it: fixtures 01 and 17 carry
byte-identical disclosures whose only difference is whether a successor exists, and the test
asserts the sealed one earns the outcome and the tail one does not. The adversarial reading is in
the test's docstring: a chain truncated immediately after a disclosure is byte-for-byte an honest
live tail, so this is also the assertion that truncation buys an advisory, never an outcome.
Fixtures 16 and 18 carry the two near misses: a contradiction standing at the tail still fails,
and a stale successor-link field with no successor is still a tail.

## What a disclosed gap does not establish

That the missing receipts ever existed, how many were lost, or that the emitter did not omit them
selectively. What the splice proves is where the gap sits in the chain, and nothing else. The
section states this in its own text, and `cause` and `receipts_lost_estimate` are constrained to
descriptive self-reports a verifier MUST NOT condition any outcome on. The failure mode this
guards is a reader treating disclosure as though it restored what was lost; the text is written so
that no requirement can be read as making a disclosed gap benign.

## The interaction with the undecidable state

`receipt_gap_disclosed` and the tail's unverified state are two more instances of the pattern
being unified across #246, #190, trace-tests#82 and agent-manifest#348: "this was not established"
as a first-class outcome rather than a silent pass. Both names this PR introduces,
`receipt_gap_disclosed` and the tail's `gap_disclosure_unverified`, are candidates for that
vocabulary. Each is held in one constant in the generator and one in the test; outside those, the
disclosed name appears in one spec section and one docs table row. A rename is those sites and one
regeneration. Not blocked on it, per the carry note.

## Evidence

- Twenty conformance vectors under `examples/action-receipts/gap-disclosure/`, two independent
  vectors per rule, enforced by a margin test rather than remembered. The second vector of each
  pair sits against an implementation shortcut the first cannot detect: case-normalised key and
  session lookups, digest comparisons truncated to a prefix, sealed-ness derived from a link field
  rather than from the chain, a contradiction check gated on the sealed path, and a type check
  that only tests presence.
- The verifier consumes a named-rule registry (`RULES`) in the shape #124 settled next door, so an
  unregistered check is a check that never runs.
- The generator reproduces every fixture byte for byte, enforced by the existing
  `tests/test_generators_reproduce_fixtures.py` discovery.
- Fourteen hand-run mutants against the verifier: thirteen killed, each by the fixture built for
  it; the one survivor widens an exception catch that no JSON-expressible fixture can distinguish,
  recorded here rather than left to be found.
- Two known gaps from the issue thread are closed rather than carried: the `type` equality check
  the draft recorded as unimplemented now has a rule and two vectors (19, 20), and the Reporting
  obligations are echoed in every fixture's expected result rather than left unpinned.

## Housekeeping

Two cross-references that predate the renumbering which introduced section 3.3.1 are updated to
name section 3.3.2, where the text they cite now lives: one in section 3.3.3's profile paragraph,
one in the `receipt_unverified` row of `docs/verification.md`.

## Backward compatibility

Non-breaking. The section binds under a profile requiring action receipts; no current profile
does, and section 3.3.3 already contemplates such profiles as future work. Existing verifiers and
records are untouched; the outcome addition follows #114's decision shape, widening under opt-in
rather than in place. No schema change in this PR: per CONTRIBUTING, the schema PR tracks the
merged spec change.

## Checklist

- [x] Issue opened with the Spec change proposal template (#117, 2 August), comment window
  satisfied
- [x] Carried per CONTRIBUTING step 3; proposer credited in the CHANGELOG entry
- [x] Changed normative text marked `<!-- CHANGED: #117 -->`
- [x] `CHANGELOG.md` updated

Production origin, for the record: the mechanism is operated in a per-action receipt emitter whose
crash recovery writes a functionally equivalent disclosure; relying parties treat a disclosed gap
differently from a silent one, which is the distinction this section makes portable.

Tool-assisted: drafting, the mutant battery and the fixture regeneration. The derivation is in
the section text, the commit message and the test docstrings, where the next reader can find it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,14 @@ Format: [Semantic Versioning](https://semver.org/). Spec versions follow `MAJOR.
 
 ## [Unreleased]
 
+### Added
+
+- **Section 3.3.4: disclosed gaps in a receipt chain (#117).** Under a profile requiring action receipts, a specification that offers only "complete" and "broken" rewards concealment: an operator who backfills a lost receipt scores better than one who reports the loss. A `GapDisclosure` is a signed chain element stating that receipts which would have occupied its position were never emitted. Coverage is structural rather than asserted: the disclosure links back to the element before the gap, the next element emitted links back to the disclosure, and verification is two link checks a verifier already performs on every ordinary element. No range fields exist, because a hash chain cannot express a range and an emitter cannot know its successor's hash at write time.
+
+  The action-receipt outcome `receipt_missing_required` is narrowed to silent absence, and `receipt_gap_disclosed` is added beside it, distinct by requirement, with acceptance a verifier policy input. It never satisfies a profile requiring independently proven completeness: a disclosed gap does not establish that the missing receipts existed, how many were lost, or that omission was not selective. A disclosure at the live tail, where no successor exists to seal it, is unverified rather than disclosed or invalid: a chain truncated immediately after a disclosure is indistinguishable from an honest tail, so whatever the tail is granted, truncation is granted too. Conformance vectors in `examples/action-receipts/gap-disclosure/`, two per rule with a byte-for-byte generator; the tail case is pinned by its own test. Proposed and authored from production operation of a per-action receipt emitter; carried per the maintainer-carry provision in CONTRIBUTING.
+
+  Two cross-references that predate the renumbering which introduced section 3.3.1 are updated to name section 3.3.2, where the text they cite now lives.
+
 ### Fixed
 
 - **The exported `SCHEMA` was the live object the validator reads.** `_schema()` is `lru_cache`d and `_validator()` is built over whatever it returns, so the name exposed "for downstream tooling that needs the raw dict" and the validator's schema were one object: lowering `SCHEMA["properties"]["iat"]["minimum"]` made a record dated 1970 valid to `validate_json()`, to `iter_errors()`, and to the structural gate inside `sign.verify_record()`, for every later call in the process. Nothing about the call site looks wrong, since adapting the raw dict is the use the comment invites. It is a deep copy now; a shallow one would leave the nested `properties` dicts shared and the same edit would still land.
--- a/docs/verification.md
+++ b/docs/verification.md
@@ -264,15 +264,16 @@ Keep the verification results separate:
 | Action issuance evidence | Canonical action digest, receipt signature, trusted issuer key, session or call binding, chain order | Successful physical completion |
 | Outcome evidence | Controller or monitor decision carried by the receipt payload | Functional-safety certification unless the issuer and profile explicitly claim it |
 
-For action receipts, a verifier should distinguish five common outcomes:
+For action receipts, a verifier should distinguish six common outcomes:
 
 | Outcome | Meaning |
 |---|---|
 | `receipt_valid_accepted` | The receipt is well-formed, trusted, bound to the call, and reports acceptance. |
 | `receipt_valid_rejected` | The receipt is well-formed, trusted, bound to the call, and reports controller or policy rejection. This is valid negative evidence. |
-| `receipt_missing_required` | The profile required a receipt, but none was present for the consequential action. |
+| `receipt_missing_required` | The profile required a receipt, and none was present for the consequential action, with no valid `GapDisclosure` occupying its position in the chain. Silent absence, treated as presumptively adversarial (spec section 3.3.4). |
+| `receipt_gap_disclosed` | Required receipts are absent, and a valid `GapDisclosure` occupies their position in the chain: the emitter reported the loss and sealed the report into the chain. Emitter-attested negative evidence, distinct from silence; whether it is accepted is a verifier policy input (spec section 3.3.4). |
 | `receipt_invalid` | The receipt is present but fails signature, digest, freshness, ordering, or call-binding checks against a key the verifier holds. |
-| `receipt_unverified` | The receipt names an issuer key the verifier has not pinned, and nothing else failed. Per section 3.3.1 of the spec this is unverified, not invalid: the receipt confers no trust and proves no wrongdoing, surfaced with an advisory rather than a failure. |
+| `receipt_unverified` | The receipt names an issuer key the verifier has not pinned, and nothing else failed. Per section 3.3.2 of the spec this is unverified, not invalid: the receipt confers no trust and proves no wrongdoing, surfaced with an advisory rather than a failure. |
 
 The key boundary is that a valid rejection is not malformed evidence. It is
 evidence that the downstream authority declined the action. A valid acceptance
--- a/examples/action-receipts/gap-disclosure/01-gap-disclosed-valid.json
+++ b/examples/action-receipts/gap-disclosure/01-gap-disclosed-valid.json
@@ -1,0 +1,72 @@
+{
+  "name": "gap-disclosed-valid",
+  "description": "A disclosure spliced into the chain: it links back to a present element and the next element links back to it. Coverage is structural, so nothing is asserted about a range.",
+  "profile": "trace.action_receipt.conformance.v0",
+  "source": {
+    "issue": "agentrust-io/trace-spec#117",
+    "spec": "spec/trace-v0.2.md section 3.3.4"
+  },
+  "context": {
+    "session_id": "trace-session-2026-07-06T15:22:11Z",
+    "require_receipt": true,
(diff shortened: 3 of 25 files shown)


imran-siddique (approved):
Merging, and closing #117 with it. Nineteen days of silence from me on a question you asked precisely so you would not have to assume, then a day to deliver this. Thank you for waiting.

**The tail case is vector 17 and not a paragraph, which is what I asked for.** Its own description carries the argument:

> A well-formed disclosure at the live tail of the chain: signed by the chain key, linked to a present predecessor, and with no successor yet to seal it. Not invalid, nothing is defective, and not disclosed either: half a splice covers nothing until the chain resumes and the seal can be checked. Re-verification after resumption upgrades it.

And the test file states the reason it must be its own case:

> indistinguishable from an honest live tail, so whatever outcome the tail gets, a deliberate truncation gets too

That is the whole argument for the refusal, in one sentence, sitting next to the thing that enforces it. A disclosure mechanism whose best outcome is reachable by disclosing where nothing can seal the claim would have converted an honesty feature into a laundering one, and you found it in your own text while applying someone else's review.

**Twenty vectors with a generator rather than twenty hand-written fixtures.** `gen_gap_disclosure_vectors.py` at 461 lines means a future revision can be shown to change or preserve outcomes rather than argued about, and it means nobody has to trust that fixture 11's confusable ancestor key differs from fixture 7's unknown key in the way the filename claims. The adversarial set is the right set: dangling predecessor, successor that does not link, contradicted, foreign key, tampered, case variants on both key and stream, replayed stream, four separate tail conditions, and both unknown and absent type versions.

All six checks green: `test (3.11)`, `test (3.12)`, `check`, `Analyze (python)`, `CodeQL`, `gate`.

**What I am not claiming.** I have read the structure, vector 17, and the test file's reasoning about the tail. I have not line-read all 109 lines of §3.3.4 or all twenty vectors. I am merging on the generator, the green suite, the adversarial coverage and the fact that the one case most likely to be wrong is the one you found and pinned yourself. Carrying it as accepting maintainer under the arrangement on #117: authorship yours, the requirement mine to answer for.

**Expect one rename.** `receipt_gap_disclosed` is now the sixth outcome across this codebase needing a first-class "not established" state, alongside trace-spec#246 on revocation, #190 on unresolvable citations, trace-tests#82's SKIP, agent-manifest#348's undecidable admissibility and #298's `indeterminate`. Unifying that vocabulary is mine to do. The semantics here are right and will not change; the name might.

@carloshvp, your five review points are all in this and it went straight to a PR because of that.

Issue #245, opened 2026-08-29
test: independent-producer evidence for the parent-record canonicalization boundary


chernistry:
`delegation.parent_record_hash` is the field a verifier walks to reconstruct a delegation chain, and nothing normative says what it is a digest *of*.

At `e7e2eca`:

- The schema gives it a format and a sentence: `"pattern": "^sha(256:[0-9a-f]{64}|384:[0-9a-f]{96})$"`, described as "SHA-256 or SHA-384 digest of the parent hop's Trust Record."
- `spec/trace-v0.2.md` does not mention the field at all (zero occurrences).
- `docs/rfcs/a2a-delegation-profile.md` §4.1 does state a preimage, and D-3 (`parent_not_found`) is defined in terms of it — but that document says **"Status: Draft proposal. Binds nothing."**

"the parent hop's Trust Record" has at least three natural readings, and they give different digests:

1. The published file's bytes.
2. RFC 8785 (JCS) over the complete record, signature included.
3. RFC 8785 over the record with `signature` removed — the same body the signature itself covers.

I computed all three against our own delegated-parent record. Readings 1 and 2 agree on it and match the child's link; reading 3 gives a different digest. So two implementers reading only the schema can both act in good faith and produce chains that resolve to dangling links against each other, with no diagnostic on either side saying why.

## The complete-record reading is the right one, and it should be normative

§4.1 already takes it, for a reason worth promoting out of a draft: a digest over the signed body alone commits to *what the parent said* and not to *who said it*, so any holder of any key can re-sign bytes that satisfy the same link. That is the difference between a chain that binds hops and one that binds statements.

Asking for that sentence to move somewhere that binds — the schema description, or `spec/trace-v0.2.md` — rather than for a new rule.

## One reading is a trap rather than a disagreement

Plain `json.dumps(obj, sort_keys=True, separators=(",", ":"))` over the complete record produces the same digest as JCS on records like ours, so an implementer who reaches for it interoperates and has no reason to look further.

It stops matching the first time a record carries a non-ASCII character, and the cause is a default rather than a choice: `json.dumps` escapes non-ASCII to `\uXXXX` unless `ensure_ascii=False` is passed. Putting `模型-π` in `model.model_id` and `v1-é` in `policy.version`, both of which the schema accepts:

```
JCS                              91eca6bbe63ebe2ffeba09c0...
json.dumps (default)             fd74a7636b2665d69185150c...   diverges
[...]

PR #270, opened 2026-08-31
test(schema): probe every pattern constraint, and hold the mirroring models.py claims

chernistry:
Closes the generator half of #247.

The matrix in `test_the_schema_and_the_models_agree.py` answers "do the two validators disagree about any of these values". It cannot answer "is each constraint discriminated at all" — whether a value reaches the constraint's own boundary, or fails so early that both validators reject for an unrelated reason and agree by accident. Two gaps followed from that, and neither produced a failing test.

**Half the pattern constraints were unreachable.** `BASE` carries five of the schema's ten. `model.weights_digest`, `delegation.parent_record_hash`, `references[].retention`, `references[].digest` and `signature` are optional, absent from the fixture, and therefore never mutated. A sweep cannot disagree about a field it never sets, and a green run says the same thing either way.

**The mirroring was claimed and not held.** `models.py` says its pattern constants are "mirrored verbatim in `schema/trace-claim.json` and its copy, and held there by `tests/test_the_schema_and_the_models_agree.py`". They were not held here. Two of the ten are pinned in `test_references_block.py`; the other eight were maintained by hand, under a comment naming this file.

## What it does

`_pattern_constraints()` reads the ten patterns **out of the schema** rather than listing them, so a pattern added to the published artifact cannot be added without appearing here. `FULL` extends `BASE` so every one of those ten paths exists. `_probes(pattern, valid)` derives boundary values from the pattern text and from a valid instance of it — structural prefixes of the valid value, literal runs taken out of the pattern, case flip, one character shorter and longer, one character outside the class, and the empty/space/tab values a producer sends for an unset field.

Then, per constraint, one of three outcomes and nothing silent:

- **Mirrored** — the model constrains the field with the same pattern string the schema publishes. No string can split them, and saying so is a proof rather than an observation about the values that happened to be tried. This is where all ten sit today.
- **Split** — a probe the two disagree about, which goes into `DECLARED_DIVERGENCES` with its reason, not into a second exemption list.
- **Neither** — different patterns and nothing here can tell them apart. That is the state `models.py` already claims is impossible, and the one that produced #244. It fails unless a human writes down why, in `UNMIRRORED_AND_UNSPLIT`, which is two-sided the same way: an entry that becomes mirrored has to leave the set.

## On the three steers

1. **The floor is discrimination, per pattern, and it is stronger than "at least one splitting value".** For a mirrored constraint no splitting value can exist, so demanding one would be demanding a value that cannot be constructed. The test asks instead for the fact that makes that true — the schema's pattern string is one the model carries — and falls through to requiring a declared splitter for any constraint where it is not. The old floors are untouched; both sweeps run.
2. **`DECLARED_DIVERGENCES` is reused, not duplicated.** No new entries: the generator finds no split against the current artifacts. `UNMIRRORED_AND_UNSPLIT` ships empty and exists so the third case cannot pass in silence. I read the computed `_URI_FORMAT_ENFORCED` exemption before touching the assertions it feeds — `format: uri` is not a `pattern`, so it is outside the ten and the new tests do not reach it.
3. **The tagged-artifact sentence** — noted, and separately: `tests/test_validate.py:142` already pins the two schema copies against each other, so that part of #247's third bullet is covered. What is not covered is a producer's claim against a *published* version, which is the precedence text rather than a test.

## Verification

`1027 passed, 1 skipped`; `ruff check src tests scripts` clean; `mypy` clean on the file.

Each new test was verified to fail when the thing it guards is broken, rather than only to pass today:

| counterfactual | result |
|---|---|
| drop `signature` from `FULL` | coverage floor fails |
| `_DIGEST_RE` changed so it no longer mirrors the schema | mirror test fails |
| a stale `UNMIRRORED_AND_UNSPLIT` entry | staleness test fails |
| `subject` restored to the pre-#244 prefix pattern `^(spiffe://\|did:)` | fails, and the generator splits the validators on `spiffe://` |

That last row is the one worth reading. `spiffe://` is not written anywhere in this file — it comes out of the pattern's own literal runs — and it is the class the hand-written matrix could not reach, because every value in it fails a prefix test as well, so both validators rejected and agreed.

The preimage bytes for #245 are posted on that issue.

### DIFF
--- a/tests/test_the_schema_and_the_models_agree.py
+++ b/tests/test_the_schema_and_the_models_agree.py
@@ -341,7 +341,18 @@ def _probes(pattern: str, valid: str) -> list[str]:
 
 
 def _splitters(path: tuple[str, ...], pattern: str) -> list[tuple[str, bool, bool]]:
-    """Probes at *path* the two validators disagree about."""
+    """Probes at *path* the two validators disagree about.
+
+    Raises ``LookupError`` naming *path* when ``FULL`` does not reach it, rather than
+    letting the plain ``KeyError`` out of ``_get`` propagate. This is the same gap
+    ``test_the_fixture_reaches_every_pattern_in_the_schema`` already reports by field
+    name via ``_reachable``; this read the path unguarded.
+    """
+    if not _reachable(FULL, path):
+        raise LookupError(
+            f"{'.'.join(path)}: pattern {pattern!r} constrains a path FULL does not "
+            "reach, so nothing here can probe it"
+        )
     valid = _get(FULL, path)
     split: list[tuple[str, bool, bool]] = []
     for probe in _probes(pattern, valid):
@@ -534,7 +545,11 @@ def _unaccounted_patterns(
         dotted = ".".join(path)
         if model_field_patterns.get(path) == pattern:
             continue
-        split = _splitters(path, pattern)
+        try:
+            split = _splitters(path, pattern)
+        except LookupError as exc:
+            unaccounted.append(str(exc))
+            continue
         declared = [s for s in split if (dotted, repr(s[0])[:26]) in DECLARED_DIVERGENCES]
         if split and len(declared) == len(split):
             continue
@@ -641,4 +656,27 @@ def test_the_mirror_check_is_by_field_not_flat_membership() -> None:
         "a schema pattern reassigned onto a field the model does not constrain with it "
         "must be flagged, even though the pattern string itself is genuinely in use "
         "elsewhere in the model:\n" + "\n".join(unaccounted)
+    )
+
+
+def test_a_missing_path_fails_with_the_field_name_not_a_keyerror() -> None:
+    """Review counterfactual: a pattern on a field ``FULL`` omits must fail naming
+    that field, not with a raw ``KeyError`` out of ``_get``.
+
+    ``test_the_fixture_reaches_every_pattern_in_the_schema`` already catches the same
+    reachability gap cleanly, by field name, via ``_reachable``. ``_splitters`` is the
+    other place that reads an arbitrary schema path out of ``FULL``, and it read the
+    path unguarded, so the same gap surfaced there as a bare ``KeyError`` instead.
+    """
+    bogus_path = ("build_provenance", "does_not_exist_in_full")
+    assert not _reachable(FULL, bogus_path), "setup: the path must genuinely be absent from FULL"
+
+    with pytest.raises(Exception) as excinfo:
+        _splitters(bogus_path, "^irrelevant$")
+
+    assert not isinstance(excinfo.value, KeyError), (
+        f"a missing path must not surface as a raw KeyError: {excinfo.value!r}"
+    )
+    assert "build_provenance.does_not_exist_in_full" in str(excinfo.value), (
+        f"the failure must name the missing field; got {excinfo.value!r}"
     )
--- a/tests/test_the_schema_and_the_models_agree.py
+++ b/tests/test_the_schema_and_the_models_agree.py
@@ -63,10 +63,11 @@ import copy
 import json
 import re
 from pathlib import Path
-from typing import Any, get_args
+from typing import Any, get_args, get_origin
 
 import jsonschema
 import pytest
+from pydantic import BaseModel
 
 from agentrust_trace import TrustRecord, validate_json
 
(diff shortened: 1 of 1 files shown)


lywinged (dismissed):
Approve. This is the generator from #247, and it does what was asked: it turns the subject finding into a property every pattern constraint is held to, rather than a second fixed matrix.

I ran it against the two regressions it claims to catch, on both sides of the split, since a discrimination test is only worth its own counterfactual.

- Put the pre-#244 prefix pattern back on `subject` in the schema: `test_every_pattern_is_mirrored_or_split_or_declared` fails and names `spiffe://` and `did:`. The matrix carries those two literals today, but only because they were hand-added as the guard after #244 exposed the miss; the generator derives them from the pattern, so the same prefix shape on a field nobody has patched is caught without anyone thinking to list it.
- Drop the `signature` pattern from the model, a field the old sweep never touched because it is optional and absent from `BASE`: the same test fails and names the splitting probes. So the mirroring check has teeth on the five optional pattern fields `FULL` reaches that `BASE` did not (`weights_digest`, `parent_record_hash`, `references[].digest`, `references[].retention`, `signature`), which is the coverage that did not exist before this.

Two things I checked that hold: the ten patterns are read out of the schema, so a new one cannot be added without this file failing until it is accounted for; and the three-indirection walk in `_model_pattern_strings` is load-bearing, `signature`'s pattern is found only by the full walk and drops out of a shallow one, which is the miss the docstring describes.

Two non-blocking notes, neither a reason to hold it:

1. The mirror match at `if pattern in model_patterns` is by string against a flat set, not by field. It is sound today because every field carrying the shared `sha...` digest pattern is genuinely constrained by it in the model, which I checked by injecting a malformed digest at each. It would stop being sound if a later field took that pattern in the schema and was left unconstrained in the model: the string would still be in the set and the field would read as mirrored. Matching the pattern to the field's own model constraint would close that, whenever it is convenient rather than now.
2. A pattern on a field absent from `FULL` makes `_splitters` raise `KeyError` from `_get(FULL, path)` rather than the clean message `test_the_fixture_reaches_every_pattern_in_the_schema` is written to give. The reachability test goes red in the same run, so nothing slips, but the message a maintainer sees is the raw exception rather than the field name.

Suite is green here and the change is one file over the current head.

chernistry:
Thanks for running both counterfactuals rather than reading the diff — the `signature` one is the case I most wanted checked, since it is exactly the coverage `BASE` never reached. On your two notes: I would rather fold both in here than defer them, since each is a few lines — matching the pattern to the field's own model constraint, and catching the missing path so the reachability failure prints the field name instead of a raw `KeyError`. Say the word if you would prefer the PR stay as approved and both land as a follow-up. Separately, the `gate` check is red here and fails in about five seconds; it fails the same way on #234, so I suspect it is repo-side rather than this branch — is there something I should do on my end, or does it need a maintainer?

chernistry:
Both notes are folded in.

- Mirror match by field, not flat-set membership: e5f6cdb
- \`_splitters\` names the field instead of raising a bare KeyError: 4d3efd1

Each commit adds a test that fails for the stated reason on the prior commit and passes after it; full suite, ruff, and mypy are clean on the branch.

imran-siddique (approved):
Merging. This is the sharper half of #247 and it is not the half I filed.

I opened #247 on the premise that four surfaces state the rules and nothing says which wins. @lywinged's #248 then audited all ten schema patterns against the model and found one divergence, which I recorded as narrowing the issue. You have found the reason that audit could not have said what I took it to say:

> A sweep cannot disagree about a field it never sets, and a green run says the same thing either way.

Five of the ten patterns are on optional fields absent from `BASE`, so they were never mutated and never reached their own boundary. The test answered "do these two validators disagree about the values I tried" and I read it as answering "are these ten constraints consistent". Those are different questions and only one of them was being asked.

That is the same defect class the test exists to catch, one level up: something that looks like coverage, reports green, and is not discriminating. A test that cannot fail is worse than an absent one, because the absent one does not stop anyone looking.

**Probing each constraint's own boundary is the right fix**, and holding `models.py`'s mirroring claims rather than trusting the comment that says they mirror is the other half. A constant asserting it is "mirrored verbatim" is a claim, and until something compares the two it is an unverified one.

Merging, and I am updating #247 with what this changes about the issue.

chernistry:
Thanks for the writeup, and for the extra scrutiny on this one — the sweep-vs-audit distinction (a test that can't disagree on a field it never sets vs. one that probes each field's own boundary) is a good general lesson to keep in mind past this specific fix.

PR #277, opened 2026-09-03
docs(rfc): propose runtime evidence, and what a verifier may conclude without it

imran-siddique:
## The gap

A Trust Record states a hardware platform and a measurement, and carries nothing a relying party can check either against. `runtime` is `additionalProperties: false` over five members, none of which is the quote, the certificate chain, or the report signature. `appraisal` is a status enum plus a `verifier` URI, which is a pointer to a party who says they checked.

`verify_record()` performs no attestation verification, and the package contains none: the string `quote` appears in `src/agentrust_trace/` only inside comments. That is not an implementation gap. There is nothing in the record for such code to consume.

So the chain a relying party actually gets is: the record is signed, the signature is sound, and it attests **authorship**. Every hardware claim inside is the issuer's word, and a verifier reporting "TEE-attested" on that basis is repeating the issuer's assertion in the verifier's voice.

## Why this is our own doctrine, not a new idea

The spec already reasons about assurance laundering, carefully, and in one direction only.

§3.1.1 forces `software-only` onto an assembled record because "an importer holding someone else's log has no quote to present". §3.1.2 rule 3 forbids treating a resolved `references` entry as attested evidence, naming assurance laundering as the reason. Both govern pointers to things **outside** the record. Neither governs `runtime`, where the same move is not merely possible but is all that is on offer: `rim_uri` is a pointer, `measurement` is a transcription, `appraisal.verifier` names someone else's verdict.

It is also a requirement already made and never made checkable. `docs/trust-levels.md:50` requires a Level 1 signing key to be generated inside a verified TEE, and `spec/trace-v0.2.md` §5 maps `runtime` to "RATS Evidence + vendor RIM". Nothing represents either. The requirement, the mapping, and the schema disagree, and the schema is what runs.

## What this adds

- **`runtime.evidence`** carrying the quote, in `schema/trace-claim-v0.3-draft.json`, generated from the v0.2 schema so the diff is exactly one member deep.
- **Seven verification rules**, envelope-first, so the record signature binds a quote to *this* record rather than to any record.
- **Three grades** rather than a boolean: `unattested`, `platform-attested`, `attested`. The middle one was forced by the artifacts, not chosen.
- **§6.1**: a grade on the record does not propagate to the claims inside it.

Additive. No existing field, schema, or record changes. A record without `evidence` grades `unattested`, which is a name for what every v0.2 record already was.

## Conformance material

Nine vectors over **two genuine Intel TDX v4 quotes** from a GCP C3 guest, verified by `agent-manifest`'s verifier imported unmodified. Nothing is minted. TRACE does not implement attestation and this proposal does not start.

```
9/9 vectors behaved as the profile says they must (1 of them documenting a limit).
```

## Three findings, all kept rather than smoothed

**The substitution vector was written to be a rejection and is not one.** Both captures come from one TD, so they share an MRTD, and the measurement rule cannot separate two quotes that agree on the measurement. The rule binds a record to a *measurement*, never to a *quote*. That caps the middle grade at "genuine silicon reporting this measurement" and is the argument for the top grade existing. A synthetic corpus would have minted different measurements and passed.

**The top grade is currently unreachable.** No capture we hold binds a record-signing key, and the pre-image behind the bindings we do have was never committed, so `REPORT_DATA` is a value that verifiably came from the TEE and cannot be opened. Both are capture-procedure gaps, cheap to close on the next TDX run, and cheaper to say now than to discover during adoption.

**The first draft of §6.1 contained the bug §6.1 forbids.** It graded the model claim by reading `evidence.binds`, which would let a producer raise its own claim by writing a string. §3 already said a verifier recomputes `binds` and never reads it; the implementation forty lines away did the opposite. Same shape as the gap in `runtime` this whole proposal is about. Fixed in the second commit, with a vector that declares a binding it does not have and stays self-reported.

## Status

Informative, lowercase keywords, per `CONTRIBUTING.md`. **Binds nothing.** It does not resolve §5.2: the reference producer binds a manifest digest rather than the signing key today, and a profile is not adopted by writing down which side should move.

https://claude.ai/code/session_01LKXk6sx5CzAsHmTEJhk7Gy


lywinged (changes requested):
Changes requested on `efad21c`. Item 1 fails the suite and item 2 is a rule contradicted by its own reference implementation; the rest are edits to the text. All of it reproduces at the pins below.

Reproduced at the same pins: this head `efad21c` on base `5c69dc1`, agent-manifest at `9348097`, cryptography 50.0.1. `generate.py` prints nine rows matching the table in §7, the accept vector measures 11,923 bytes compact against 1,184 without evidence, and the two captures parse to one MRTD with different REPORT_DATA, so the §7.1 limit reproduces from the captures and not from anything the script does. For the next run: the imported verifier reads `not_valid_before_utc` at `_cert_chain.py` line 59, which cryptography 41 does not have, so 41 fails with an AttributeError. agent-manifest's pyproject pins `cryptography>=42,<51`, but the script imports it by path rather than installing it, and the README's own dependency line names no version.

Four items, each visible from the repository's own checks or text.

1. The suite fails. `tests/test_safe_integer_range.py::test_every_schema_in_the_repository_is_classified` enumerates `schema/` and requires every file to be listed as bounded or unbounded, and `trace-claim-v0.3-draft.json` is in neither list. 1 failed, 1,166 passed, 1 skipped.

2. Rule 2 and the reference implementation disagree. §4 rule 2 says a verifier with no implementation for a `format` "treats the record as if the block were absent" and grades `unattested`. §7 says `generate.py` implements §4, and line 143 raises `Reject` for any format other than `tdx-quote-v4`, so a record carrying `sev-snp-report-v2` or `tpm2-quote`, both in the draft enum, is refused rather than downgraded. Either the script grades `unattested` on an unknown format or rule 2 changes; the rule's second sentence argues for the first.

3. Line 6 of the RFC says 8 vectors. §7 counts 9/9, the README table has nine rows, and the PR body says nine; the ninth came in with the second commit, which updated §7 and the README but not line 6.

4. The other two RFCs are in the mkdocs nav, at lines 226 and 227 of `mkdocs.yml`; this one is not yet. `mkdocs build --strict` passes with or without the entry.

Two things the rules leave unstated, found by running inputs outside the nine.

REPORT_DATA is 64 bytes, and `appraise` at line 177 and `grade_model_claim` at line 208 both compare the first 32 against their candidate. With both comparisons on the same 32 bytes, `attested` and `model claim: attested` cannot both hold for one record. `binds` being a single enum suggests that is intended; if so, §6 should say a model claim can be attested only on a `platform-attested` record. Both captures carry zeros in the second 32 bytes, so nothing in the corpus reaches the question, and the text does not say whether that half is reserved, ignored, or where a second binding would go.

§5.1's cost is measured on the 8,000-byte capture files, not on the quote structure. Both files declare a structure that ends at byte 4,935, and the remaining 3,065 bytes are zeros the verifier never reads: any prefix of 4,935 bytes or more verifies, bytes appended verify, and a byte flipped inside the tail verifies. Trimmed to the structure, the quote is 6,580 characters and the accept vector is 7,836 bytes compact rather than 11,923. §5.1 can say which of the two it measures.

The same fact bears on the schema's "Byte-identical to what the hardware produced". No verifier can check that against the original, so the description is a producer obligation; the text should say so, or say what a verifier may do with bytes past the declared structure, since the corpus itself carries 3,065 of them. `collateral` is the same shape: §3 says the record should state it, §4 never reads it, so `required` on a TDX record grades the same as `embedded`. Rule 3 already refuses a `platform` the evidence does not root; the same check would refuse `required` on `tdx-quote-v4`.

None of this touches the shape of the proposal. The envelope-first order, the middle grade, and the recompute-never-read rule in §6.1 all do what the text says, and the limit vector reports what the rules do.

imran-siddique:
@lywinged this is the review I wanted and did not get from myself. All four items reproduce, three are fixed in `d0fdf88`, and the two things you found by running inputs outside the corpus are the more valuable half. Answering everything.

## Item 1, already fixed before your review landed

`dddcc70` added `schema/trace-claim-v0.3-draft.json` to `BOUNDED_SCHEMAS`, so `test_every_schema_in_the_repository_is_classified` passes. Your reproduction at `efad21c` was accurate; the commit crossed with it.

## Item 2, you were right and the rule wins

The script disagreed with itself, which is the part I should have caught: `generate.py` refused any format other than `tdx-quote-v4`, while the by-reference branch **six lines below** already returned `unattested` with the same reasoning rule 2 gives. Fixed in the direction you argued for, and for the reason rule 2's second sentence gives: refusing makes a verifier's coverage gap indistinguishable from bad evidence, and a record carrying sound SNP evidence is not defective because this verifier cannot read SNP.

Added `downgrade-unsupported-format` so it is in the corpus rather than only in the text. It carries the genuine quote bytes under a `sev-snp-report-v2` label with `platform: amd-sev-snp`, which incidentally demonstrates **rule 2 short-circuiting before rule 3**: the verifier cannot read the evidence, so it never reaches the question of whether platform and evidence agree. Corpus is 10/10.

The size pin in `test_the_corpus_did_not_shrink` moves 9 to 10. That test exists to make a shrinking corpus a visible act, and growing it should be equally visible, so bumping the constant is the intended cost rather than a workaround.

## Items 3 and 4

Line 6 now says 10, matching §7, the README and the corpus. Nav entry added after the A2A profile at `mkdocs.yml:228`.

## REPORT_DATA's second 32 bytes

You are right that nothing in the corpus reaches it and the text does not say what it is. **It is reserved, and the profile should say so explicitly rather than leave it to be discovered from the code.**

The reason to reserve rather than define: the first 32 bytes carry the binding this profile is about, the `cnf` key commitment, and that is the only thing rule 6 needs. A second binding in the remaining 32 would be a second commitment made by the same guest-controlled write, so it inherits exactly the trust properties of the first and adds no independent evidence. Defining it now would invite producers to put something load-bearing there before anyone has said what a verifier may conclude from it, which is the mistake §6 exists to prevent.

I will add a sentence to §5.2 saying the second half is reserved, that a verifier MUST NOT derive a grade from it, and that a producer writing into it is making a claim this profile does not evaluate.

## §5.1's byte cost, which is the sharper finding

You are right and the number is wrong in the direction that flatters the proposal. The captures are 8,000 bytes, the declared structure ends at 4,935, and the remaining 3,065 are zeros no verifier reads. So §5.1's 11,923 measures padding as if it were cost.

**§5.1 should measure the declared structure, so 7,836 bytes compact.** That is the honest number and it is the one an implementer sizing a record cares about, since nobody ships the padding deliberately.

Your related point lands harder. The schema says the quote is "byte-identical to what the hardware produced", and no verifier can check that against an original it does not have, so it is a producer obligation stated as if it were a verifiable property. Two things follow, both of which I will write in:

- the description says it is a producer obligation, not something a verifier establishes;
- the text says what a verifier may do with bytes past the declared structure. Given the corpus itself carries 3,065 of them and they verify either way, the honest rule is that a verifier evaluates the declared structure and MUST NOT treat trailing bytes as evidence of anything, in either direction.

## `collateral`

Also right. §3 says a record should state it, §4 never reads it, so `required` and `embedded` grade identically on a TDX record. Rule 3 already refuses a `platform` the evidence does not root, and the same check applied to `collateral` would refuse `required` on `tdx-quote-v4`, where the PCK chain travels inside the quote. I will make that rule 3's second clause rather than a new rule.

## What is not in this push

The three text changes above, §5.2's reserved sentence, §5.1's measurement basis, and `collateral` in rule 3. They are wording rather than behaviour and I would rather land them in one edit after you have said whether the reserved-not-defined call is the one you would make. The code and the corpus are consistent as of `d0fdf88`.

Also noting your environment finding for anyone who runs this next: the imported verifier reads `not_valid_before_utc` at `_cert_chain.py:59`, which cryptography 41 lacks, and the README names no version. I will pin it in the README.

lywinged (commented):
`d0fdf88` checked: 10/10 against agent-manifest `9348097`, and regenerating with the published key reproduces all ten committed vectors byte for byte. Suite 1,214 passed, 1 skipped. Line 6 and `mkdocs.yml` line 228 as described.

Reserved is the call I would make, for two reasons beyond the one given. The producer already in use writes it that way: `TDXProvider.extend_manifest_hash` requests `sha256(pre_image) || 0x00*32`, and the verifier on the other side compares the first 32 bytes only. And rule 6 has to recompute from the record alone, and the record-signing key is the only bound object a record carries, so a second slot would have nothing to hold that a verifier could check offline. On wording: say the verifier does not read the second 32 bytes, rather than requiring them to be zero, so a producer that fills all 64 is not refused for it.

One question the reserved sentence brings forward, for the text and not for this push. §6.1 grounds an attested claim in what hardware measured. `grade_model_claim` recomputes from REPORT_DATA, which the guest writes when it asks for the quote, so a weights digest committed there is the producer's statement, made earlier and inside the envelope. The registers that hold measurements are the RTMRs. The parser already exposes them, rule 4 reads only the MRTD, and both captures carry RTMR0 to RTMR2 set with RTMR3 zero. Two ways to close it: point the model claim's recompute at a runtime register, or have §6.1 say that a REPORT_DATA commitment is a claim rather than a measurement and grade it so. Either way REPORT_DATA then carries the key and nothing else, and the exclusivity I raised stops being a question.

A smaller one in the same area. Rule 6 speaks of the platform's guest-controlled field. agent-manifest's own notes say that on Azure confidential VMs the guest does not control it, the paravisor binds the vTPM AK there, and the manifest hash is bound by extending a PCR under an AK-signed quote. `tpm2-quote` is in the enum with no binding rule, so on that path the top grade is unreachable by construction rather than by capture. A line saying the binding rule is per format, as the measurement rule already is, would cover it.

The reserved sentence as drafted in the comment carries an uppercase keyword; line 9 of the RFC says lowercase throughout, and the docs guard wants a citation on the same line as any uppercase one.

I will approve on the head that carries the five text changes: §5.2 reserved, §5.1 measured on the declared structure, the `quote` description, `collateral` in rule 3, and the README pin. The two questions above are not conditions.

altrudev (commented):
DDC radial review on current head `d0fdf88`: two boundaries look load-bearing and are not covered by the existing review.

1. **The reference rule implementation does not use the trust context the RFC says a verifier is given.** §2 lists `trusted_root_keys` as context because "a record naming its own key as trusted is not evidence". But `generate.py::check_envelope(record)` takes only the record, reads `record["cnf"]["jwk"]`, and verifies the record signature against that embedded key. `appraise()` calls that function with no externally trusted key or resolver. So the corpus establishes only that the record is self-consistent under a key it supplies; it does not establish that the signer is a trusted issuer. This is the same trust-boundary `agentrust_trace.verify_record(record, trusted_jwk)` deliberately keeps external. A producer can generate its own key, sign an otherwise valid record around a genuine quote, and reach `platform-attested` without any trust decision over the signer. If the grade is intentionally orthogonal to signer trust, §2 should say `trusted_root_keys` is not part of this grading function and the result must be combined with a separate signer-trust result. If signer trust is part of the grade, the reference implementation needs to consume external trust context rather than the embedded JWK.

2. **The draft calls itself v0.3 while still carrying the v0.2 profile identifier.** `schema/trace-claim-v0.3-draft.json` has `$id`/title for v0.3 but `eat_profile.const` remains `tag:agentrust-io.com,2026:trace-v0.2`; `generate.py` likewise sets `PROFILE` to the v0.2 URI. This matters because `runtime.evidence` is not additive to a closed-world v0.2 validator: current v0.2 `runtime` has `additionalProperties: false`, so a v0.2 verifier rejects the new member. The proposal itself targets v0.3, and #116 exists specifically to make unsupported profile versions refuse rather than parse under semantics they do not implement. Keeping the v0.2 identifier means the wire artifact cannot signal which semantics it requires. The draft can still be generated mechanically from v0.2, but the profile URI should move with the semantic version boundary before the vectors are treated as v0.3 examples.

I have not found either point already raised in the thread. Both are independent of the existing REPORT_DATA/collateral/size findings and do not challenge the proposal's core three-grade model.

lywinged:
@altrudev I checked both boundaries at `d0fdf88`, against the schemas in this repository rather
than against the argument. The second one holds and the measurement is sharper than the way it is
put; the first needs narrowing.

**The profile point holds.** All ten vectors declare `tag:agentrust-io.com,2026:trace-v0.2`.
Validated against the published `schema/trace-claim.json`, at `f24e1c1` and again at this
branch's merge base `5c69dc1`, nine of them fail with "Additional properties are not allowed
('evidence' was unexpected)", because `runtime` and the top level are both `additionalProperties:
false`. The tenth, `downgrade-evidence-absent`, passes, and only because it carries no evidence:
removing the member from a failing vector makes that vector pass, and adding one to the tenth
makes it fail, so the member is the whole cause. All ten validate against
`schema/trace-claim-v0.3-draft.json`. That draft also pins `eat_profile` to `"const":
"tag:agentrust-io.com,2026:trace-v0.2"`, with a description calling the record a TRACE v0.2 Trust
Record.

So the wire artifact declares v0.2 and cannot be accepted by a verifier validating against the
schema that profile names. That lands on §8: "It does not change any existing field, and adds no
requirement to any v0.2 record." True of the fields, and these vectors are still records that
announce a profile whose published schema refuses them. `eat_profile` is the only version signal
they carry, so it is the only thing available to tell that verifier not to try.

The two ways out differ in cost. Minting a v0.3 tag changes a value the specification fixes at
`spec/trace-v0.2.md:132`, and which `docs/schema.md:15` says a record must carry, so it is a
decision larger than this proposal. Saying in §7 that the corpus carries a placeholder profile
and is not v0.2-valid costs a sentence, and it removes the trap for whoever copies a vector as a
starting point.

**The trust-context point needs narrowing.** What is true: `appraise(record)` takes the record
and nothing else, `check_envelope` verifies the signature against the key the record itself
carries, and none of `trusted_root_keys`, `accepted_measurements`, `platform_verifiers` or `now`
appears anywhere in `generate.py`.

What does not follow is that a producer signing with its own key and reaching `platform-attested`
is a gap. That grade is defined as exactly this case. §6 lists what it does not establish, first
item: "that this record's signer ran inside that TEE". Whether the signer is an identity you
accept is `trusted_root_keys`, and §2 keeps that out of the record deliberately, on the ground
that a record naming its own key as trusted is not evidence. Grading is not authorization here,
and the top grade does not become one either: it says the record-signing key is committed to
inside the TEE, not that the key is one you trust.

The residue is worth keeping and it is editorial. §2 enumerates four context items as what a
verifier is given, and the seven rules in §4 consume one, the platform verifier that `format`
selects. Either name which rule consumes which item, or narrow the table. As it stands §2 reads
as though the rules use all four, and the reference implementation is the thing that shows they
do not.

Neither is among the five text changes my approval is pending on. The profile line is worth
settling before these vectors are cited as examples, since a reader who copies one gets a record
the published schema rejects. The §2 one can ride along with the rest.

Tool-assisted: the schema validation runs and this write-up.

Qiang-Xu (dismissed):
It looks good to me.

lywinged (changes requested):
Requesting changes on `d0fdf88`, and correcting my own review first.

My 4 September review was submitted as **Approve**. Its own text says I would approve on the head that carries the five text changes, which is the opposite of what the state says. The badge was wrong, not the sentence, and this review supersedes it. I should have left it as a comment.

On the five: you said you would land them in one edit once I said whether reserved rather than defined was the call I would make. I answered on 4 September that it is, with the two additional reasons, so nothing is waiting on me there. They are §5.2 reserved, §5.1 measured on the declared structure, the `quote` description as a producer obligation with a rule for trailing bytes, `collateral` as rule 3's second clause, and the README pin.

What has arrived since is @altrudev's second point of 6 September, and I agree with it. The draft targets v0.3 while `eat_profile.const` stays at the v0.2 URI, so the wire artifact cannot say which semantics it requires. I am not restating his argument. What I can add is what it costs against a released library rather than against a closed-world reading.

**Nine of the ten vectors are refused by agentrust-trace 0.10.0 as published.** Run each through `validate_json` from the published package: one accepted, nine refused, every refusal on `Additional properties are not allowed ('evidence' was unexpected)`. The accepted one is `downgrade-evidence-absent`, which carries no evidence member. Validating directly against `schema/trace-claim.json` at this head gives the same nine. `schema/trace-claim-v0.3-draft.json` accepts all ten while pinning the same const, so the two schemas disagree about the corpus and nothing in a record says which to apply.

This is not a draft rule biting early. `verify_record` did not validate against the schema in 0.9.0 and does in 0.10.0, so the corpus is refused by the shipped verifier today.

The same boundary arrived from the other side on agentrust-io/integrations#170, opened 7 September: a downstream mapper whose records passed 0.9.0 and still pass `trace-tests verify --level 0` is refused by 0.10.0 for missing required members. That one is about members the schema requires and this is about a member it forbids, but both are a record valid under one reading of v0.2 and invalid under another. Whatever settles the identifier here settles that too.

**Why this blocks rather than waits.** The corpus is the proposal's evidence and the thing an implementer copies. Nine of the ten cannot be verified by the released library, so anyone working from the examples emits records that are refused, and conformance material that does not itself conform teaches the wrong shape. #116 is open on precisely this: a v0.2 verifier refuses these records for the wrong reason, `additionalProperties` at `runtime`, while their `eat_profile` satisfies the v0.2 const it checks, and a verifier that does implement the new semantics has no way to know it may accept them. The timing principle from #178 points the other way here, and that is the point. There, moving the key material early would have made downstream migrate twice, so waiting was right. Here nothing has pinned these vectors yet, so moving the identifier before they are published costs no migration and moving it after costs one.

So the edit you were holding is now two things rather than one, and the second is the one I would not decide for you. I will approve on a head that carries the five and a corpus that validates under a single declared profile.

pforest (dismissed):

lywinged (dismissed):

lywinged (approved):
Approving on `4272703`, the head that carries the two text corrections and closes the model-claim boundary.

Verified against the pinned verifier, `agent-manifest` at `9348097`, and both genuine captures, on Python 3.11.15 with cryptography 50.0.1:

- `pytest`: 1227 passed, 1 skipped.
- `examples/runtime-evidence/test_appraisal.py`: 20 passed, all 13 vectors at their committed grades and their specific rejection reasons.
- Regenerating the corpus leaves the working tree clean, so all 13 vectors reproduce byte for byte.
- `ruff check`, `mypy` and `tools/check_dashes.py` clean.

I checked the §5.1 figures independently rather than reading them off the run. The accept record is 11,923 bytes compact against 1,184 without evidence, the quote is 10,667 base64url characters over 8,000 bytes, the declared structure ends at 4,935, and the remaining 3,065 bytes are confirmed zero. The 7,836 figure my previous approval left standing measures a record carrying only the declared structure, which is not what the corpus ships. Both numbers are arithmetically right and they measure different things, so stating all three for what they are is the correct fix rather than swapping one for the other.

On the model claim: `grade_model_claim` no longer reads the evidence at all, which is the second of the two ways I named on 4 September. REPORT_DATA is guest-controlled, so a recomputable match authenticates a commitment and not a measurement of the named weights, and `commitment-cannot-attest-model` holds that case on the unchanged genuine quote. The collateral pair closes the gap where rule 3 gained a clause that no vector exercised, and both records are re-signed so neither result can come from the envelope instead of the rule under test.

Two things I would take as follow-ups rather than conditions:

- `runtime-evidence.yml` lists `main`, `feat/**` and `fix/**` on its push trigger, so this branch is reached by the `pull_request` path filter but not by push.
- It refers to `actions/checkout@v7` and `actions/setup-python@v7` by tag, which matches `ci.yml` but not the SHA-pinned workflows in this repository.

---

StevenMih:
Late to this — it merged clean without me, and I'd rather engage with what's here than dwell on the timing.

The piece I can add from the other side of the seam: the capsule layer is the *consumer* of this grading. A record that cites a trust record needs exactly this `unattested / platform-attested / attested` distinction to carry a grade forward without laundering it — and it's why the non-propagation rule in §6.1 matters *more* outside TRACE than inside it. A citing record is where the temptation to round up lives: an attested platform measurement sitting next to a self-reported workload string, with nothing stopping a naive consumer from reading the second as though the first vouched for it. §6.1 is the rule that stops that, and it does its most important work in the format doing the citing. I've opened a worked example of that consumption as a PR on the AAC repo (action-state-group/agent-action-capsule#95 — open, precisely so it can take this kind of correction): a `ran_under` citation value that carries the cited record's grade only for what it measured, never for what the measured thing asserts. It's provisional, the name most of all, and your read on it is welcome.

On @altrudev's point that the reference implementation verifies against the record's own embedded key — that's the right thing to flag, and it's the same principle from the verifier side: a record that names its own key as trusted is not evidence, any more than a mutable reference is. The trust has to come from outside the record.

And the ask that would let the capsule side actually rely on this: the MUST-FAIL vectors that move `trace-trust-record` off provisional — a forged quote, a mismatched measurement, a record self-signing its own trust root. You'd mentioned you'd supply them; with the registry moratorium expired, they're the thing that lets a consumer treat a passing verify as meaning something.

altrudev:
I traced Steven's three requested MUST-FAIL cases against the current trees before adding anything new, and there are two separate layers here.

**1. The CPB `trace-trust-record` provisional reason appears stale against current TRACE main.**

The provisional CPB entry is pinned to `e0afe8e` and says `examples/canonicalization-boundary/` has four positive-only records, with promotion gated on negative counterparts.

That gap was subsequently closed by trace-spec #306 (merged): current main carries two schema-valid negative signing-preimage controls:

- `05-ascii-escaped-signature.json`
- `06-codepoint-order-signature.json`

Both carry genuine Ed25519 signatures over non-JCS alternate preimages and are required to fail TRACE verification specifically as `signature_invalid`; the tests also re-sign the same payload canonically as the positive control. So the exact "positive-only at this layer" reason recorded in the provisional entry no longer describes current TRACE.

**2. Two of the three runtime-evidence negatives Steven named already exist in the merged #277 corpus.**

Current `examples/runtime-evidence/vectors/` contains:

- `reject-forged-quote.json`
- `reject-measurement-mismatch.json`

So I do not think we should duplicate those.

The remaining case, "a record self-signing its own trust root", belongs to the external trust-context boundary, not to the runtime-evidence grade itself. The merged runtime profile now says this explicitly: the generator checks internal signature consistency under `cnf.jwk`, but does not promote that key into `trusted_root_keys`.

There is already a portable negative expressing the same trust property at the delegation verifier surface: `examples/delegation-link/08-root-key-untrusted.json`. Every record/signature/hash is internally consistent, but the root key is not in the verifier's externally supplied `trusted_root_keys`, and the required result is `provenance-invalid / root_key_untrusted`.

The remaining question for a capsule consumer is therefore narrower: whether it needs a **plain Trust Record trust-context fixture** carrying that same invariant, rather than borrowing the delegation fixture. I would keep that separate from the runtime-evidence grade so a valid `platform-attested` record is not incorrectly treated as defective merely because its signer is not trusted by a particular relying party.

My proposed next step is therefore:
1. update/check the CPB provisional pin against #306 rather than create more canonicalization negatives;
2. reuse the existing forged-quote and measurement-mismatch runtime vectors;
3. add only the missing portable external-trust negative if the consumer needs a plain-record fixture rather than the existing delegation case.

That keeps signature integrity, runtime-evidence grading, and signer trust as three distinct claims rather than letting one stand in for another.

altrudev:
Follow-up with the exact promotion evidence pin Steven asked for on action-state-group/scitt-payload-binding#67:

TRACE #306 is merged. Stable merge commit:

`7d754cde293ebf11ac6a5bb26edbed5751440f55`

That commit extends `examples/canonicalization-boundary/` from the four positive controls referenced by the provisional CPB entry to a two-sided set by adding:

- `05-ascii-escaped-signature.json`
- `06-codepoint-order-signature.json`

Both are complete schema-valid Trust Records with genuine Ed25519 signatures over deliberately non-JCS alternate preimages, and both are required to fail TRACE verification as `signature_invalid`.

The tests also verify the alternate signatures cryptographically over their declared bytes, prove those bytes differ from the RFC 8785 preimage, and canonically re-sign the same unsigned payload as a positive control.

Pinned blobs at that merged commit:

- `05-ascii-escaped-signature.json`: `9a1c2c21a5ab6fd5c0bc7b036c8a4e8127c0172f`
- `06-codepoint-order-signature.json`: `9951be1a72d377af46320592ef262f56daf401ef`
- `tests/test_canonicalization_boundary.py`: `242bb2bb6c0401b78702d03e6e67e2f4fcb8261b`

So the specific Rung-3 reason recorded in the CPB provisional entry ("positive-only at this layer; MUST-FAIL counterparts are in preparation") has been satisfied in the owner repository.

I am not asserting registry promotion on the owner's behalf, only supplying the exact merged evidence pin requested in the original CPB thread. The separate `json-sk-ascii` provisional algorithm row may have its own promotion gate and should remain a separate decision.

StevenMih:
Checked the pin before answering. 7d754cd resolves, and 05-ascii-escaped-signature.json and 06-codepoint-order-signature.json are there at exactly the blobs you gave — 9a1c2c21… and 9951be1a…. Thank you for pinning to the blob rather than the path; it made this a two-minute check instead of a conversation.

On the three cases: you're right on all three counts, and the third is a correction rather than a scoping difference, so let me take it properly. "A record self-signing its own trust root" isn't a runtime-evidence negative — it's signer trust, and putting it in the grade would make a sound platform-attested record look defective because one relying party doesn't accept its signer. That's the line @imran-siddique drew in this thread already ("grading is not authorization here"), and it's the same one our own verifier enforces from the other side: a verifying signature renders skip, never pass, because it establishes that bytes are unaltered and not who signed them. Signature integrity, runtime grade, signer trust — three claims, and you're right that none should stand in for another.

On the narrow question you left open: yes, a capsule consumer wants the plain Trust Record fixture rather than the delegation one, and the reason is shape. A capsule cites a Trust Record by digest, so the negative that matters to us is a Trust Record whose signer isn't in the verifier's externally supplied roots — 08-root-key-untrusted.json carries the right invariant on an object we don't cite. Not a promotion gate, just the fixture that would let a consumer test the boundary against the thing it actually holds.

On the provisional pin — @imran-siddique, this is yours rather than mine to close. Your #67 said you'd supply the commit when the negatives landed, and #306 appears to be it. Two questions before I move the entry: is 7d754cd the pin you'd supply, and given your note that a promoted entry should cite the governed specification location rather than an Opaque-controlled commit, is that URL available yet or do we promote against the commit and re-pin at the LF move?

altrudev:
@StevenMih I closed the third consumer-side boundary you asked for.

PR #341 adds the missing external-trust conformance vector. The corpus already had the other two cases:
- forged quote -> reject
- valid quote with mismatched runtime.measurement -> reject

The new vector proves the separate trust dimension: a record can verify its own envelope under embedded cnf.jwk and still remain outside the relying party's configured trust roots. In the vector, runtime evidence remains platform-attested while signer context is untrusted.

The exact head is e0fe814ee69a87e7c3207995ddcea3c7b512bba1. runtime-evidence (real pinned TDX verifier), CI, and CodeQL are all green.

PR: #341

AI-assisted analysis and implementation; I remain responsible for the contribution.

StevenMih:
That's the shape, and the part that makes it useful to a consumer is that both dimensions are held at once — platform-attested runtime evidence on a record whose signer is outside the verifier's roots. A citing record needs exactly that case to exist, because it's the one where the temptation to round up is strongest: a good grade sitting next to a key nobody vouched for. Saying it as a consumer rather than a reviewer — lywinged has the review.

It also pairs with the rule already in §6.1. That one says a grade doesn't extend to claims stated inside the record; this says it doesn't extend to whether you trust who signed it. Two directions, same point: the grade is narrower than it looks.

PR #278, opened 2026-09-03
test(provenance): reject digest under wrong algorithm label

solloek369-arch:
Follow-up to #243. In the approving review, @lywinged wrote the missing wrong-label regression control directly. This PR commits that control as an executable test and verifies with a permissive mutant that it is load-bearing.

This commits the exact wrong-label control from @lywinged's review: a record declaring `sha384:X` must not bind when the same hexadecimal value appears only under the subject's `sha256` label. It distinguishes the specific permissive `digest.values()`-style implementation that the previous tests did not catch.

That specific mutant passes the previous 48 tests and fails this new control. This is a bounded reviewer-loop closure, not a claim of complete algorithm/value/subject binding coverage.

This changes tests only. It does not change verifier behavior, schemas, normative text, or wire formats.

Testing:

- Python 3.11: 49 focused passed; 1168 full-suite passed, 1 skipped
- Python 3.12: 49 focused passed; 1168 full-suite passed, 1 skipped
- Ruff, mypy, dash policy, and `git diff --check`: pass

### DIFF
--- a/tests/test_build_provenance_depth_vectors.py
+++ b/tests/test_build_provenance_depth_vectors.py
@@ -297,6 +297,25 @@ def test_sha384_subject_binding_rejects_the_wrong_digest() -> None:
     }
 
 
+def _mislabelled_control() -> dict[str, Any]:
+    """Record declares sha384:X; the subject carries the same hex under `sha256`."""
+    vector = copy.deepcopy(CONTROL)
+    hexadecimal = "a" * 96
+    digest = "sha384:" + hexadecimal
+    vector["build_provenance"]["digest"] = digest
+    vector["context"]["artifact_digest"] = digest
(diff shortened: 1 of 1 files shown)


lywinged:
@imran-siddique this one needs your review rather than mine. The test is the one I wrote in the #243 review, so my approving it would not be a second pair of eyes.

imran-siddique (approved):
Ran the focused provenance suite on 413eddb: 49 passed. Replacing algorithm-specific digest lookup with the permissive digest.values() lookup leaves the previous 48 cases passing and makes exactly this new wrong-label control fail. Restored the original test file afterward. This is a signed-off, test-only change; the current Python and CodeQL checks have passed and there are no unresolved review threads.

PR #282, opened 2026-09-04
fix(content-marking): reject boolean assertion version

### LINKED ISSUE #281: content-marking: boolean true is accepted as assertion version 1
## Problem

`verify_assertion()` compares the peer-supplied assertion version with:

```python
if data.get("version") != ASSERTION_VERSION:
```

where `ASSERTION_VERSION = 1`.

In Python, `True == 1`, so a JSON assertion carrying:

```json
{"version": true}
```

is accepted as version 1 instead of being refused as an invalid version primitive.

The companion content-marking specification defines the assertion with `"version": 1` and says an unknown version is rejected rather than parsed best-effort. The current verifier therefore admits a boolean value as the known version because of host-language equality semantics.

## Reproduction

Starting from a valid assertion produced by `build_assertion()`, replace only:

```python
assertion["data"]["version"] = True
```

`verify_assertion(assertion, record_bytes)` succeeds on current `main`.

The existing version regression covers `version = 2`, but not the Python `bool`/`int` boundary.

## Proposed change

Establish the version primitive before accepting version 1, explicitly excluding `bool`.

A minimal regression set should hold:

- `1` -> accepted;
- `true` -> refused;
- `false` -> refused;
- `"1"` -> refused;
- `null` -> refused;
- another integer such as `2` -> refused.

This issue deliberately does not make a claim about floating-point JSON numbers; the defect is the unambiguous Python `bool`/`int` equivalence.

## Scope

Consumer-side validation only. No wire-format, schema, cryptographic, downgrade, or C2PA change is proposed.

altrudev:
Closes #281.

## What

`content_marking.verify_assertion()` compared the peer-supplied assertion version directly with integer `1`:

```python
if data.get("version") != ASSERTION_VERSION:
```

Because Python treats `True == 1`, a JSON assertion carrying `"version": true` was accepted as version 1.

This change establishes the boolean boundary explicitly before the existing version comparison:

```python
version = data.get("version")
if isinstance(version, bool) or version != ASSERTION_VERSION:
```

The change is deliberately narrower than a general JSON-number rule. It does not make a claim about floating-point representations; it closes the unambiguous Python `bool`/`int` alias.

## Regression coverage

Focused coverage holds:

- integer `1` -> unchanged success;
- boolean `true` -> refusal;
- boolean `false` -> refusal;
- string `"1"` -> refusal;
- `null` -> refusal;
- integer `2` -> refusal.

A dedicated regression names the mutation property: restoring the old plain `!= 1` comparison makes boolean `true` alias the supported version again.

## Verification

The exact branch source and regression file were re-read after the writes. The predicate matrix was independently exercised for `1`, `true`, `false`, `"1"`, `null`, and `2`; only integer `1` is accepted by the new condition, while the previous condition accepts both `1` and `true`.

A repository-wide pytest/ruff/mypy run is not claimed here because the available execution environment for this contribution does not currently have the TRACE checkout/toolchain. No GitHub Actions run is claimed.

## Scope

Consumer-side primitive validation only. No wire-format, schema, cryptographic, downgrade, or C2PA behavior change.

AI-assistance disclosure: ChatGPT assisted with source triage, adversarial-case design, implementation drafting, and diff review. `altrudev` reviewed the bounded claim and remains responsible for the contribution.

### DIFF
--- a/src/agentrust_trace/content_marking.py
+++ b/src/agentrust_trace/content_marking.py
@@ -136,9 +136,10 @@ def verify_assertion(assertion: dict[str, Any], record_bytes: bytes) -> dict[str
     data = assertion.get("data")
     if not isinstance(data, dict):
         raise ContentMarkingError("assertion has no data object")
-    if data.get("version") != ASSERTION_VERSION:
+    version = data.get("version")
+    if isinstance(version, bool) or version != ASSERTION_VERSION:
         raise ContentMarkingError(
-            f"unknown assertion version {data.get('version')!r}; expected {ASSERTION_VERSION}. "
+            f"unknown assertion version {version!r}; expected {ASSERTION_VERSION}. "
             "An unknown version is rejected rather than parsed best-effort."
         )
 
--- a/tests/test_content_marking_version_boundary.py
+++ b/tests/test_content_marking_version_boundary.py
@@ -1,0 +1,46 @@
+from __future__ import annotations
+
+import json
+
+import pytest
+
+from agentrust_trace.content_marking import ContentMarkingError, build_assertion, verify_assertion
+
+URL = "https://registry.example/records/abc123.json"
+
+
+def _record_bytes() -> bytes:
+    return json.dumps(
+        {
+            "eat_profile": "tag:agentrust-io.com,2026:trace-v0.2",
+            "iat": 1760000000,
+            "subject": "spiffe://example.org/agent/image-bot",
+            "data_class": "public",
+        }
+    ).encode()
+
+
+def _assertion_with_version(version):
+    raw = _record_bytes()
+    assertion = build_assertion(raw, url=URL)
+    assertion["data"]["version"] = version
+    return assertion, raw
+
+
+def test_integer_version_one_is_accepted() -> None:
+    assertion, raw = _assertion_with_version(1)
+    assert verify_assertion(assertion, raw)["subject"] == "spiffe://example.org/agent/image-bot"
+
+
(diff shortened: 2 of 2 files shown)


lywinged (approved):
Approved on `37b8e2b` over `fc38496`. Reproduced on main: an assertion from `build_assertion()` with `data.version` set to `true` passes `verify_assertion()`. On this head it is refused and integer `1` still passes; the PR's own test file has 2 of 7 cases failing against main and none here. Suite 1,174 passed, 1 skipped; ruff, mypy and the dash check clean. One thing outside the claim, for a follow-up rather than this PR: `1.0` passes on main and on this head, since `1.0 == 1` too, and a serializer that writes JSON numbers as floats will produce it.

imran-siddique (commented):
Review/test follow-up on `37b8e2ba807c`.

The explicit boolean rejection closes Python's bool-as-int path for assertion versions. Earlier focused validation passed 7 cases; 2 failed against the old implementation. The full hosted CI rerun passed. No additional blocker found in this change.

Full local `python -m pytest -q` (Windows, Python 3.12, existing review environment): **5 failed, 1169 passed, 1 skipped in 10.45s**.

Contribution requirement still outstanding: these commits have no `Signed-off-by:` trailer: `6ee3154f`, `37b8e2ba`. Please have the respective commit authors add their own DCO sign-offs, including merge commits, then request review of the resulting head.

Fresh workflow evidence: [33877476135](https://github.com/agentrust-io/trace-spec/actions/runs/33877476135) (success).

PR #284, opened 2026-09-04
fix(content-marking): validate record.url external references

### LINKED ISSUE #283: content-marking: record.url accepts values outside the C2PA external-reference shape
## Problem

`spec/content-marking-v1.md` says the assertion's `record` object follows C2PA's `hashed-ext-uri-map`: a URI plus a hash of the referenced bytes.

The reference implementation currently treats `record.url` as a truthiness check only.

Producer:

```python
if not url:
```

Consumer:

```python
if not isinstance(ref, dict) or not ref.get("url"):
```

That means truthy non-string values can be emitted or accepted as the record URL, and arbitrary non-HTTP strings are also accepted. Examples include `True`, `1`, `[1]`, `{"x": 1}`, `"not a url"`, and `"ftp://example.test/record.json"`.

This is a boundary mismatch in both directions: `build_assertion()` can produce a value outside the external-reference shape the companion spec says it implements, and `verify_assertion()` can accept the same shape from a peer assertion.

## Proposed change

Establish `record.url` as an absolute HTTP(S) URI at both producer and consumer boundaries before it is used in the assertion or in verifier diagnostics.

A focused regression matrix should hold:

- `https://registry.example/record.json` -> accepted;
- `http://registry.example/record.json` -> accepted;
- boolean/integer/list/object values -> refused;
- whitespace or arbitrary text -> refused;
- `ftp://...` -> refused;
- HTTP(S) URI without a host -> refused.

The implementation should avoid inventing a broader URI policy than the C2PA external-reference contract requires.

## Scope

Content-marking producer/consumer validation only. No change to hashing, Trust Record verification, C2PA signature handling, or TRACE wire format.

altrudev:
Closes #283.

## What

`spec/content-marking-v1.md` says the assertion's `record` object follows C2PA's external-reference shape, but the reference implementation previously treated `record.url` as truthiness only in both producer and consumer paths.

That allowed truthy non-string values and malformed/non-HTTP references to be emitted or accepted.

This change adds one shared `_record_url()` boundary and uses it from both `build_assertion()` and `verify_assertion()`.

The boundary requires:

- a string value;
- no whitespace;
- an RFC 3986-valid URI;
- `http` or `https` scheme;
- a non-empty authority/host component;
- a syntactically valid port in the 0-65535 range when present.

It deliberately does not add broader URL policy beyond the external-reference boundary.

## Regression coverage

Positive controls:

- absolute HTTPS URL;
- absolute HTTP URL.

Refusals on both producer and verifier paths:

- boolean;
- integer;
- list;
- object;
- whitespace-only string;
- arbitrary text;
- FTP URL;
- HTTP(S) URL with no authority;
- malformed authority with no hostname;
- non-numeric port.

A separate mutation-oriented test records the original failure class: several malformed values are truthy, so restoring the old `if not url` / `if not ref.get("url")` checks would reopen the boundary.

## DDC Radial review

The first candidate used `urllib.parse.urlsplit()`. The repository's offline invariant deliberately forbids importing network-adjacent standard-library modules, so CI correctly rejected that approach even though the focused URL tests passed. The implementation was replaced with the repository's already-installed `rfc3986-validator` plus a narrow authority/port boundary.

A second CI pass then caught that `rfc3986-validator` does not ship typing metadata. The final import carries a narrowly scoped `# type: ignore[import-untyped]`; mypy is not weakened globally.

The final exact head `e15651d9f6a23c04f2d07e111563b9e76e36bd6e` passed both repository CI and CodeQL (CI run 607, CodeQL run 407).

## Scope

Content-marking producer/consumer validation only. No change to hashing, Trust Record verification, C2PA signature handling, or TRACE wire format.

AI-assistance disclosure: ChatGPT assisted with source triage, adversarial-case design, implementation drafting, CI failure analysis, and diff review. `altrudev` reviewed the bounded claim and remains responsible for the contribution.

### DIFF
--- a/src/agentrust_trace/content_marking.py
+++ b/src/agentrust_trace/content_marking.py
@@ -21,7 +21,7 @@ import hashlib
 import re
 from typing import Any
 
-from rfc3986_validator import validate_rfc3986
+from rfc3986_validator import validate_rfc3986  # type: ignore[import-untyped]
 
 __all__ = [
     "ASSERTION_LABEL",
--- a/src/agentrust_trace/content_marking.py
+++ b/src/agentrust_trace/content_marking.py
@@ -20,6 +20,7 @@ from __future__ import annotations
 import hashlib
 import re
 from typing import Any
+from urllib.parse import urlsplit
 
 __all__ = [
     "ASSERTION_LABEL",
@@ -58,6 +59,19 @@ def _digest(data: bytes, alg: str) -> str:
     return f"{alg}:{_ALGS[alg](data).hexdigest()}"
 
 
+def _record_url(value: Any) -> str:
+    """Return a C2PA external-reference URL or refuse the malformed value."""
+    if not isinstance(value, str) or not value or any(ch.isspace() for ch in value):
+        raise ContentMarkingError("record.url must be an absolute http(s) URI")
+    try:
+        parsed = urlsplit(value)
+    except ValueError as exc:
+        raise ContentMarkingError("record.url must be an absolute http(s) URI") from exc
+    if parsed.scheme not in {"http", "https"} or not parsed.netloc:
+        raise ContentMarkingError("record.url must be an absolute http(s) URI")
+    return value
+
+
 def build_assertion(
     record_bytes: bytes,
     *,
@@ -77,8 +91,7 @@ def build_assertion(
             "record_bytes must be the serialized record as it will be served. A hash "
             "over a re-serialized object is a hash of bytes nobody will fetch."
         )
-    if not url:
-        raise ContentMarkingError("url is required: an assertion with no reference binds nothing")
+    url = _record_url(url)
 
     import json
 
@@ -143,8 +156,9 @@ def verify_assertion(assertion: dict[str, Any], record_bytes: bytes) -> dict[str
         )
 
     ref = data.get("record")
-    if not isinstance(ref, dict) or not ref.get("url"):
+    if not isinstance(ref, dict):
         raise ContentMarkingError("assertion carries no record reference")
+    url = _record_url(ref.get("url"))
     alg = ref.get("alg")
     if not isinstance(alg, str):
         raise ContentMarkingError(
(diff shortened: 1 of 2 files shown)


lywinged (approved):
Approved on `e15651d` over `fc38496`. Reproduced on main: `build_assertion(url=True)`, and `verify_assertion()` with `record.url` set to `true` or to an `ftp://` value, all pass. On this head all three are refused and the two positive controls pass; the PR's own test file has 20 of 23 cases failing against main and none here. Beyond the PR's matrix I put 24 more inputs through `_record_url()`: it accepts the RFC 3986 forms, upper-case scheme, userinfo, IPv6 literals, ports 0 to 65535, a trailing-dot host, and refuses port 65536, an empty port, whitespace and `%zz`. One behaviour the body does not name: a non-ASCII host such as `https://例子.测试/r` is refused while its punycode form is accepted. That is the right call for a C2PA URI, and it means a producer with an IDN host now has to encode it, so it belongs in the body. `rfc3986_validator` is already a declared dependency. Merges cleanly with #282; the combined tree passes 1,197. Suite 1,190 passed, 1 skipped; ruff, mypy and the dash check clean.

imran-siddique (commented):
Review/test follow-up on `e15651d9f6a2`.

The external-reference URL checks cover malformed values before dereference. Earlier focused validation passed 23 cases; 20 failed against the old implementation. The full hosted CI rerun passed. No additional blocker found in this change.

Full local `python -m pytest -q` (Windows, Python 3.12, existing review environment): **5 failed, 1185 passed, 1 skipped in 10.49s**.

Contribution requirement still outstanding: these commits have no `Signed-off-by:` trailer: `af3f4c1b`, `ef377a3d`, `80c14d7a`, `ace43b54`, `0a75dd60`, `8ab1c8c6`, `e15651d9`. Please have the respective commit authors add their own DCO sign-offs, including merge commits, then request review of the resulting head.

Fresh workflow evidence: [33880965950](https://github.com/agentrust-io/trace-spec/actions/runs/33880965950) (success).

PR #287, opened 2026-09-04
fix(provenance): require textual identity locators

### LINKED ISSUE #285: provenance: identity locator fields accept truthy non-string values
`spec/server-provenance-v1.md` defines the two identity locators as textual identifiers:

- `identity.artifact.package` is a Package URL;
- `identity.endpoint.url` is a URL.

The shared producer/consumer structural check currently tests both fields by truthiness only:

```python
if not artifact.get("package"):
    ...

if not endpoint.get("url"):
    ...
```

As a result, truthy non-string JSON values such as `true`, `1`, `[1]`, or `{"x": 1}` satisfy the locator-presence check. A signed peer provenance record can therefore verify with a non-text `artifact.package` or `endpoint.url` as long as the sibling digest is valid.

This is narrower than Package URL / URI syntax validation. The immediate boundary is only the JSON primitive type the specification names: both locator fields should be non-empty strings rather than arbitrary truthy values.

## Reproduction

Against current `main` at `fc38496f3ddacaa7ce922e1218b6fcd91735d751`, the exact predicates above accept each of:

```text
true
1
[1]
{"x": 1}
```

Existing coverage checks a missing endpoint key digest but does not exercise non-string locator values.

## Proposed bounded fix

Use one small non-empty-string boundary for `artifact.package` and `endpoint.url` in `_check_structure()`, exercised through both `build_record()` and `verify_record()`.

Do not expand this change into full Package URL or URL syntax policy; that can be decided separately if desired.

AI-assistance disclosure: ChatGPT assisted with source triage, adversarial-case design, duplicate search, and issue drafting. `altrudev` reviewed the bounded claim and remains responsible for the contribution.

altrudev:
Closes #285.

## What

`spec/server-provenance-v1.md` defines `identity.artifact.package` as a Package URL and `identity.endpoint.url` as a URL. The shared producer/consumer structural check previously used truthiness only, so truthy non-string JSON values could satisfy both locator-presence checks.

This change adds one small non-empty-string boundary and applies it to both locator fields from `_check_structure()`, which is shared by `build_record()` and `verify_record()`.

## Regression coverage

For both producer and verifier paths, the tests refuse:

- `true`;
- integer `1`;
- array `[1]`;
- object `{\"x\": 1}`.

Positive controls keep ordinary textual Package URL and HTTPS locator values working.

## Scope

This is intentionally only the primitive/type boundary. It does not add full Package URL parsing or broader URL syntax policy.

AI-assistance disclosure: ChatGPT assisted with source triage, adversarial-case design, implementation drafting, and diff review. `altrudev` reviewed the bounded claim and remains responsible for the contribution.

### DIFF
--- a/src/agentrust_trace/provenance.py
+++ b/src/agentrust_trace/provenance.py
@@ -66,6 +66,13 @@ def _as_object(value: Any, field: str) -> dict[str, Any]:
     return value
 
 
+def _nonempty_string(value: Any, field: str) -> str:
+    """Return a required textual locator, rejecting truthy non-string JSON values."""
+    if not isinstance(value, str) or not value:
+        raise ProvenanceError(f"{field} must be a non-empty string")
+    return value
+
+
 def _tool_count(catalog: dict[str, Any]) -> int:
     """Return the required catalog count as a JSON integer."""
     value = catalog.get("tool_count")
@@ -153,8 +160,7 @@ def _check_structure(
             raise ProvenanceError(
                 f"identity.artifact must be an object, got {type(artifact).__name__}"
             )
-        if not artifact.get("package"):
-            raise ProvenanceError("artifact.package is required (a Package URL)")
+        _nonempty_string(artifact.get("package"), "artifact.package")
         if not _DIGEST_RE.match(str(artifact.get("digest", ""))):
             raise ProvenanceError(
                 "artifact.digest must be a sha256: digest of the entrypoint. For an "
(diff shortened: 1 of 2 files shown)


lywinged (approved):
Approved on `c943434` over `fc38496`. Reproduced on main: `_check_structure()` tests `artifact.package` and `endpoint.url` by truthiness, so `true`, `1`, `[1]` and `{"x": 1}` satisfy the presence check. On this head each is refused through both `build_record()` and `verify_record()` and textual locators pass; the PR's own test file has 16 of 17 cases failing against main and none here. Suite 1,184 passed, 1 skipped; ruff, mypy and the dash check clean.

imran-siddique (commented):
Review/test follow-up on `c9434342c494`.

Identity locators now require textual values rather than accepting truthy objects or numbers. Earlier focused validation passed 17 cases; 16 failed against the old implementation. The full hosted CI rerun passed. No additional blocker found in this change.

Full local `python -m pytest -q` (Windows, Python 3.12, existing review environment): **5 failed, 1179 passed, 1 skipped in 10.49s**.

Contribution requirement still outstanding: these commits have no `Signed-off-by:` trailer: `cd5343d3`, `c9434342`. Please have the respective commit authors add their own DCO sign-offs, including merge commits, then request review of the resulting head.

Fresh workflow evidence: [33886018121](https://github.com/agentrust-io/trace-spec/actions/runs/33886018121) (success).

PR #288, opened 2026-09-04
fix(revocation): validate bundle freshness policy inputs

### LINKED ISSUE #286: revocation bundle policy inputs accept boolean values as seconds
The revocation-bundle freshness path still has two public-boundary gaps that are separate from the existing freshness work in #239.

`sign.verify_record()` already rejects boolean `now`, but its **bundle-specific** `max_bundle_age_seconds` input is only checked with `< 0`:

```python
if max_bundle_age_seconds < 0:
    ...
```

In Python, `bool` is an `int` subclass, so `True` is acted on as `1` second and `False` as `0` seconds.

The lower-level public `revocation.check_bundle()` accepts `now`, `max_bundle_age_seconds`, and `max_future_skew_seconds` and immediately uses them in freshness arithmetic without runtime primitive checks. The same boolean-as-0/1 behavior therefore exists when callers use that public entry point directly.

These inputs change the revocation result rather than merely formatting an error: a boolean configuration can make a bundle appear fresh, expired, or future-dated under a 0/1-second policy instead of being refused as malformed configuration.

## Ownership boundary

PR #239 already owns **top-level Trust Record freshness** validation for `sign.verify_record(max_age_seconds=...)` and `sign.verify_record(max_future_skew_seconds=...)`. This issue does not duplicate or modify that lane.

The bounded surface here is only:

- `sign.verify_record(max_bundle_age_seconds=...)`;
- `revocation.check_bundle(now=...)`;
- `revocation.check_bundle(max_bundle_age_seconds=...)`;
- `revocation.check_bundle(max_future_skew_seconds=...)`.

## Reproduction

Against current `main` at `fc38496f3ddacaa7ce922e1218b6fcd91735d751`:

```text
sign.verify_record(..., max_bundle_age_seconds=True)  -> treated as 1
sign.verify_record(..., max_bundle_age_seconds=False) -> treated as 0

revocation.check_bundle(..., now=True, ...)                    -> treated as 1
revocation.check_bundle(..., max_bundle_age_seconds=True, ...) -> treated as 1
revocation.check_bundle(..., max_future_skew_seconds=False, ...) -> treated as 0
```

The repository's public-function sweep varies `check_bundle`'s first positional `bundle` argument while holding these keyword policy arguments valid, so it does not exercise this surface.

## Proposed bounded fix

Hold the bundle-freshness policy inputs to integer seconds before they are acted on, explicitly excluding `bool`. Preserve all existing defaults and freshness semantics. Do not absorb #239's top-level `max_age_seconds` / `max_future_skew_seconds` work.

AI-assistance disclosure: ChatGPT assisted with source triage, adversarial-case design, duplicate/ownership review, and issue drafting. `altrudev` reviewed the bounded claim and remains responsible for the contribution.

altrudev:
Closes #286.

## What

The public revocation-bundle consumer used `now`, `max_bundle_age_seconds`, and `max_future_skew_seconds` directly in freshness arithmetic without holding them to the documented integer-seconds boundary.

In Python, booleans are integer subclasses, so `True`/`False` were acted on as `1`/`0` seconds. Because `sign.verify_record()` delegates bundle freshness to `check_bundle()`, the same boundary also affects its bundle-specific `max_bundle_age_seconds` when a bundle is actually checked.

This change adds one local `_check_seconds()` boundary in `revocation.py` and applies it before bundle shape or freshness evaluation.

## Ownership boundary

PR #239 already owns top-level Trust Record freshness validation for `sign.verify_record(max_age_seconds=...)` and `sign.verify_record(max_future_skew_seconds=...)`. This PR does not modify that lane.

## Regression coverage

The focused tests refuse `True` and `False` for:

- `revocation.check_bundle(now=...)`;
- `revocation.check_bundle(max_bundle_age_seconds=...)`;
- `revocation.check_bundle(max_future_skew_seconds=...)`;
- `sign.verify_record(max_bundle_age_seconds=...)` when a revocation bundle is checked.

Integer controls show the calls proceed to the existing bundle-shape outcome instead of being rejected by the new boundary.

No floating-point/double cases are part of this claim.

## Scope

No change to defaults, bundle schema, freshness arithmetic, signature verification, or the three revocation outcome states.

AI-assistance disclosure: ChatGPT assisted with source triage, adversarial-case design, duplicate/ownership review, implementation drafting, and diff review. `altrudev` reviewed the bounded claim and remains responsible for the contribution.

### DIFF
--- a/src/agentrust_trace/revocation.py
+++ b/src/agentrust_trace/revocation.py
@@ -121,6 +121,14 @@ def _unverified(cause: Cause, evidence: dict[str, Any]) -> RevocationCheck:
     return RevocationCheck(outcome="unverified_for_revocation", cause=cause, evidence=evidence)
 
 
+def _check_seconds(name: str, value: Any) -> None:
+    """Reject a malformed bundle-freshness input before using it as seconds."""
+    if isinstance(value, bool) or not isinstance(value, int):
+        raise ValueError(f"{name} must be an integer, got {type(value).__name__}")
+    if value < 0:
+        raise ValueError(f"{name} must be non-negative, got {value}")
+
+
 @lru_cache(maxsize=1)
 def _bundle_validator() -> jsonschema.Draft202012Validator:
     """A validator for the bundle schema that resolves its statement ``$ref`` locally.
@@ -198,6 +206,9 @@ def check_bundle(
     key. That is evidence failing rather than evidence absent, and it fails closed
     like the ``revocation`` store does.
     """
+    _check_seconds("now", now)
+    _check_seconds("max_bundle_age_seconds", max_bundle_age_seconds)
+    _check_seconds("max_future_skew_seconds", max_future_skew_seconds)
     trusted_ids = list(trusted_key_identifiers)
     trusted_bundle_keys = list(trusted_bundle_keys)
 
--- a/tests/test_revocation_policy_types.py
+++ b/tests/test_revocation_policy_types.py
@@ -1,0 +1,105 @@
+"""Regression tests for revocation-bundle freshness policy primitive boundaries."""
+
+from __future__ import annotations
+
+import time
+
+import pytest
+
+from agentrust_trace import generate_key, key_to_jwk, sign_record, verify_record
+from agentrust_trace.revocation import check_bundle
+
+
+def _record() -> dict:
+    return {
+        "eat_profile": "tag:agentrust-io.com,2026:trace-v0.2",
+        "iat": int(time.time()),
+        "subject": "did:mesh:spiffe://factory.example/agent/payments/prod",
+        "model": {"provider": "anthropic", "model_id": "claude-sonnet-4-6"},
+        "runtime": {
+            "platform": "software-only",
+            "measurement": "sha256:" + "0" * 64,
+        },
+        "policy": {
+            "bundle_hash": "sha256:" + "a" * 64,
+            "enforcement_mode": "enforce",
+        },
+        "data_class": "confidential",
+        "build_provenance": {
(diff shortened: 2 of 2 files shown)


lywinged (approved):
Approved on `8de5b5b` over `fc38496`. Reproduced on main: `check_bundle()` with `max_bundle_age_seconds=True` and `max_future_skew_seconds=False` runs through to an outcome. On this head each of the three seconds inputs refuses `True` and `False` with a `ValueError` naming the argument, and integers proceed to the bundle-shape outcome as before; the PR's own test file has 8 of 10 cases failing against main and none here. `sign.verify_record()` already refuses a boolean `now`, and `max_age_seconds` stays with #239 as the body says. Suite 1,177 passed, 1 skipped; ruff, mypy and the dash check clean.

imran-siddique (commented):
Review/test follow-up on `8de5b5b3efb4`.

The freshness-policy checks reject malformed policy values before arithmetic or bundle evaluation. Earlier focused validation passed 10 cases; 8 failed against the old implementation. The full hosted CI rerun passed. No additional blocker found in this change.

Full local `python -m pytest -q` (Windows, Python 3.12, existing review environment): **5 failed, 1172 passed, 1 skipped in 10.98s**.

Contribution requirement still outstanding: these commits have no `Signed-off-by:` trailer: `85b715db`, `8de5b5b3`. Please have the respective commit authors add their own DCO sign-offs, including merge commits, then request review of the resulting head.

Fresh workflow evidence: [33886941089](https://github.com/agentrust-io/trace-spec/actions/runs/33886941089) (success).

PR #293, opened 2026-09-05
docs: run the checks CI runs, and what a green one does not prove

lywinged:
## What this changes

`CONTRIBUTING.md` only. The command block under "Running the reference-library checks" now lists the checks CI runs, and a paragraph after it says what a passing check does not establish.

Two of CI's steps were not reachable from this file. `ci.yml` runs `ruff check src tests scripts`; the file said `ruff check src tests`. `ci.yml` runs `python tools/check_dashes.py` as a step of its own; the file did not mention it. A contributor following this file meets both for the first time as a red build. The five commands now match CI one for one, except that CI adds coverage flags to `pytest`; the block keeps the bare form it already carried.

The paragraph is there because of what a red build invites. A ban on a character cannot see the shape of its replacement: take an en dash out of a range, put a bare `to` in its place, and the checker passes on `1to3` where `1 to 3` was meant. The general form is that the check you just fixed is the one instrument guaranteed not to see what your fix introduced, so it cannot be the evidence that your fix is right. `tools/check_dashes.py` already prints the intended form for each character it bans, which is advice rather than an instrument, and nothing verifies that the repair used it.

## Type of change

## Spec section

None. No normative text, no schema change.

## Checklist

- [ ] `CHANGELOG.md` updated (for any normative change): not applicable, and recent docs-only changes to this file carry no entry
- [ ] Breaking changes marked with `<!-- CHANGED: #NNN: description -->` in spec text: not applicable
- [ ] Backward compatibility statement included (for breaking changes): not applicable

Tool-assisted: the comparison against `ci.yml` and this write-up.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,7 +9,8 @@ Install the development dependencies and run the suite from the repository root:
 ```bash
 pip install -e ".[dev]"
 pytest
-ruff check src tests
+ruff check src tests scripts
+python tools/check_dashes.py
 mypy src/agentrust_trace
 ```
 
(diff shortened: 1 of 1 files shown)


imran-siddique (approved):
Checked the full CONTRIBUTING.md diff against the current CI workflow: the added scripts path in Ruff and the dash-check command match the workflow, while pytest keeps its documented local form without CI's coverage flags. This is an editorial contribution with a signed-off commit. The current head has green Python 3.11/3.12 and CodeQL results and no unresolved review threads.

PR #300, opened 2026-09-06
fix(provenance): refuse conflicting tool schema aliases

### LINKED ISSUE #297: tool_catalog_hash silently prefers input_schema over inputSchema when both are present, letting a decoy schema pass check_tool_catalog
What is wrong

tool_catalog_hash() resolves each tool's schema with a silent fallback (provenance.py:117):
`"input_schema": t.get("input_schema", t.get("inputSchema")),`

When a tool object carries only one of input_schema / inputSchema this is fine — the docstring at line 90 and test_camel_case_input_schema_is_accepted explain the fallback exists because real MCP servers emit the field as inputSchema on the wire. But when a tool object carries both keys and they disagree, input_schema silently wins and inputSchema is never hashed. Nothing constrains or detects this case.

That's exploitable: a malicious or compromised MCP server can serve a tool with a decoy input_schema matching whatever was signed into the provenance record, and a real inputSchema — the field an actual MCP client validates tool calls against — that is different, and can be strictly wider (e.g. "additionalProperties": true where the record says false). check_tool_catalog() (provenance.py:419, docstring at line 422: "the only step that catches a live attack", Specification §5 step 5) hashes the decoy field, reports a clean match, and the real enforced schema diverges from the record undetected.

PoC (run against a fresh checkout, pip install -e .):

python

```
from agentrust_trace.provenance import tool_catalog_hash, check_tool_catalog

honest_tool = {
    "name": "read_file",
    "description": "Read a file from the sandboxed workspace",
    "input_schema": {"type": "object", "properties": {"path": {"type": "string"}},
                      "additionalProperties": False},
}
record = {"tool_catalog": {"hash": tool_catalog_hash([honest_tool]), "tool_count": 1}}

evil_tool_on_wire = {
    "name": "read_file",
    "description": "Read a file from the sandboxed workspace",
    "input_schema": honest_tool["input_schema"],          # decoy, matches the signed record
    "inputSchema": {"type": "object", "properties": {"path": {"type": "string"}},
                     "additionalProperties": True},        # what a real client enforces
}

check_tool_catalog(record, [evil_tool_on_wire])  # passes silently — no ToolCatalogMismatch
```

I checked git blame on line 117 (traces to the original feature PR, #140) and searched all issues/PRs for inputSchema before filing — nothing anywhere addresses the dual-key-disagreement case, so this isn't a known or already-discussed gap.

Spec section or file

spec/server-provenance-v1.md §4 (tool catalog hash) — only ever mentions input_schema, defines no inputSchema precedence rule. src/agentrust_trace/provenance.py:117 (the fallback), #/tool_catalog_hash and #/check_tool_catalog (:87-126, :419-422).

Expected behavior

Either:

tool_catalog_hash() should raise (ProvenanceError) on a tool object carrying both input_schema and inputSchema with differing values, since there's no way to represent "the schema" of such a tool unambiguously; or
The spec should define, normatively, which key wins when both are present, and the implementation should hash that one consistently. Right now the precedence at provenance.py:117 is an implementation detail with no basis in server-provenance-v1.md.

Impact

Yes — verifiers accept a provenance record as matching the live tool catalog (check_tool_catalog returns cleanly, no ToolCatalogMismatch) while the schema an actual MCP client enforces at call time is different from, and can be strictly wider than, the one the record attests to. This defeats the specific purpose §5 step 5 exists for: catching a server that isn't the server the record describes.

altrudev:
Closes #297.

## What

`tool_catalog_hash()` accepts both the SDK spelling `input_schema` and the MCP wire spelling `inputSchema`. The compatibility fallback previously preferred `input_schema` silently when both were present, even when the two values disagreed.

That left the live-catalog check vulnerable to an ambiguous tool object: a server could carry a snake-case schema matching the signed provenance record while also exposing a different camel-case schema.

This change preserves the existing compatibility rule for either spelling and for equal dual-key values, but refuses a tool object when both keys are present with different values.

## Regression coverage

Focused tests hold three cases:

- equal `input_schema` / `inputSchema` values hash identically to the existing single-key form;
- conflicting values raise `ProvenanceError` before hashing;
- `check_tool_catalog()` refuses the decoy-schema construction from #297 instead of reporting a clean match.

## Scope

No schema change, wire-format change, canonicalization change, signature change, or tool-count semantic change.

Credit: the issue and proof of concept were reported by @harshnair75567-cloud.

AI-assistance disclosure: ChatGPT assisted with source triage, duplicate/ownership review, implementation drafting, regression design, and exact-diff review. `altrudev` reviewed the bounded claim and remains responsible for the contribution.

### DIFF
--- a/src/agentrust_trace/provenance.py
+++ b/src/agentrust_trace/provenance.py
@@ -112,7 +112,7 @@ def tool_catalog_hash(tools: list[dict[str, Any]]) -> str:
         if (
             "input_schema" in t
             and "inputSchema" in t
-            and t["input_schema"] != t["inputSchema"]
+            and anchor_bytes(t["input_schema"]) != anchor_bytes(t["inputSchema"])
         ):
             raise ProvenanceError(
                 f"tools[{index}] carries conflicting input_schema and inputSchema values"
--- a/src/agentrust_trace/provenance.py
+++ b/src/agentrust_trace/provenance.py
@@ -109,6 +109,14 @@ def tool_catalog_hash(tools: list[dict[str, Any]]) -> str:
     for index, t in enumerate(tools):
         if not isinstance(t, dict):
             raise ProvenanceError(f"tools[{index}] must be an object, got {type(t).__name__}")
+        if (
+            "input_schema" in t
+            and "inputSchema" in t
+            and t["input_schema"] != t["inputSchema"]
+        ):
+            raise ProvenanceError(
+                f"tools[{index}] carries conflicting input_schema and inputSchema values"
+            )
     normalized = sorted(
         (
             {
--- a/tests/test_alias_canonical_comparison.py
+++ b/tests/test_alias_canonical_comparison.py
@@ -1,0 +1,22 @@
+"""Alias equality follows the bytes hashed, including JSON type distinctions."""
+import pytest
+
+from agentrust_trace.provenance import ProvenanceError, check_tool_catalog, tool_catalog_hash
+
+
+@pytest.mark.parametrize("left,right", [
+    ({"const": True}, {"const": 1}),
+    ({"const": False}, {"const": 0}),
+    ({"properties": {"x": {"enum": [1]}}}, {"properties": {"x": {"enum": [True]}}}),
+])
+def test_python_equal_but_distinct_json_aliases_are_refused(left, right):
+    signed = {"name": "read", "description": "Read", "input_schema": left}
+    record = {"tool_catalog": {"hash": tool_catalog_hash([signed]), "tool_count": 1}}
+    with pytest.raises(ProvenanceError, match="conflicting input_schema and inputSchema"):
+        check_tool_catalog(record, [{**signed, "inputSchema": right}])
+
+
+def test_alias_object_key_order_does_not_change_the_hash():
+    tool = {"name": "read", "input_schema": {"type": "object", "additionalProperties": False}}
+    dual = {**tool, "inputSchema": {"additionalProperties": False, "type": "object"}}
+    assert tool_catalog_hash([dual]) == tool_catalog_hash([tool])
--- a/tests/test_provenance_dual_input_schema.py
+++ b/tests/test_provenance_dual_input_schema.py
@@ -1,0 +1,56 @@
+"""Regression coverage for conflicting MCP tool schema aliases (#297)."""
+
+from __future__ import annotations
+
+import pytest
+
+from agentrust_trace.provenance import ProvenanceError, check_tool_catalog, tool_catalog_hash
+
+
+def _schema(*, additional_properties: bool) -> dict[str, object]:
+    return {
+        "type": "object",
+        "properties": {"path": {"type": "string"}},
+        "additionalProperties": additional_properties,
+    }
+
+
+def _tool(schema: dict[str, object]) -> dict[str, object]:
+    return {
+        "name": "read_file",
+        "description": "Read a file from the sandboxed workspace",
+        "input_schema": schema,
+    }
+
+
+def test_equal_schema_aliases_keep_the_existing_wire_compatibility() -> None:
+    schema = _schema(additional_properties=False)
+    snake = _tool(schema)
+    both = {**snake, "inputSchema": schema.copy()}
+
+    assert tool_catalog_hash([both]) == tool_catalog_hash([snake])
+
+
+def test_conflicting_schema_aliases_are_refused_before_hashing() -> None:
+    live = _tool(_schema(additional_properties=False))
+    live["inputSchema"] = _schema(additional_properties=True)
+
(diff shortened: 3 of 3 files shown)


lywinged (changes requested):
Request changes, for one line. The direction is right and the reported vector is closed; the comparison that closes it sits one layer above the bytes the hash covers, and that layer lets a pair through.

For anyone reading this later: `tool_catalog` is a field of the MCP Server Provenance Record in `spec/server-provenance-v1.md`, not of the Trust Record, and `validate_json` refuses a Trust Record that carries it. Section 4 of that specification names the preimage as name, description and `input_schema`, and says nothing about `inputSchema`; accepting the MCP wire spelling is this implementation's convenience, and the ambiguity this PR closes is the one that convenience opened. It is reachable in practice: cmcp's proxy (`mcp/proxy.py`) sends `tools/list` to the upstream server and hands the returned objects to `check_tool_catalog` unchanged, so both spellings are the server's to set, while cmcp's catalog loader (`catalog/loader.py`) resolves `inputSchema` first and this hash resolves `input_schema` first.

What I ran on `f5f184b`. The three new tests copied into a worktree of main: two fail, one passes, which is the right shape. Full suite on this head 1188 passed and 1 skipped against 1185 on main, with `ruff`, `tools/check_dashes.py` and `mypy` clean. Two spellings alone hash identically, so a single-spelling producer sees no change. cmcp itself emits only `inputSchema`, never both, and no file in the organisation's repositories carries both spellings on one object, so nothing existing trips the new refusal.

The gap. The check is `t["input_schema"] != t["inputSchema"]`, Python equality, and `True == 1` and `False == 0` in Python while `anchor_bytes` writes `true` and `1`. On this head, a signed side of `{"additionalProperties": 1}` with a live side of `{"additionalProperties": true}` passes the new check and `check_tool_catalog` accepts it; `0` against `false` and a nested `1` against `true` behave the same. Whether that widens anything depends on the validator: Python's jsonschema treats `1` as `true` and `0` as `false`, so nothing changes there; a validator that discards an invalid keyword value rather than failing falls back to the default, and for `additionalProperties` the default is permissive. Narrow, and the function exists for exactly the cross-implementation case.

The fix is one token: compare `anchor_bytes(t["input_schema"]) != anchor_bytes(t["inputSchema"])`. On a copy of this branch that change leaves the three new tests passing, refuses the `1` and `true` pair and the `0` and `false` pair, still accepts a pair that differs only in key order, and the full suite stays at 1188. A fourth test for the `1` and `true` pair would pin it.

Not this PR, for the maintainers. Section 4 names one spelling and the two repositories resolve the pair in opposite orders; a stated precedence in the specification, or in cmcp, is what closes that for good, which is the second remedy #297 names. And a simpler design is available: refuse any tool object that carries both keys, equal or not, since neither this specification nor MCP's own wire format gives a server a reason to send both.

imran-siddique (approved):
Reviewed 427952e2dee3646b44debb4dbaaef97bbea097ca. The requested comparison now uses anchor_bytes, the same representation the catalog hashes. Three independent cases demonstrate that Python-equal boolean/integer values, including nested values, were accepted before the repair and are refused afterward. Equivalent object key order still produces the same hash. All 112 focused provenance tests pass; changed-file Ruff and hosted CI pass. No held workflows were found.

This addresses the comparison finding on the earlier head. I pushed the repair, so independent re-review is still required before merge.

lywinged (approved):
Approving at `427952e`. The comparison now runs on the bytes the hash covers, and the new file pins it.

Measured on this head. The three pairs that passed at `f5f184b` are refused: `1` against `true`, `0` against `false`, and a nested `1` against `true`. A pair that differs only in object key order is still accepted, and a genuinely different pair is still refused, so the guard did not get coarser in the process. Full suite 1192 passed and 1 skipped, with `ruff`, `tools/check_dashes.py` and `mypy` clean.

Both test files are load-bearing, checked by putting the defect back. Restoring `!=` in place of the byte comparison turns all three parametrized cases in `test_alias_canonical_comparison.py` red and leaves the earlier file green, which is the split you would want: the new file exists for exactly what the old one could not see. Removing the guard entirely turns five red across the two files.

One behaviour did change, and it is a diagnostic rather than a decision. A tool object that carries both aliases, differing values, and a number the anchor profile refuses used to fail with the conflicting-aliases error and now fails with `UnanchorableValue` naming the number, because the comparison reaches the serializer first. The other three combinations of those inputs that I tried raise what they raised before, and both outcomes are refusals, so nothing is accepted that was refused before.

Against current main, `52687e6`, which now carries #239's change to the same module: merges with no conflict, and the merged tree runs 1217 passed and 1 skipped with the three checks clean.

The two notes from the earlier review are unchanged by this and still for the maintainers, not for this PR: a stated precedence for the two spellings, and the simpler design of refusing any object that carries both keys at all.

PR #301, opened 2026-09-06
ci: add actionlint and a test-environment guard

imran-siddique:
Two guardrails, each for a silent failure that actually happened during this security sweep.

## 1. actionlint

I introduced a duplicate `env:` key in cmcp's `release.yml` while hardening tag interpolation. GitHub Actions refuses to load a workflow with a duplicate key, reported it as a "workflow file issue" on unrelated pushes, ran nothing for the release event, and left a version unpublished for hours.

It got through because `yaml.safe_load` keeps the last value for a duplicate key without complaining. actionlint says it plainly:

```
.github/workflows/repro.yml:16:9: key "env" is duplicated in element of "steps" section.
previously defined at line:13,col:9 [syntax-check]
```

Verified against a reproduction of the exact broken file before adding it.

**On how it's installed:** fetched by pinned version with a checksum verification, not run as a third-party action. actionlint publishes no official action, and a check whose whole purpose is guarding the workflow supply chain should not add a new dependency to it.

Triggered only on changes under `.github/workflows`, so it costs nothing on ordinary PRs. **Every repo in the org is clean against it today**, so this adds no backlog.

## 2. Test-environment guard

pytest puts the source tree on the path, so an in-process import always finds this tree and looks right. A test that shells out gets no such help: `subprocess.run([sys.executable, ...])` resolves the distribution normally, and with a released wheel also installed it finds site-packages.

The subprocess then exercises a **published version** while the suite reports a pass. That is exactly how `trace-spec`'s tutorial test graded against the previous schema and still looked green, and why `cmcp`'s distribution smoke test reported a version that had nothing to do with the tree.

The guard probes a subprocess once per session and fails with an actionable message:

```
ERROR: agentrust_trace resolves to ...\site-packages\agentrust_trace\__init__.py
in a subprocess, outside C:\Users\imran\source\trace-spec. Any test that shells
out would exercise that installed distribution instead of this working tree.
Install editable (pip install -e .) or uninstall the shadowing agentrust_trace.
```

**It is a no-op in CI**, which installs editable, and verified as such against a correctly configured local venv. A package that is not importable from a subprocess at all is deliberately allowed: that is a path-only setup, not a shadowing install, so nobody gets locked out of running the suite.

Applied to the seven repos that install their own package editable in CI. `examples`, `integrations` and `demos` test against third-party distributions on purpose and get actionlint only.

https://claude.ai/code/session_01XbDBXDWWvMFa7c2jGgyq9t

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -1,0 +1,41 @@
+name: Workflow lint
+
+# actionlint catches what a YAML parser cannot: duplicate mapping keys that
+# make GitHub refuse to load a workflow, invalid ${{ }} expressions, and shell
+# problems inside run: blocks. A duplicate env: key silently broke this repo
+# family's release workflow once; yaml.safe_load keeps the last value without
+# complaining, so local validation passed while Actions rejected the file.
+on:
+  push:
+    branches: [main]
+    paths: ['.github/workflows/**']
+  pull_request:
+    paths: ['.github/workflows/**']
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  actionlint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      # Fetched and checksum-verified rather than run as a third-party action,
+      # so this check adds no new action to the supply chain it exists to guard.
+      - name: Install actionlint
+        env:
+          ACTIONLINT_VERSION: 1.7.12
+          ACTIONLINT_SHA256: 8aca8db96f1b94770f1b0d72b6dddcb1ebb8123cb3712530b08cc387b349a3d8
+        run: |
+          set -euo pipefail
+          archive="actionlint_${ACTIONLINT_VERSION}_linux_amd64.tar.gz"
+          curl -sSfL -o "$archive" \
+            "https://github.com/rhysd/actionlint/releases/download/v${ACTIONLINT_VERSION}/${archive}"
+          echo "${ACTIONLINT_SHA256}  ${archive}" | sha256sum -c -
+          tar -xzf "$archive" actionlint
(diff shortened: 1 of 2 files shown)


lywinged (changes requested):
Read the description and the diff at `c638267`, then ran both guards. Everything I ran checks out; one thing sits outside the diff.

The `actionlint` half: the 1.7.12 linux_amd64 archive from the release page hashes to the `ACTIONLINT_SHA256` in the workflow, the checkout pin `3d3c42e` is the `v7.0.1` tag, the same pin six of the eight existing workflows use (`contributor-check.yml` is on `v7.0.0`, the approval gate checks nothing out), and 1.7.12 run over all nine workflow files on this head, the new one included, exits 0. The `on:` block is what the description says: pushes to `main` and pull requests, both filtered to `.github/workflows/**`, plus `workflow_dispatch`. The reproduction holds too: a step with a duplicate `env:` key is reported as `[syntax-check]` with the earlier line named, and `yaml.safe_load` loads the same file without a word and keeps the second value. The one thing I did not reproduce is `install` into `/usr/local/bin` without `sudo` on the hosted runner; the `actionlint` check on this PR is green, which answers it.

The `conftest.py` half, exercised three ways: with an editable install the guard is a no-op and the suite runs, 1217 passed, 1 skipped, the same count as `main`; with the 0.10.0 wheel installed non-editable in a fresh venv, `pytest` stops before collection with the message in the description, naming the site-packages path and this tree; with nothing installed and only `pythonpath = ["src"]`, collection proceeds, which is the path-only case the docstring allows. Ruff and the format check are clean on `tests/conftest.py`, and running `tools/check_dashes.py` reports nothing.

The one thing: `c638267` carries no `Signed-off-by` trailer. CONTRIBUTING line 48: "PRs without DCO sign-off will not be merged." `git rev-list --count` from the merge base gives 1, so `git commit --amend -s` covers the whole branch; force-push that and I approve, nothing else needs to change.

PR #305, opened 2026-09-07
fix: reject a malformed signature field with the documented error, not a crash

rajnisht7:
## What this changes

`provenance.verify_record()` and `intent_bridge.verify_bridge()` both decode the `signature` field before checking it's actually usable.

- `provenance.verify_record()` never checked that `signature` was a string. A non-string value (an int, a list, a dict) crashed with a raw `TypeError` instead of the `ProvenanceError` this function documents.
- `intent_bridge.verify_bridge()` checked the type, but a correctly-typed string that wasn't valid base64 (too short, or non-ASCII) leaked a raw `ValueError` instead of `IntentBridgeError`.

Neither is a signature bypass a bad signature was always rejected, just with the wrong exception type, so any caller catching the documented error type would crash uncaught instead.

Fix: both now reuse `sign._b64url_decode()` (already `sign.verify_record()`'s own guard for this field) and wrap its error in the right type. The `provenance.py` fix is placed exactly where the old crash was, so a record with more than one problem still reports errors in the same order as before.

## Type of change

- [ ] Schema change
- [ ] Example addition

## Spec section

None

## Checklist


imran-siddique (approved):
Reviewed 6d72ff8686f1e6d7bab9970e5fd70a91a85cbd65.

The changes normalize malformed signature refusals into the documented module exceptions without changing successful signature verification or the cnf.jwk check order. All 144 focused provenance/intent-bridge tests pass. Restoring both original implementations made 12 new cases fail while the ordering control passed. Full local Windows suite: 1,300 passed, 1 skipped, and the same five previously documented fixture-newline/schema-path failures; hosted Linux CI is green. Non-blocking documentation correction: malformed base64 strings in provenance were already caught by the old broad exception handler and raised ProvenanceError. The old uncaught type errors were the non-string inputs before that handler. The new string cases normalize the message there; intent_bridge genuinely leaked ValueError.

PR #308, opened 2026-09-08
Remove FAQ structured data from README

harshnair75567-cloud:
Removed JSON-LD structured data for TRACE FAQ from README.

## What this changes

Removes a `<script type="application/ld+json">` FAQPage structured-data block from the bottom of README.md. This JSON-LD was intended for SEO on the docs site (trace.agentrust-io.com) and doesn't belong in a GitHub-rendered README — GitHub's markdown sanitizer doesn't render `<script>` tags cleanly, so the raw tag was showing up as visible text on the README page. No content, links, or spec text is changed — only the removal of this non-rendering script block. The FAQ markdown content itself and the Contributing section are untouched.

```Standards alignment
Hosted at the Linux Foundation as its own series, "TRACE Specification, a Series of LF Projects, LLC", under [LF Projects policies](https://lfprojects.org/policies/). The Linux Foundation [announced the contribution](https://www.linuxfoundation.org/press/linux-foundation-welcomes-trace-to-advance-verifiable-runtime-evidence-for-ai-workloads) on 25 August 2026, developed with AMD, Intel, Microsoft, OPAQUE and TII. Builds on [RFC 9711 (EAT)](https://www.rfc-editor.org/rfc/rfc9711), [RFC 9334 (RATS)](https://www.rfc-editor.org/rfc/rfc9334), and SCITT draft-22.

Frequently asked questions
What is TRACE?
TRACE (Trust, Runtime Attestation, and Compliance Evidence) is an open specification for hardware-attested AI agent governance records. It defines the record format, the anchoring protocol, and the verification rules for cryptographic evidence that an AI agent ran under a specific policy, in a verified hardware environment, on a given data class, invoking identified tools.

What does a TRACE Trust Record prove?
A single signed Trust Record answers, in a form any third party can verify without trusting the operator: what model ran, where it ran, under which policy, what data class it touched, which tools were called, and whether the record is independently anchored to a SCITT transparency ledger.

What standards is TRACE built on?
TRACE builds on open IETF and IRTF standards: RFC 9711 (CBOR Web Token / EAT) for the claim envelope, RFC 9334 (RATS) for the attester, verifier, and relying-party roles, and the SCITT draft for transparency-ledger anchoring. It is designed for CoSAI WS4 interoperability.

How do I create and verify a Trust Record?
Install the Python library with pip install agentrust-trace, sign a record with TrustRecord.sign(claims, signing_key), anchor it to a SCITT ledger with record.anchor(), and check it with record.verify(verifying_key).

[...]

PR #309, opened 2026-09-08
fix(readme): name the API the package actually has, and pin it

lywinged:
## What this changes

The FAQ answer for "How do I create and verify a Trust Record?" told readers to call `TrustRecord.sign(claims, signing_key)`, `record.anchor()` and `record.verify(verifying_key)`. None of the three exists, and none ever did: `TrustRecord` carries no `sign` or `verify`, nothing exported anchors anything, and `git log -S "def anchor(self"` over the package returns nothing. The same README already carried a working example thirty six lines earlier, a fenced block ending in `signed = sign_record(record, key)`, so the file contradicted itself.

`pyproject.toml` sets `readme = "README.md"`, so that sentence is also the PyPI long description. It has shipped in every release since 0.4.0 on 2026-07-21 and is on the project page for 0.10.0, which makes it the first instruction a reader gets after `pip install agentrust-trace`.

The replacement names `sign_record` and `verify_record`, both of which `tests/test_docs_quickstart.py` already executes out of `docs/quickstart.md`, and states the boundaries a reader would otherwise infer: `verify_record` checks revocation only when it is handed a revocation source, it performs no attestation verification, and the package does not anchor a record to a transparency ledger, `transparency` being a string member that carries the SCITT receipt URI once a ledger returns one.

`tests/test_readme_api_claims.py` keeps it from drifting back. Nothing checked what the README says before this: the quickstart test executes the fenced blocks under `docs/`, and the README had no instrument at all. The new file carries two, one per way the README states an API. The fenced block is executed, the way `tests/test_docs_quickstart.py` runs the pages under `docs/`. The inline spans are resolved by a deliberately narrow rule: a call written in an inline span has to name something that exists, and not one of the names every object carries, which would resolve for any receiver at all. Its negative control is this defect itself, so the check cannot quietly stop resolving anything, and when it fails it names the call and says what to write instead.

Three notes for review. The same sentence also sits inside the `application/ld+json` block that #308 removes; this PR leaves that copy alone rather than conflict with #308, which deletes it outright, and the two merge cleanly in either order. The guard reads inline code spans only, which is why it does not see that copy. And it is scoped to the README on purpose: pointed at the whole tree it would flag ten files that legitimately name `json.dumps(...)` and friends in prose, which would make it a style check rather than a fact check.

Checked on this branch: `pytest` 1310 passed and 1 skipped on 3.11 and on 3.12, `ruff check src tests scripts`, `python tools/check_dashes.py` and `mypy src/agentrust_trace` all clean, on a fresh clone with only `pip install -e ".[dev]"`. Rendering the new README with `readme_renderer`, which is what PyPI uses, introduces no escaped markup of its own.

## Type of change

- [ ] Schema change
- [ ] Example addition

## Spec section

None. `spec/trace-v0.2.md` is not touched.

## Checklist

- [ ] `CHANGELOG.md` updated (for any normative change): not a normative change

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ TRACE builds on open IETF and IRTF standards: RFC 9711 (CBOR Web Token / EAT) fo
 
 ### How do I create and verify a Trust Record?
 
-Install the Python library with `pip install agentrust-trace`, sign a record with `TrustRecord.sign(claims, signing_key)`, anchor it to a SCITT ledger with `record.anchor()`, and check it with `record.verify(verifying_key)`.
+Install the library with `pip install agentrust-trace`, sign a record with `sign_record(record, key)`, and check one with `verify_record(record, public_key_or_jwk=trusted_key)`, both imported from `agentrust_trace`. `verify_record` checks the profile URI, the schema, the signature over the RFC 8785 canonical form, and freshness; it checks revocation only when you pass it a revocation source, and it performs no attestation verification. The library does not anchor a record to a transparency ledger: `transparency` is a string member holding the SCITT receipt URI, which a producer fills in once its ledger returns one. The Quickstart page carries a runnable version of both steps.
 
 ### How does TRACE relate to AGT and cMCP?
 
--- a/tests/test_readme_api_claims.py
+++ b/tests/test_readme_api_claims.py
@@ -1,0 +1,157 @@
+"""Every API call the README names has to exist in the package.
+
+`pyproject.toml` sets `readme = "README.md"`, so this file is the PyPI long
+description: a call written here is the first instruction a new user reads, and
+it is the text an answer engine quotes back when asked how to use TRACE. Through
+0.10.0 the README told them to call `TrustRecord.sign(claims, signing_key)`,
+`record.anchor()` and `record.verify(verifying_key)`. None of the three has ever
+existed. The package signs with `sign_record`, checks with `verify_record`, and
+does not anchor anything.
+
+Nothing caught it. `tests/test_docs_quickstart.py` executes the fenced blocks in
+`docs/`, and the README states its API in inline spans inside prose, which no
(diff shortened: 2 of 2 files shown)


imran-siddique (approved):
Approving. I checked the three names against the package rather than against the README, and against the published artifact rather than the tree, because both halves of your claim are about things a reader meets outside this repository.

In a clean environment: `TrustRecord` exists and has no `sign`, no `verify` and no `anchor`. `sign_record` and `verify_record` are both exported. So the FAQ was naming an API that the package has never offered, sitting thirty-six lines below a fenced block that uses the real one.

The published side is worse than the PR states. `agentrust-trace` 0.10.0's long description on PyPI, fetched from the JSON API, contains `TrustRecord.sign(`, `record.anchor()` and `record.verify(` verbatim. That is the first instruction after `pip install agentrust-trace`, and `verify_record(` does not appear anywhere in it. I did not check every release back to 0.4.0, so treat that part of your body as unverified by me; 0.10.0 I confirmed.

The test is the part worth having. The reason nothing caught this is precise and you named it: `test_docs_quickstart.py` executes fenced blocks under `docs/`, and the README states its API in inline spans inside prose, which no instrument read. Resolving a qualified name attribute by attribute, and requiring an unqualified receiver's method to exist on some exported class, keeps it a fact check rather than a style check. That boundary is what makes it maintainable.

Two things for the record. This merges cleanly after #308, which removes the JSON-LD block from the same file; I ran the sequential merge rather than predicting it. And neither fix reaches a user until the next release, because PyPI serves the description uploaded with 0.10.0. The README is right on main and wrong on the project page until then, which is worth a release rather than waiting for one to happen for another reason.

`needs-review:UNKNOWN` on this PR is the contributor-check bot failing closed on a transient error. It carries no signal and should not colour anyone's read of this.

Issue #307, opened 2026-09-08
Express proof-of-possession for identity credentials using a non-JWK binding (e.g. DPoP)


trustedaicompute-ops:
**Spec section affected**
spec/trace-v0.2.md, the cnf / proof-of-possession requirement checked by TR-SIG-004 in agentrust-trace-tests.

**Problem**
We are integrating ComputeID (cryptographic identity for AI agents, hybrid RSA-SHA256 + ML-DSA-65 signing) as a TRACE external-evidence-source (see agentrust-io/integrations#176). Our mapped record fails Level 0 conformance on "cnf must contain jwk with kty" because our credentials do not bind a proof-of-possession key via a JWK in the cnf claim. ComputeID already has a working proof-of-possession mechanism, DPoP (RFC 9449), used elsewhere in our authorization flow. The current spec assumes JWK is the only representation, which does not accommodate credential systems that already have a different, working PoP mechanism.

**Proposed change**
Not yet a specific normative text proposal. Opening this as a discussion first: should cnf accept a DPoP-style proof-of-possession representation as an alternative to JWK/kty, or should proof-of-possession conformance at Level 0 be satisfiable by any spec-recognized mechanism rather than requiring the JWK form specifically?

**Backward compatibility**
- [x] Non-breaking (new optional field, informative addition)
- [ ] - [ ] Breaking (removes or changes required field, changes wire format, changes algorithm set)
**Motivation**
Closes a real conformance gap for identity-issuance systems that already implement proof-of-possession through a mechanism other than cnf/JWK, without requiring them to adopt a second, redundant PoP scheme solely to satisfy TRACE.

**Related issues or PRs**
agentrust-io/integrations#176 (ComputeID AgentPassport TRACE adapter, CONFORMANCE.md documents this as one of the real, current Level 0 failures)

lywinged:
Thanks for the write up, and for shipping an offline verifier and the fake-CA rejection tests alongside it. That made this straightforward to reason about.

[...]

PR #311, opened 2026-09-09
fix(schema): hold an RSA confirmation key to carrying its key material

lywinged:
## What this changes

The confirmation key says of itself that "Keys must carry actual key material: OKP keys require crv and x; EC keys require crv, x, and y", and the `allOf` behind that sentence carried branches for `OKP` and `EC` and for nothing else. So this validated:

```json
"cnf": { "jwk": { "kty": "RSA" } }
```

and the record then failed inside the verifier, where `sign.jwk_thumbprint` reports `jwk with kty='RSA' is missing required thumbprint member 'e'`, and in the conformance suite, where TR-SIG accepts `OKP` and `EC`.

Nothing was accepted that should have been refused: every path downstream fails closed. What was wrong is which instrument spoke. The schema is the artifact an implementation in any language validates against, and it was not the thing that told the producer the key was unusable. A producer working in another language saw a valid record and learned otherwise from a Python traceback, from a conformance run, or not at all.

`RSA` now requires `n` and `e`, declared as strings. Those declarations are load-bearing rather than decorative: without them a non-string modulus reaches `additionalProperties`, which admits any canonicalizable value, and a test covers that. The reasoning for the constraint sits in the `$comment` beside it rather than in `description`, which is the convention the private-member fix set in the same block. Both copies of the schema move together, the published `schema/trace-claim.json` and the packaged `src/agentrust_trace/schema/trace-v0.2.json`, and a test asserts they are the same bytes, as `test_provenance_cnf_boundary` already does for its own reason. The v0.1 schema in the package carries the same two branches and is left alone, on the grounds #296 left it alone: nothing loads it, and `test_the_superseded_schema_is_still_loaded_by_nothing` fails if anything starts to.

## What it tracks

A schema change has to track normative text rather than a description, so: section 3.2.2 requires the record to be bound by a signature made by the key in `cnf`, and section 3.3 rule 1 has a verifier check that binding before any other field is trusted. A confirmation key carrying no key material cannot satisfy either. `sign.jwk_thumbprint` already knows `OKP`, `EC` and `RSA`, so naming the RSA members adds no key type the package does not handle, and nothing that verifies today is refused.

## What this deliberately does not do

A `kty` enum would also close the unrecognised-key-type case, and it is not proposed here. Section 3.2.1 states signing algorithms per envelope context, `ES256`, `ES384` or `EdDSA` for JWT and `COSE_Sign1` or `COSE_Sign` for CBOR-COSE, and the embedded-signature form in section 3.2.2 names no algorithm set at all. A schema enum would add a constraint the specification does not make, which is a normative question and belongs in a spec change proposal rather than in a schema fix.

It is worth answering deliberately rather than by hard-coding a list now: RFC 9964 registered `AKP` as a JWK key type for ML-DSA in JOSE and COSE, so the set of key types someone might reasonably put here is not the set it was a year ago.

Found while answering #307, which is unrelated in substance: the reporter there is asking about a different member of `cnf`.

Checked on the branch: `pytest` 1328 passed and 1 skipped on 3.11 and on 3.12, with `ruff check src tests scripts`, `python tools/check_dashes.py` and `mypy src/agentrust_trace` clean, on a fresh clone with only `pip install -e ".[dev]"`. Every committed example record still validates, and the private-member refusal from #296 still holds. Removing the RSA branch fails three of the new tests, and removing the `n` and `e` declarations fails a fourth.

## Type of change

- [x] Schema change
- [ ] Example addition

## Spec section

None changed. The schema is brought into line with sections 3.2.2 and 3.3, both already merged.

## Checklist

- [x] `CHANGELOG.md` updated (for any normative change): entry added under Unreleased, Fixed


rajnisht7:
This PR corrects the JSON schema, now `n` and `e` are required in RSA Key, but python model is still old `'models.JWK`, it still accepts keys as `{"kty": "RSA"}` that is without `n` and `e`. And also the new tests validate the schema but not the complete producer/model path.

lywinged:
@rajnisht7  Both points are right, and the second one is the sharper of the two. Fixed in 33dbe14.

`models.JWK` carried this:

```python
required_by_kty = {"OKP": ("crv", "x"), "EC": ("crv", "x", "y")}
required = required_by_kty.get(self.kty, ())
```

which is the same two-entry table the schema's `allOf` carried, and `.get(kty, ())` gives `RSA` no requirement at all. So `JWK(kty="RSA")` constructed, and `JWK` is exported. I corrected one of the two artifacts that make this decision and left the other saying what it said before, which is the failure this PR is about, one layer up.

The repository already names that state. `test_all_three_layers_draw_the_line_in_the_same_place` was written for the same shape on `iat`, and its docstring says it: "while it was half made, the model and the schema disagreed."

What changed:

- `RSA: ("n", "e")` added to the model's table.
- `n` and `e` declared as members rather than left to `extra="allow"`. Two controls say the declarations are load-bearing rather than tidiness. Without them the validator raises `AttributeError` on a missing member instead of refusing the key, because `getattr` on an absent extra is not `None`. And `{"kty": "RSA", "n": 123}` is stored as an untyped extra rather than refused, which is the second reason the same two declarations are load-bearing in the schema.
- Three tests in `test_models.py` mirroring the `OKP` and `EC` pairs already there.

On your second point, the answer I think is the real one is not more cases but a different instrument. `test_cnf_jwk_carries_key_material.py` now ends with a parametrized table of eleven confirmation keys checked against the schema and the model together, asserting the two agree on each. Removing either half of the fix fails it. It also pins the deliberate open end: a `kty` neither artifact names is accepted by both, because section 3.2.1 fixes no algorithm set for the embedded-signature form of section 3.2.2, so refusing an unnamed `kty` would be a normative change and not this PR.

Where I would push back slightly is on "the complete producer path", not because the concern is wrong but because this package cannot host that test honestly. `sign_record` takes an `Ed25519PrivateKey` and `_pubkey_from_jwk` refuses any `kty` that is not `OKP`, so there is no RSA record this package can produce or verify end to end. The layer an RSA key actually reaches here is the model, which is exactly the layer you found unguarded. A producer path test would have to assert something the reference implementation does not do.

One risk this carries that is worth stating: declaring two optional members on a model whose output gets signed. That is already instrumented rather than argued. `_omit_absent_optionals` keeps them out of a dump, `test_the_round_trip_returns_the_same_record` compares the whole record including `cnf`, and disabling that serializer makes both new members appear as `null` and fails it. A record round-trips to identical canonical bytes under RFC 8785.

1342 passed and 1 skipped on 3.11 and on 3.12, up 14 from the previous commit.

imran-siddique (approved):
Approving. Nothing unsafe was ever accepted here, so the case for changing the schema rather than adding a docs note rests entirely on which artifact an implementer validates against, and the PR body makes that argument rather than assuming it.

**Reproduced against both versions**, validating the `cnf.jwk` subschema directly:

| case | main | this PR |
|---|---|---|
| `{"kty":"RSA"}` | accepted | refused, `'n' is a required property` |
| `{"kty":"RSA","n":12345,"e":"AQAB"}` | accepted | refused, `12345 is not of type 'string'` |
| `{"kty":"RSA","n":"...","e":"AQAB"}` | accepted | accepted |
| `{"kty":"OKP","crv":"Ed25519","x":"..."}` | accepted | accepted |
| `{"kty":"EC"}` | refused | refused |

The non-string modulus row is the one I would have missed reading the diff, and it is why the `"type": "string"` declarations are load-bearing rather than decorative: without them the value falls through to `additionalProperties` and anything canonicalizable is admitted. The EC row is there as a control, so the change is shown to refuse in one direction and accept in the other rather than just tightening.

`cmp schema/trace-claim.json src/agentrust_trace/schema/trace-v0.2.json` reports the files identical, and leaving the packaged v0.1 schema alone is consistent with #296.

**On the test numbers, because the raw figure is misleading.** A clean full clone with `core.autocrlf=false` gives 5 failures on this branch, and **the same 5 on pristine main**, while main's CI is green at `a79ad17c`. They are Windows artifacts:
`test_every_schema_in_the_repository_is_classified` compares os-native path strings, so it sees
`schema\trace-claim.json` against `schema/trace-claim.json`, and the generator-reproduces-fixtures
cases go the same way. A trial merge of this branch onto current main is clean and gives 1397 passed,
1 skipped, with those same 5. The RSA behaviour above survives that merge unchanged.

One note for anyone triaging behind me: the `needs-review:UNKNOWN` label on this is the contributor-check bot failing closed on a transient error, not a signal. It should not colour a read of this PR.

PR #327, opened 2026-09-10
test(vectors): pin which vector carries which discrimination

lywinged:
## What this changes

Takes the `DEFECTS` guard from #178 on its own, ahead of the `01-09` reissue, on @imran-siddique's ruling of 10 September: a guard that lands after the thing it guards protects nothing. The reissue keeps its original timing.

`test_margins_have_not_thinned` stores a count and `_margin` returns a set, so both are blind to identity. A vector rewritten into a copy of its partner leaves the rule's margin at two, and `test_vectors_for_each_rule_are_independent` still passes whenever some other declared defect happens to separate the pair. What is lost in that case is the specific property the vector was written to exercise, and nothing recorded which vector carried it.

`tests/vector_roles.json` records, per fixture, the rules it is load-bearing for and the declared defects it separates. `test_no_vector_has_lost_its_role` fails by name when a fixture drops one, on the same ratchet terms as the margins file: gaining a role is an ordinary PR, losing one is updated in the same commit with a reason.

## Why it is this shape

Of the nine fixtures the reissue moves, exactly one carries a discrimination, which the corpus itself says rather than the argument: measured over all thirty, `04` is the only vector whose issuer key the verifier holds and whose 64-byte signature still fails to verify, so it separates `checks_structure_only` while `24`, at 32 bytes, does not. `14` and `23` carry 64-byte signatures the verifier has no key for, `03` and `17` carry no receipt, and the other twenty-four verify.

What the guard holds is that shape, not the signer. A single-key reissue that collapses `04` to the wrong length, or renames it, fails by name. One that corrupts the signature at fixed length passes every guard, measured by flipping one bit of `04`'s 64 bytes: the shape survives even though the fixture no longer models a wrong signer. So the second published key that #178 asks for stays right, on the fixture's meaning, and it is a decision the reissue makes on purpose rather than one this test enforces. An earlier version of this paragraph, of the README paragraph, of the defect comment and of the docstring said the shape exists only under a second key; `f9841a4` corrects the three files, and this paragraph is the fourth surface.

The remaining six of the nine record as load-bearing with no defect of their own, which is what they are: their partners in `17-30` carry the discrimination. `01` and `02` are absent from the file altogether, because they are the valid vectors and deleting a rule does not change the outcome of a fixture that passes.

The two-key requirement is written down in two places, the `signature_or_key_mismatch` defect comment and the `examples/action-receipts/README.md` paragraph that already explains why `01-09` pin an unpublished key.

## Measured, not argued

Three mutations, each run against all three guards, and two more added after review:

| Mutation | independence | margin ratchet | this guard |
|---|---|---|---|
| `04` collapsed to a wrong-length signature | fails | passes | fails, names `04` |
| `06` and `25` swap which sits inside the grace window | passes | passes | fails, names `25` |
| `04` renamed | passes | passes | fails, names `04` |
| `04` with one bit of its 64-byte signature flipped | passes | passes | passes |
| `tests/vector_roles.json` deleted, on `62fea76` | passes | passes | skipped and rewrote the file; fails by name since `154e7e6` |

The second touches only unsigned context, so both vectors stay stale and the fixture-correctness suite still passes 35 tests: the swap is a real role exchange rather than a broken fixture. The third is the likeliest thing a reissue actually does to a file. In both of those the existing instruments see nothing, because the count is unchanged and some declared defect still separates the pair. The fourth is the limit of the guard, stated above. The fifth is the reviewer's finding, below.

Each of the three ways to lose a role reports its own reason rather than sharing one closing sentence that fits only one of them.

`vector_roles.json` is byte-identical across `PYTHONHASHSEED` values 0, 1, 7, 42, 12345, 31337 and 99999, which matters because `_margin` returns a set and `_roles` iterates it.

Suite is 1403 passed and 1 skipped at `f9841a4` on 3.11 and 3.12, with `ruff check src tests scripts`, `mypy src/agentrust_trace` and `tools/check_dashes.py` clean. The one skip is `test_version`'s environment check.

## Since opened

- `154e7e6`: both ratchets in the module fail by name when their baseline file is missing, instead of writing one and skipping. Found by @rajnisht7: with the file deleted in the same change as a degrading reissue, the first run skipped and recorded the degraded state as the baseline, and the second run passed.
- `f9841a4`: the two-key sentence corrected in the README, the defect comment and the docstring, as above. No code path changes.

## Type of change

- [ ] Schema change
- [ ] Example addition

None of the five. Tests and one README paragraph. No schema, wire format, signature semantics or verifier behaviour changes, and no fixture is modified.

## Spec section

None.

## Checklist

- [ ] `CHANGELOG.md` updated (for any normative change): not applicable, nothing normative changes

### DIFF
--- a/examples/action-receipts/README.md
+++ b/examples/action-receipts/README.md
@@ -135,12 +135,14 @@ the key used by `01`–`09` is not published; `gen_rule_coverage_vectors.py` reg
 the range byte-for-byte and only public JWKs appear in the files.
 
 Reissuing `01`–`09` from published keys takes two of them, not one (#178). `04` is the
-only vector whose signature is structurally sound and still does not verify, which is
-the shape that separates it from `24`, and that shape exists only when the signer is a
-key the verifier does not hold: under a single key the obvious way to keep `04` failing
-is to corrupt its signature, which makes it a second `24`. `tests/vector_roles.json`
-records which vector carries which discrimination, and
-`test_no_vector_has_lost_its_role` fails by name if a reissue drops one.
+only vector whose issuer key the verifier holds and whose 64-byte signature still fails
+to verify, which is the shape that separates it from `24`. A single key can keep that
+shape, by corrupting the signature at fixed length; what it cannot keep is the fixture's
+meaning, a signature made by a key the verifier does not hold. The guard below sees the
+shape and not the signer, so the second key is a decision the reissue makes on purpose
+rather than one a test enforces. `tests/vector_roles.json` records which vector carries
+which discrimination, and `test_no_vector_has_lost_its_role` fails by name if a reissue
+drops one.
 
 `tests/test_action_receipt_fixtures.py` recomputes each digest, verifies each
 signature against the pinned key, checks session and call binding, enforces
--- a/examples/action-receipts/README.md
+++ b/examples/action-receipts/README.md
@@ -134,6 +134,14 @@ Everything from `10` up pins its own deterministic test key, since the private h
 the key used by `01`–`09` is not published; `gen_rule_coverage_vectors.py` regenerates
 the range byte-for-byte and only public JWKs appear in the files.
 
+Reissuing `01`–`09` from published keys takes two of them, not one (#178). `04` is the
+only vector whose signature is structurally sound and still does not verify, which is
+the shape that separates it from `24`, and that shape exists only when the signer is a
+key the verifier does not hold: under a single key the obvious way to keep `04` failing
+is to corrupt its signature, which makes it a second `24`. `tests/vector_roles.json`
+records which vector carries which discrimination, and
+`test_no_vector_has_lost_its_role` fails by name if a reissue drops one.
+
 `tests/test_action_receipt_fixtures.py` recomputes each digest, verifies each
 signature against the pinned key, checks session and call binding, enforces
 freshness and receipt-chain ordering, and compares the result with each
--- a/tests/test_vector_completeness.py
+++ b/tests/test_vector_completeness.py
@@ -31,6 +31,7 @@ The questions, in increasing strength:
    implementation defect that one catches and the other misses? Two copies of the
    same vector satisfy question 3 and fail this one.
 5. Has any rule's margin dropped below what it was? (silent thinning)
+6. Is each vector still the one doing the discriminating? (silent substitution)
 
 Independence is #124's definition made executable. For every rule, ``DEFECTS``
 declares at least one *weakened* variant of its check, each modelling a real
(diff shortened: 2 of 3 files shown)


rajnisht7:
the new test `test_no_vector_has_lost_its_role` checks the file `vector_roles.json` to confirm if all vectors are working properly, but if the file is missing, the code creates the file and makes the test green.
the issue is if someone deletes the file, the file will be created automatically during the test, and will make the test green but the deletion of file is never detected, that is guard can turn itself off

lywinged:
> the new test `test_no_vector_has_lost_its_role` checks the file `vector_roles.json` to confirm if all vectors are working properly, but if the file is missing, the code creates the file and makes the test green.
> 
> the issue is if someone deletes the file, the file will be created automatically during the test, and will make the test green but the deletion of file is never detected, that is guard can turn itself off 

Thank you @rajnisht7 🙏, this is exactly the case the guard was supposed to make loud and it made it quiet instead. With tests/vector_roles.json deleted, the test wrote a fresh file from the current tree and reported skipped, and the next run passed. Measured on 62fea76 with 04 collapsed to a 32-byte signature and the file removed in the same change: first run skipped and rewrote the file with 04's discrimination gone, second run passed, nothing red.
Fixed in 154e7e6. A missing file now fails by name, and the message says to restore it from history, or to rebaseline on purpose in the same commit that changes the fixtures, with the reason. The margins ratchet above it had the same shape and takes the same fix, since your point applies to both. With either file deleted its test fails; with both present the module passes 37 and the suite 1403 with 1 skipped on 3.11 and 3.12, the one skip being test_version's environment check, and CI is green on this head.

imran-siddique (approved):
Approving. This is the guard from my 10 September ruling on #178, landed on its own ahead of the reissue, and it closes a hole the ruling did not name.

**What the ruling asked for, and what each piece does.** I said the `DEFECTS` assertion lands first because a guard that arrives after the thing it guards protects nothing, and that the reissue keeps its original timing. Both hold here: `tests/vector_roles.json` and `test_no_vector_has_lost_its_role` are the guard, the reissue is untouched.

**The part beyond the ruling is the part worth having.** `test_margins_have_not_thinned` stores a count and `_margin` returns a set, so both are blind to identity. A vector rewritten into a copy of its partner leaves the rule's margin at two and `test_vectors_for_each_rule_are_independent` still passes whenever some other declared defect separates the pair. The property that vector was written to exercise is gone and nothing records which vector carried it. That is a real gap and I had not seen it.

**Ran it rather than read it.** `vector_roles.json` records 28 fixtures. Suite is 37 passed. Then I mutated `04`'s signature from 64 bytes to 32, making it a structural copy of `24`, which is the exact scenario described above. Two tests fail by name:

```
04-signature-key-mismatch.json: no longer separates ['checks_structure_only'] for
'signature_or_key_mismatch'. It still fails for the rule, so what it lost is the
margin rather than the coverage: it has become a copy of its partner.
```

The message distinguishes losing the margin from losing the coverage, which is the distinction that makes the failure actionable rather than just red.

Your measurement that `04` is the only vector of the thirty whose issuer key the verifier holds and whose 64-byte signature still fails to verify matches what I measured independently on 9 September, from the fixture bytes: `04` decodes to 64 and `24` to 32, and `vector_margins.json` records the rule at two.

The ratchet terms are right too: gaining a role is an ordinary PR, losing one is declared in the same commit with a reason. That is what keeps this from becoming a file people edit to make CI green.

PR #329, opened 2026-09-11
test(sandbox): name verify_record(expected_nonce=...) in the docstring

imran-siddique:
Follow-up to #315, closing the one clause @lywinged flagged and explicitly declined to hold that merge for.

After #315 the `nonce` field docstring names `verify_record(expected_nonce=...)`, while the test docstring beside it still said `nonce` is never checked "against anything else". A verifier can compare it against a value it chose; what that establishes is freshness rather than key binding, which is the distinction the surrounding paragraph is drawing. The docstring now says that.

Also drops a doubled `unrelated,` in the same sentence.

Docstring only. No test behaviour changes.

https://claude.ai/code/session_01X27MBo6tGVw1UmmchuGFnW

### DIFF
--- a/tests/test_sandbox_adapter.py
+++ b/tests/test_sandbox_adapter.py
@@ -244,9 +244,10 @@ def test_genuinely_verified_evidence_is_still_not_bound_to_the_signing_key() -> 
     asks of them -- still cannot get real Level 1 assurance out of this adapter, because
     nothing here binds that evidence to the specific key the record ends up signed
     with. ``nonce`` is carried through verbatim and is never checked against
-    ``cnf.jwk``, against the signing key passed to ``sign_record``, or against
-    anything else. Two records built from the identical (hypothetically genuine)
-    attestation but signed with unrelated, unrelated-to-the-hardware keys both verify
+    ``cnf.jwk`` or against the signing key passed to ``sign_record``. A verifier can
+    compare it against a value it chose, via ``verify_record(expected_nonce=...)``,
+    which establishes freshness rather than key binding. Two records built from the
+    identical (hypothetically genuine) attestation but signed with unrelated keys both verify
     successfully; nothing distinguishes "the verified key" from "any key the caller
     felt like using afterwards". Per docs/trust-levels.md, Level 1 requires
     "authenticated evidence binding the record-signing key to the expected


lywinged (approved):
Approved on `73a472b`.

Stripped of every docstring, `tests/test_sandbox_adapter.py` parses to the same tree on this head and on `df0120b`. The four changed lines, 247 to 250, sit inside the docstring that runs from 240 to 256, so "Docstring only" is measured. "unrelated" occurs twice in that docstring on `df0120b` and once here, which is the doubled word the body names. No CHANGELOG entry, as with #270, #278 and #327.

The check the new sentence names is there: `verify_record` takes `expected_nonce` as a keyword-only argument and compares it with `runtime.nonce` through `hmac.compare_digest`, `sign.py` lines 706 to 709. The test's own two records, built from the identical attestation and signed with two fresh keys, both verify, each against its own key, once `expected_nonce="claimed-binding-to-key-A"` is added to the call. The check is satisfied under either key and says nothing about which one. A different expected value refuses both with `record runtime.nonce does not match expected_nonce`. That runs the key-binding half; the freshness half is the spec's classification rather than a run: `spec/trace-v0.2.md` puts challenge-nonce binding under the Freshness paragraph of 3.2.2, and `sign.py` labels the check the same way at line 705.

The four CI steps with the hash-pinned install, on 3.11 and on 3.12: green, 1404 passed, 1 skipped, the sandbox file 45 passed. Merged onto `main` at `10fcba4`: 1410 passed, 1 skipped on both interpreters, and no file here is one #322 or #327 touched. GitHub's merge ref for this PR, `e05a3a9`, is `73a472b` onto `df0120b`, so the run onto `10fcba4` is the one that includes those two.

One correction to my round two on #315, which called this test docstring "the one line left saying the other". It was not the only one. The `SandboxAttestation` class docstring carries the same closing clause at `sandbox.py` lines 125 to 126, "against any key, or against anything else", in the same shape as the sentence this PR corrects. It arrived in the same diff, and both my reviews there passed over it. So the class denies in that sentence the check its own field docstring names at line 146. The module docstring's own denials stay true: "nothing here checks a quote, a signature, or a nonce" at line 35 and "nothing here ties" at line 52 are scoped to the adapter, and "nothing in this codebase does" at lines 64 to 65 to the key binding. A case-insensitive sweep of this head for "anything else" gives five lines; the other four, in `spec/registry-anchor-v1.md`, `content_marking.py`, `sign.py` and `tests/test_public_functions_raise_what_they_document.py`, are not about the nonce. For lines 124 to 126, something like, in the field docstring's own words:

```
    that binding on its own -- ``nonce`` is carried through to the record verbatim, on
    trust, and is not checked against ``measurement`` or against any key; downstream
    ``verify_record(expected_nonce=...)`` compares it only with a value the verifier
    already knows. If you bind your quote's challenge to your signing key yourself,
```

Not this PR's line and not a reason to hold it. Yours to take here or in a follow-up.

Issue #213, opened 2026-08-24
docs: stable per-primitive anchors so external standards can deep link (OpenCRE prerequisite)


imran-siddique:
## Problem

TRACE is being mapped into [OpenCRE](https://opencre.org), which cross-links a topic to ASVS, CWE, ISO 27001, NIST 800-53, SAMM and the OWASP AI Exchange from a single stable link.

Working the first pass surfaced a prerequisite we do not meet. OpenCRE links into a **stable anchor for one specific requirement** and regenerates those links from its API in CI. Our spec pages have no stable per-primitive anchor, so an external standard can only link to the page.

## What is needed

Durable anchor IDs for the primitives an external standard would actually cite, starting with:

- Trust levels L0 / L1 / L2 in `docs/trust-levels.md`
- The evidence record structure in `docs/schema.md`

Durability is the point. OpenCRE's pitch is that its links do not rot, so an anchor that shifts when a heading is reworded defeats the purpose.

## Wider than this repo

The same gap applies to every agentrust-io spec carrying a control an external framework would reference: `cmcp` (tool invocation control), `ca2a` (inter-agent trust), `agent-manifest` (agent identity, capability manifest), `agentic-usage-control` (continuous usage control), `weight-custody-manifest` (weight custody).

Proposing we settle the convention here first and then apply it across the others rather than inventing a different scheme per repo.

## Context

First-pass mapping found 25 of 30 controls land on existing CREs and 19 look like genuine additions. Discussion with the OpenCRE maintainers is in flight.

imran-siddique:
Correcting the premise in the body, and the requirement has grown a second half.

**The "19 genuine additions" line above is out of date.** Rob van der Veer's rule on OWASP/OpenCRE#1045 is that OpenCRE organizes requirements that standards state, it does not state them, and a sub-control earns its place once several standards have sections to group there. Applied to all 30 controls, all 19 proposed CREs collapsed into links to existing CREs. The second pass proposes none: 30 sections, 62 links, 53 distinct CREs. That makes this issue more important rather than less, because the entire contribution is now link-shaped.

**The second half.** OpenCRE's `docs/CONTRIBUTING.md` names an alternative linking method and calls it the preferred one, because it is self-maintaining:

[...]

PR #330, opened 2026-09-11
docs: stable per-primitive anchors on trust levels and the schema reference

imran-siddique:
Closes the prerequisite in #213, and proposes the convention there for the other five repos.

## The scheme

`trace-` prefix, then the primitive's own name, never the heading prose.

| Anchor | Heading today |
|---|---|
| `trace-l0` | Level 0: software-only |
| `trace-l1` | Level 1: hardware evidence |
| `trace-l2` | Level 2: transparency anchoring |
| `trace-field-model` | `` `model` `` |
| `trace-field-build-provenance` | `` `build_provenance` `` |

Rewording "Level 1: hardware evidence" leaves `trace-l1` untouched. That is the property OpenCRE's pitch rests on: it regenerates its links from its API in CI, so an anchor that tracks prose is an anchor that rots.

## Existing links are preserved

Every heading gets an explicit `<a id="old-slug"></a>` above it, so all 22 current deep links still resolve.

The old slugs were measured by rendering the files before the change rather than derived from a slugify function. That is how the underscores in `data_class`, `tool_transcript` and `build_provenance` survived: `toc` preserves them and my first pass had stripped them.

## Verification

`mkdocs build` aborts in my environment on three plugins that are not installed (`llmstxt`, `minify`, `mkdocstrings`), so this was checked with `python-markdown` using the `attr_list` and `toc` extensions `mkdocs.yml` declares: 21 new ids render, and every original anchor still resolves in both files. CI is the real check.

## Scope

Two files, both named in #213. No prose changes, no heading text changes, no nav changes.

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -1,4 +1,6 @@
-# Schema Reference
+<a id="schema-reference"></a>
+
+# Schema Reference {#trace-schema}
 
 JSON Schema for the TRACE v0.2 Trust Record. Source: [`schema/trace-claim.json`](https://github.com/agentrust-io/trace-spec/blob/main/schema/trace-claim.json).
 
@@ -8,7 +10,9 @@ unambiguously, since RFC 8785 serializes numbers through an IEEE 754 double and 
 can share one. A value that needs to be larger is carried as a string. The same bound applies to members a
 `cnf.jwk` carries that this schema does not name.
 
-## Top-level fields
+<a id="top-level-fields"></a>
+
+## Top-level fields {#trace-fields}
 
 | Field | Type | Required | Description |
 |---|---|---|---|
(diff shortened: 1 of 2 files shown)


lywinged (approved):
Approved. Reviewed baac9b6. The anchor changes look good to merge.

I completed full MkDocs builds of the base and head locally using the documentation workflow’s assembly and configuration, with the documentation dependencies installed from the hash-pinned lockfile. The focused render comparison preserves all 22 existing IDs and adds 21 new IDs, with no duplicates or changes to the visible prose. The built pages also preserve the existing heading targets.

A near-miss check confirms the underscore caveat already described in the PR: a simple slugifier matches 19 of the 22 legacy heading IDs and breaks #data_class, #tool_transcript and #build_provenance. Please retain those exact aliases during future cleanup.

One non-blocking correction to the verification paragraph: “CI is the real check” does not describe a pre-merge check on this PR. The Docs workflow runs on a push to main or a manual dispatch, not on pull_request; the current 6 checks contain 5 successes and 1 skipped result, with no documentation build. The full local builds above supply that verification for this commit. Please distinguish that local evidence from the documentation CI that runs after merge.

Issue #338, opened 2026-09-12
spec: bind successor observations without treating transcript.after as proof of closure


altrudev:
## Problem

The PIC/TRACE bridge currently binds the signed authorization to the declaration and exact executed tool call, and when a transcript is required it binds `transcript.before.tool_call` to that execution.

`transcript.after`, however, is only required to be a JSON object.

That establishes presence, not identity or evidentiary sufficiency. A verifier therefore cannot distinguish:

1. the actual successor observation produced for this execution,
2. a substituted or replayed object,
3. a self-asserted outcome supplied by the executor,
4. a valid observation that nevertheless is insufficient to establish the requested transition.

This boundary is already stated in `docs/integration/pic-trace-bridge-v1.md`: the bridge binds authorization to the call and execution evidence "without claiming that TRACE proves the real-world outcome of the call."

#332 confirmed that no existing normative mechanism closes this layer and identified `transcript.after` as the narrow surface worth addressing.

This issue proposes only that narrow change. It does **not** propose importing STBP wholesale, turning TRACE into an application-state store, or making real-world outcome claims from execution evidence alone.

Related: #332, #279.

## Requirements

A successor-observation extension to the PIC/TRACE bridge should satisfy all of the following.

### 1. Bind the exact successor observation

When a profile elects to make a successor-state claim, the verifier MUST be able to establish the exact observation bytes to which the claim refers.

The identity relation should use the bridge's existing RFC 8785 / SHA-256 convention rather than host-language object equality.

A candidate minimal field is:

```json
{
  "after": {
    "observation": { "...": "..." },
    "observation_digest": "sha256:..."
  }
}
```

The precise schema shape is intentionally open. A detached successor-observation artifact bound by digest may fit the existing bridge architecture better than embedding the complete object.

The requirement is the invariant, not that specific representation:

> the bridge MUST bind the exact successor observation used by the verifier.

### 2. Keep observation integrity separate from observation sufficiency

Matching a digest proves identity of the observation. It does not prove that the observation is authoritative enough to establish the requested transition.

The verifier therefore MUST NOT equate:

```text
successor observation is integrity-bound
```

with:

```text
authorized transition is established
```

A successor observation may be correctly bound and still be insufficient because its source is untrusted, stale, incomplete, or otherwise outside the verifier's policy.

### 3. Make the not-established result first-class

This is an instance of the discipline tracked in #279.

[...]

PR #340, opened 2026-09-12
feat(intent-bridge): bind successor observations for #338

altrudev:
Implements #338 as the narrow successor-observation extension accepted in the issue review.

## Governing rule

The bridge must not treat a bound `transcript.after` object as proof that a requested real-world transition occurred.

This PR keeps successor **integrity** separate from successor **sufficiency**:

- when `transcript_required=true`, the signed authorization carries `successor_observation_digest`;
- `verify_bridge()` requires `transcript.after` to be exactly the successor envelope whose RFC 8785 / SHA-256 digest matches that signed value;
- the binding covers observation content, observer identity, and observation timestamp;
- a caller cannot substitute an observation and a matching expected digest without invalidating the authorizer signature;
- `successor_observation_digest` is forbidden when `transcript_required=false`, so non-transcript bridge use is not widened accidentally;
- trust, freshness, observer-independence policy, and the transition predicate are evaluated separately;
- outcomes stay surface-local: `established`, `contradicted`, or `not-established`;
- malformed artifacts and digest mismatches are refusals, not evidence outcomes;
- absence at the separate successor-evaluation surface is `not-established`; absence at the bridge layer when the signed authorization requires a transcript is a refusal;
- executor/observer separation is required only when verifier policy requires it.

## Integration

The exact-envelope binding now lives in `agentrust_trace.intent_bridge`, beside the existing canonical-byte binding for `transcript.before.tool_call`.

The expected successor digest is authenticated by the same signed authorization that binds the tool call. It is not caller-supplied independently.

The tri-state successor evaluator remains separate so a successful integrity check cannot silently become transition closure.

The informative PIC/TRACE bridge profile now states the assurance boundary for requirements 2, 4, and 5, including the treatment of absent successor evidence.

No TRACE Trust Record schema change. No replay semantics. No universal transition-predicate language. No application-state store.

## Adversarial coverage

The matrix covers:

- valid bound trusted successor -> established;
- trusted contradictory successor -> contradicted;
- absent successor at evaluation surface -> not-established;
- observation substitution -> refusal;
- attempt to substitute both observation and expected digest without re-signing -> refusal at authorization signature;
- observer relabelling -> refusal;
- timestamp retiming -> refusal;
- future/stale/untrusted successor -> not-established;
- zero-age freshness boundary;
- empty trusted-observer set -> not-established;
- executor self-observation with independence required -> not-established;
- executor self-observation when independence is not required -> established;
- undecidable predicate -> not-established;
- transcript-required / successor-digest presence relation in both directions.

AI-assistance disclosure: ChatGPT assisted with DDC radial analysis, implementation drafting, and adversarial review. `altrudev` remains responsible for the contribution.


rajnisht7 (commented):
Integrity vs sufficiency split is in looks in the right direction, but there are a few gaps:

- `verify_bridge()` still does not bind the successor When transcript_required is true, the path still ends roughly as:
```python
if not isinstance(transcript.get("after"), dict):
    raise AuthorizationMismatch(...)
```
and there is no call to _bind_successor_observation(), and no check against a successor digest.

- `expected_successor_digest` is not authenticated by the bridge evaluate_successor_observation(..., expected_successor_digest=...) only checks:
`digest(after) == expected_successor_digest`
That expected value is caller-supplied and it is not a field of the signed authorization, and schema/pic-trace-bridge-v1.json has no successor digest either. The module docstring even says the prototype does not decide whether that digest lives in the signed authorization, the transcript, or a detached artifact.

- #338 says that when a profile makes a successor-state claim, the verifier must establish the exact observation for this evaluation, so substituted/replayed after objects can be detected.
If the caller can also choose expected_successor_digest, a substituted observation can still be made to “match” by passing expected = digest(forged_after). Until the expected digest is tied to something authenticated (signed authorization, committed transcript field, or a detached artifact that is itself verified), requirement 1 is only partly met.

- Two meanings of after:
In verify_bridge: after = any execution-result object (dict).
In _bind_successor_observation: after = exactly {observation, observer, observed_at}.

- Docstrings still say “prototype” while the PR title is feat(...).

- Optional tests: future observed_at, max_age_seconds == 0, empty trusted_observers (logic looks fine when tried manually).

altrudev:
Addressed the review on current head `6e05c5f52112e4418c66ae19d4994a74ffa60237`.

Changes against your points:

- `verify_bridge()` now binds `transcript.after` directly.
- The expected successor digest is no longer an unauthenticated caller input at the bridge layer. It is `authorization.successor_observation_digest`, covered by the authorizer signature.
- A caller that substitutes both the successor envelope and a matching digest without re-signing now fails at the authorization signature.
- `transcript.after` has one meaning when `transcript_required=true`: exactly `{observation, observer, observed_at}`.
- `successor_observation_digest` is required iff `transcript_required=true`; it is forbidden when false so non-transcript bridge use is not widened.
- The remaining prototype wording was removed/updated to reflect the authenticated binding contract.
- Added the optional edge cases you suggested: future `observed_at`, `max_age_seconds == 0`, and an empty trusted-observer set.
- Updated the repo-wide public-function and safe-integer fixtures to the new bridge shape.

Exact-head checks are green:
- CI: Python 3.11 + 3.12, including lint, house style, mypy, and full tests
- CodeQL: success

I also tightened the profile text to distinguish bridge-layer absence (refusal when the signed authorization requires a transcript) from absence at the separate evidence-evaluation surface (`not-established`).

lywinged (dismissed):
Approving 

This was verified by running the branch, not by reading the diff alone. The four checks ci.yml runs, ruff, the em-dash check, mypy and the test run, all pass on this head in my own checkout, and the branch merges into current main at 372b9af without conflict, with the full suite green on the merged result, so the seven commits that landed on main since this was opened cost nothing.

The description's claims hold. Its adversarial matrix is fourteen bullets, sixteen cases once the future, stale and untrusted bullet is split, and each one maps to a test that exists by name. The authorizer signature is computed over {profile, authorization}, so it covers successor_observation_digest, and substituting the observation together with a matching expected digest, which is the substitution the earlier review raised, fails at the signature rather than at the binding; test_caller_cannot_substitute_successor_and_matching_digest_without_resigning is that case. The digest covers the complete envelope, so relabelling the observer or retiming the observation breaks the binding, and both have their own test. successor_observation_digest is required when transcript_required is true and forbidden when it is false, in verify_bridge and in the schema's allOf if/then/else, so the two statements of the same rule cannot drift apart quietly. Integrity and sufficiency stay apart: the bridge binds, evaluate_successor_observation decides, malformed artifacts and digest mismatches raise rather than returning a status, and all seven paths that fall short, absent, untrusted, future-dated, stale, no executor identity under required independence, observer equal to executor under it, and an undecidable predicate, land on not-established.

The matrix discriminates rather than only passing. Dropping the forbidden-when-false branch, the future-dated check, or the independence check each fails exactly one test and nothing else.

One clause is not pinned. Removing the rejection of unknown fields in transcript.after leaves all seventy tests in the two bridge files passing and the full suite at its baseline of 1585 passed and 1 skipped, so nothing in the repository notices. It is not a forgery path: the digest covers the whole envelope, so an added field changes the digest and fails the binding unless the authorizer signed that envelope deliberately. The rule is enforced in the code. What nothing exercises is the enforcement, the shape the description states: transcript.after exactly {observation, observer, observed_at} when a transcript is required.

One thing this approval covers deliberately. #338 left open whether an absent after is a refusal or not-established, leaning to not-established because the bridge cannot tell a profile that elected a successor claim from one that did not. This resolves it the other way at the bridge layer: a refusal when the signed authorization requires the transcript, not-established at the separate evaluation surface. The signed successor_observation_digest is what now lets the bridge tell the two apart, so the reason for the leaning no longer holds, and the profile text says which applies where. I am approving that resolution rather than passing over it.

All six points in the earlier review are addressed on this head, the prototype wording included: the word is no longer anywhere in the repository. The repository-wide public-function and safe-integer fixtures are extended to the new shape rather than relaxed.

altrudev:
@lywinged I closed the one unpinned clause from your approval on a test-only head update.

New exact-head test: `test_successor_envelope_rejects_unknown_fields_even_when_signed`.

It deliberately re-signs the authorization over the extended envelope's digest before verification, so it isolates the profile-shape rule itself: an authentic, correctly bound `transcript.after` carrying an extra field is still refused because the profile requires exactly `{observation, observer, observed_at}`. Removing only the unknown-field rejection now fails this test.

No production code or profile semantics changed from the head you approved.

Current head: `fdc40f1d2ab8113838f6e8b20b7b8ad3506d9574`
- CI: success (3.11 + 3.12, lint, house style, mypy, full tests)
- CodeQL: success

The previous approval was automatically dismissed by the new commit.

lywinged (approved):

imran-siddique (approved):
Reviewed fdc40f1d2ab8113838f6e8b20b7b8ad3506d9574.

The signed authorization binds the complete successor envelope, including observer and time. Trust, freshness, independence policy, and the transition predicate are evaluated separately; insufficient evidence stays not-established. The latest test also pins rejection of unknown envelope fields. Trial merge with current main is clean. Local bridge/successor tests: 71 passed; removing the unknown-field guard causes one failure. Ruff and mypy pass. Full Windows run: 1,596 passed, one skipped, five failures; the identical five generator/path-classification failures reproduce on unchanged main. Hosted CI and CodeQL pass on the reviewed head. No blocking findings in this change.

PR #341, opened 2026-09-13
test(runtime-evidence): add external signer context vector

altrudev:
Adds the missing consumer-side trust-context case discussed on #277.

The runtime-evidence corpus already covers:
- forged quote -> reject
- valid quote with mismatched runtime.measurement -> reject

This PR adds the separate signer-trust boundary: a record may verify its own signature under its embedded `cnf.jwk` and still fail to establish signer trust under the relying party's externally configured trust roots.

The committed vector keeps those dimensions separate:
- runtime evidence: `platform-attested`
- model claim: `self-reported`
- signer trust: `not-established`

The verifier context is intentionally non-empty and contains a real deterministic but different trusted Ed25519 key. The test proves:
1. the record envelope is internally valid under the embedded key;
2. the embedded key is not one of the configured trusted roots;
3. the configured trusted key cannot authenticate the record signature.

The vector metadata is data-driven; `generate.py` has no filename-specific special case.

No runtime-grade algorithm change and no schema change.

AI-assistance disclosure: ChatGPT assisted with DDC radial analysis, implementation drafting, and test design. `altrudev` reviewed the bounded claim and remains responsible for the contribution.


rajnisht7 (commented):
the name of the file is `reject-embedded-key-as-trust-root` but the grade is platform-attested, not reject. All other reject-* vectors actually fail. kindly rename it

The record is full copy of accept-real-quote-platform-attested.json. Only the extra signer_trust / empty trusted_root_keys fields are new. That’s okay for a trust-context example, but the name should not say “reject”

In test_embedded_key_is_not_external_verifier_context, trusted_root_keys is always [], so embedded not in configured always passes.

there is hard-coded name in generate.py Special-casing if name == "reject-embedded-key-as-trust-root" is fragile. Please make context / extra expected fields part of the normal vector data instead.

lywinged (changes requested):
The vector itself I could not fault. Two things before it goes in, neither of them about the vector.
1. Two documents still describe a corpus of thirteen. The count moved in
tests/test_runtime_evidence_vectors.py, in the docstring and in the assertion, and nowhere else.
docs/rfcs/runtime-evidence-profile.md still says "13 vectors, generator, and reference rules" in
its header, still quotes the run closing at 13/13, still says the job "regenerates all 13
vectors", and its vector table has thirteen rows and none for this one.
examples/runtime-evidence/README.md, in the directory that header calls the conformance material,
has its own thirteen-row table under "The vectors" and never names this vector. The 13/13 is not
a phrase to leave alone: generate.py builds that line from len(rows) rather than writing it
out, so it counts whatever the generator builds, and the profile is quoting a run over thirteen.
Nothing in the suite asserts what either document says, so it stays green while both of them
disagree with what is in vectors/.
2. No commit on the branch carries a DCO sign-off. All eleven, 60c16d2 through e15312a,
carry no trailer block at all. CONTRIBUTING.md line 42 requires it and line 48 says a pull
request without it will not be merged, the pull request template carries it on its checklist, and
no workflow in this repository checks it, so it falls to review. Better said now than found at
merge: git rebase --signoff 6737005 fixes the branch in one command.
What I checked, so that the second round is about those two and nothing else. The record verifies
under its embedded cnf.jwk over the RFC 8785 bytes with signature removed, the configured
trusted key refuses that same signature with InvalidSignature, and the two keys differ. I
reproduced your three points with cryptography and rfc8785 directly rather than through
generate.py, which imports agent_manifest._tdx_verify and needs the pinned verifier I do not
have here. I also built the trust context two other ways: pointing it at the record's own embedded
key fails the identity check and the cryptographic one, each exercised on its own, and emptying it
fails the non-empty guard, so the two vacuous versions of this case are refused. Dropping
signer_trust from expected fails the expectation check. The refactor of build_corpus leaves
every other committed vector alone, one added file and nothing modified under vectors/. The
branch merges cleanly into main at 3aaf0dd and the suite is green on the result. I read #277
and its thread, and this is the third consumer-side boundary as it was asked for there, kept out of
the grade.
One thing worth knowing rather than finding later. The two assertions that are the point of this
vector, that the embedded key is not among the configured roots and that a configured root cannot
authenticate the signature, run only in the appraise job, because pyproject.toml sets
testpaths = ["tests"] and the root suite never collects that file. What the root suite does carry
is the corpus count, the profile across the corpus, the two ceilings on the grade, and per vector
the expectation, the draft schema, the signature and the evidence shape.

altrudev:
Addressed both requested changes on current head `287af14656ac04ed250ad04bb1ab1ea755480da0`.

1. Corpus documentation is now consistent at 14 vectors:
- `docs/rfcs/runtime-evidence-profile.md` now says 14 vectors, reports 14/14, regenerates all 14, and includes `context-embedded-key-not-trusted` in the table.
- `examples/runtime-evidence/README.md` now includes the same vector in its table.
- I checked both documents for the stale phrases `13 vectors`, `13/13`, and `all 13 vectors`; none remain.

2. DCO is repaired:
- I first rewrote all branch commits with `Signed-off-by: Altru.dev <altrudevelop@gmail.com>`.
- A radial merge check then found the rewritten history was no longer mergeable against current main.
- I therefore rebuilt the final reviewed state directly on current upstream main `cd92e16dde6658eedbe1058c903f3fce92bb3a58` as one signed-off commit, preserving the upstream link fixes in `runtime-evidence-profile.md`.
- The PR is now one commit, six intended files, and reports mergeable.

Exact-head validation:
- CI: success on Python 3.11 and 3.12, including lint, dash/style check, mypy, and full tests.
- runtime-evidence: success with the pinned real external verifier.
- CodeQL: still in progress on this exact head at the time of this comment.

No runtime-grade algorithm or TRACE schema change was introduced by the review fixes.

altrudev:
Follow-up on the exact head above: CodeQL has now completed successfully as well. CI, runtime-evidence, and CodeQL are all green on `287af14656ac04ed250ad04bb1ab1ea755480da0`.

lywinged (approved):
Approved at `287af14`. Both things I asked for are done: the branch is one commit carrying a sign-off, and the count now reads 14 in the profile header, in the `14/14` line, in the regeneration sentence and in the README table, which names this vector. `pytest` 1659 passed, 1 skipped.

imran-siddique (approved):
Approving at `287af14`. The three points from @rajnisht7 are resolved in the diff: the vector is named `context-embedded-key-not-trusted`, `generate.py` carries the context as vector data with no filename case, and the test asserts the trust context is non-empty and that the configured key fails to verify the signature. The corpus count reads 14 in both documents, as @lywinged asked.

PR #313, opened 2026-09-10
docs(verification): add Article 12 and correct the Article 111(2) attribution

ioanavalea:
## What this changes

Two corrections to the informative section added in #310, in one diff at @lywinged's suggestion in review there.

Article 12 joins Article 11 and Annex IV as a provision a reader might expect to supply a verification depth, and does not. The Article 111(2) sentence is corrected, because it presented a recital as operative text.

## Type of change

- [x] Editorial (typo, link fix, clarification: no normative effect)
- [ ] Non-breaking spec change (new optional field, new platform profile, informative addition)
- [ ] Breaking spec change (requires 14-day comment period and Project Lead sign-off)
- [ ] Schema change
- [ ] Example addition

## Spec section

None. `docs/verification.md` only. Section 3.3.1 of `spec/trace-v0.2.md` is unchanged and no schema file is touched.

## Why

**Article 12.** Suggested on the ground that a reader who knows this repository will look for it. Article 12(1) requires that a high-risk system technically allow the automatic recording of events over its lifetime, which addresses neither provenance nor build inputs, so it makes the section's argument rather than complicating it. Article 12 sits in Chapter III Section 2, so the closing sentence's list of provisions is extended to `Articles 11, 12, 15 and 25` rather than left inconsistent. The count in "Two adjacent provisions are sometimes read as supplying one" is deliberately untouched, since that sentence refers to Article 25(4) and Article 15(5).

**Article 111(2).** The merged sentence made the dates "subject to the Article 111(2) grace period for units of a type and model already placed on the market". Article 1 point (39)(a) of Regulation (EU) 2026/1744 replaces Article 111(2), and the replacement carries the significant-changes test and a 2 August 2030 compliance date for providers and deployers of high-risk systems intended to be used by public authorities. It does not carry the type-and-model language. That reading is recital 39, and "type and model" occurs three times in the whole Regulation, all three inside that recital and none in the operative articles. The reading was right and the attribution was not.

## Checklist

- [x] DCO sign-off on all commits (`git commit -s`)
- [ ] `CHANGELOG.md` updated (for any normative change): not applicable, no normative change
- [ ] Breaking changes marked in spec text: not applicable
- [ ] Backward compatibility statement included: not applicable

## AI assistance

[...]

PR #343, opened 2026-09-14
fix(adapters): validate iat before it reaches the record, not after

harshnair75567-cloud:
SandboxSessionResult and AGTSessionResult wrote session.iat straight into build_trust_record's top-level "iat" key with no type check. Every other field reaches the record through a models.py pydantic constructor, which coerces or refuses a bad value before it is dumped -- iat bypassed all of them.

TrustRecord.model_validate(), which sign_record's own docstring recommends for structural validity before writing, gave a false pass on a numeric-string iat: pydantic's lax mode coerces "1800000000" -> 1800000000 during validation, but that happens on a new object and never touches the dict being signed. The wire bytes kept the string, which jsonschema.validate against trace-v0.2.json and this library's own verify_record both then reject.

AGTSessionResult additionally had no post_init at all.

Same failure mode as provenance.build_record's pre-#320 issued_at coercion, fixed there by validating the caller's explicit value before use instead of converting it. This applies the same fix to both adapters, enforcing the TRACE v0.2 contract floor (TRACE_MIN_IAT = 1700000000, matching models.py and both schema copies) up to JCS_SAFE_INTEGER, with bool excluded since bool subclasses int. Both session classes are now frozen to close plain reassignment of a checked value after construction, and the bound is also enforced at the point the record is built, since freezing alone does not stop vars() or object.__setattr__.

1583 passed, 1 skipped. ruff and mypy --strict clean on both changed source files.

What this changes
Type of change

* Editorial (typo, link fix, clarification: no normative effect)
* Non-breaking spec change (new optional field, new platform profile, informative addition)
* Breaking spec change (requires 14-day comment period and Project Lead sign-off)
* Schema change
* Example addition

Spec section
Checklist

* [x] DCO sign-off on all commits (`git commit -s`)
* CHANGELOG.md updated (for any normative change)
* Breaking changes marked with `<!-- CHANGED: #NNN: description -->` in spec text
* Backward compatibility statement included (for breaking changes)


rajnisht7 (commented):
reviewed the PR head [604a635](https://github.com/agentrust-io/trace-spec/pull/343/commits/604a635136b8ed43b501fce088340f5d27a8bf4d), and there a few things to look upon:

- The validation added in both adapters is incomplete.
Current check:
```python
or self.iat < 0
or self.iat > JCS_SAFE_INTEGER
```
But the TRACE v0.2 contract for iat is:

> Schema (schema/trace-claim.json): "minimum": 1700000000
> TrustRecord model: Field(ge=1700000000, le=JCS_SAFE_INTEGER)

So values like iat=0, iat=1, iat=1699999999 still successfully construct `AGTSessionResult` / `SandboxSessionResult` and get written into the record by build_trust_record().

- Tests also miss the actual lower bound.

Current parametrization:
```python
["1800000000", True, False, -5, 1.5, 2**60]
```
This covers type/range problems on the upper side and non-integers, but never tests values in [0, 1699999999].
Kindly also add: 0, 1, 1699999999, exact minimum 1700000000 (should pass), exact maximum JCS_SAFE_INTEGER (should pass)

- The closing parenthesis of the existing `image_digest` `raise ValueError` got mis-indented

- Minor style: missing blank line between the `iat` field docstring and `def __post_init__`.

imran-siddique (changes requested):
Confirmed Rajnish’s lower-bound finding at `604a635`: both adapters accept `iat=0`, `1`, and `1699999999`, then emit records rejected by the schema.

Please enforce `1700000000 <= iat <= JCS_SAFE_INTEGER` in both session constructors, retaining the type checks and boolean rejection. Add rejection tests for those three values and positive tests at both exact boundaries, validating the resulting records against the schema.

Please also fix the indentation and blank-line issues Rajnish identified. The existing 84 adapter tests pass but do not cover the missing lower bound.

Could you update this by September 18?

harshnair75567-cloud:
The fix is almost complete and I will be updating the pr today itself

rajnisht7:
The core fix looks fine, but a few things needs to be looked:

- the boundary tests accept 1700000000 and `JCS_SAFE_INTEGER`, but they only check `session.iat` also call `build_trust_record()` and validate the resulting record with `validate_json(record)` (schema check) also `TrustRecord.model_validate(record)` can be added as an extra model-level check, but alone it is not enough as that was the original false-pass path.

- the AGT test is still named test_iat_must_be_a_bounded_non_negative_integer even though the check is no longer “non-negative” and fix the comment that points at schema/trace-v0.2.json the source schema is schema/trace-claim.json.

- the second commit [f39d0c5](https://github.com/agentrust-io/trace-spec/pull/343/commits/f39d0c5fedd553fd3e1ae41400ac42c6d2688b90) isnt signed off.

- nit: there is only one blank line before the boundary test in test_sandbox_adapter.py (PEP 8 wants two between top-level functions).

harshnair75567-cloud:
@rajnisht7 thank you for your review can i ask you to take over this pr and complete it because tommorrow is 18 september and i cant be available for 2 days

lywinged (changes requested):
**Verdict: Request changes**, on @imran-siddique's ask of 15 September, which is still open at this head. Finding 1 below is a substantive one neither earlier review raised, and it is a recommendation rather than a second condition: its fix is two words and it is in the patch, but it hardens a path no caller in this repository reaches, so it is yours to take or to decline with a reason. Everything else is small, and the complete patch is at the end.

Reviewed at `f39d0c5`, fork point `6737005`, 18 commits behind `main` at `cd92e16`. Read first: #320 and #334, which closed it; @rajnisht7's review at `604a635` and his comment this morning; @imran-siddique's changes-requested review; `CONTRIBUTING.md` on sign-off.

The diagnosis is @harshnair75567-cloud's and it is right. I ran the whole chain rather than the end of it: `TrustRecord.model_validate()` accepts the record dict carrying `iat="1800000000"`, the dict still holds the string afterwards, `jsonschema` then rejects it with `'1800000000' is not of type 'integer'`, and `verify_record` on the signed form raises `record does not conform to the TRACE v0.2 schema`. Every step of the commit message's account is what happens, including the part about pydantic's lax mode coercing on a new object and never touching the dict being signed. One correction to the wording, since the substance survives it: the commit message says every other field reaches the record through a `models.py` constructor. Four members carry a value from the caller, and none of the four goes through one. `iat`, `subject`, `data_class` and `transparency` are written into the dict directly in both adapters; the six that do go through a constructor carry values the adapter assembled, and `eat_profile` and `cnf` are literals. What made `iat` the defect is not that it was written directly but that it was written unchecked, and on that the message is exactly right: the other three are each refused before they get there, `subject` by `_SUBJECT_RE` in the same `__post_init__` and the other two in the adapter's `__init__`. The floor is now the contract's, `1700000000`, matching `models.py:359` and the `minimum` in both schema copies, and @rajnisht7's mis-indented `image_digest` parenthesis is fixed.

## 1. Passing `__post_init__` is not a property of the value that gets signed

Neither session class is frozen. The check validates what the constructor received; `build_trust_record` reads `session.iat` later. One assignment in between reopens the hole and the record reaches the wire:

```
sandbox: session.iat = '1800000000' after __init__  ->  iat='1800000000'  schema: "'1800000000' is not of type 'integer'"
sandbox: session.iat = 1            after __init__  ->  iat=1             schema: "1 is less than the minimum of 1700000000"
agt:     session.iat = '1800000000' after __init__  ->  iat='1800000000'  schema: rejected, same message
```

**The honest size of this.** Nothing in this repository assigns to a session after constructing one. I swept every field of both dataclasses rather than the ones this pull request touches. Over the 70 Python files in `src`, `tests`, `docs` and `examples`, an AST walk counts 24 attribute assignments of any kind, so the instrument is not blind, and none of them is to a session field. So this is not a defect a caller meets today, and it is a smaller thing than the one the pull request fixes. It is also not quite the shape of #320: there, `issued_at` is a keyword argument of the documented call, so `build_record(issued_at="123")` reached the coercion directly. Here a caller has to reach back into a live object. What the two do share is the sentence #334 put in its own body, that a guard "can only speak about what the caller passed", and at this head the guard speaks about a value that can still change before it is used.

**Two words close the ordinary route, and the file already uses them.** `SandboxAttestation`, the class immediately above `SandboxSessionResult` in the same module, is already `@dataclass(frozen=True)`. Doing the same to both session classes closes plain reassignment, and closes it wider than `iat`: `sandbox_id` and `image_digest` are validated in the same `__post_init__` and have the same gap.

**It does not close the route, and I checked rather than assuming it did.** `frozen=True` guards `__setattr__` and nothing else. On a tree with it applied:

```
session.iat = '1800000000'                  FrozenInstanceError
vars(session)['iat'] = '1800000000'         reaches the wire, schema rejects
object.__setattr__(session, 'iat', ...)     reaches the wire, schema rejects
```

So the two words raise the cost of the accident and do not establish the property. The thing that does is a check where the value is read, which is what #334 did and what this pull request's own title says: *validate iat before it reaches the record*. Neither is a security boundary, since the caller is inside the process; the question is only which mistakes get caught, and the second catches them whatever route they arrive by.

```python
@dataclass(frozen=True)
class SandboxSessionResult:
```

**So the patch carries both**, and each half is load-bearing on its own: reverting only the two words leaves **8 failed, 1575 passed**, and reverting only the point-of-use call leaves **8 failed, 1575 passed**. With both, all three routes above are refused and the suite is **1583 passed, 1 skipped**.

I checked what freezing costs rather than assuming: `dataclasses.replace`, `dataclasses.asdict`, `copy.copy`, `copy.deepcopy`, a pickle round-trip and `__eq__` all still work, nothing in the repository subclasses either class, and `hash()` on a `SandboxSessionResult` fails before and after for the same reason, the unhashable `decisions` list, so that is not a behaviour this changes. `del session.iat` becomes a `FrozenInstanceError`, which nothing here does. The one cost is that freezing is a behaviour change for any caller outside this repository that assigns to a session after building one. Both classes are in `agentrust_trace.__all__`, not only the submodule's, so they are the package's public surface and whether the repository wants that change is a maintainer's call rather than a reviewer's. If the answer is no, drop the two words and keep the rest: the point-of-use half alone closes every route for `iat`, changes no class's shape, and is the half that matches the title.

## 2. The boundary tests never build a record, which is the ask that is still open

@imran-siddique on 15 September: "Add rejection tests for those three values and positive tests at both exact boundaries, **validating the resulting records against the schema**." @rajnisht7 said the same this morning. At `f39d0c5` both boundary tests stop at the constructor:

```python
@pytest.mark.parametrize("iat", [1700000000, JCS_SAFE_INTEGER])
def test_boundary_iat_values_are_accepted(iat) -> None:
    """The exact contract bounds from models.py must construct, not just values
    inside them."""
    assert _make_session(iat=iat).iat == iat
```

That is `tests/test_sandbox_adapter.py`; the AGT one is the same assertion over two lines with a different docstring.

The rejection half did land: `0`, `1` and `1699999999` are in both parametrize lists and all three refuse. What is missing is the half that exercises the path this pull request is about, and I checked whether some other test already covers it rather than assuming not. Nothing in either file varies `iat` and then validates the record: the three calls to `validate_json` in `tests/test_sandbox_adapter.py` are at lines 434, 441 and 448 and none of them mentions `iat`, and `tests/test_agt_adapter.py` does not call `validate_json` anywhere. So at this head no test in the AGT file puts a record in front of the schema at all. @rajnisht7 also suggested adding `TrustRecord.model_validate(record)` beside `validate_json`. The patch below does not, and his own comment says why: `model_validate` alone is not enough, because it is the call that gave the false pass this pull request exists to close. Against a record the schema already accepts it can only agree, so it would be a second assertion that cannot fail on its own. Both boundary records do validate today, so this is a coverage gap rather than a defect:

```
sandbox iat=1700000000  ACCEPTED    agt iat=1700000000  ACCEPTED
sandbox iat=9007199254740991  ACCEPTED    agt iat=9007199254740991  ACCEPTED
```

## Smaller

**The second commit is not signed off.** `f39d0c5` carries no `Signed-off-by`; `604a635` does. `CONTRIBUTING.md` requires it on all commits and no workflow checks it, so an approval is the only thing standing between it and a merge. The body's checklist marks that line `- [.]` rather than `- [x]`, so the body and the commits agree with each other. @rajnisht7 raised this this morning.

**`TRACE_MIN_IAT` is a third and fourth copy of a literal, and the same expression imports the other bound.** `1700000000` is a bare literal at `models.py:237` and `:359`, and now twice more in the adapters, with a comment in each asserting they agree and nothing testing it: nothing under `tests/` names `TRACE_MIN_IAT`. `JCS_SAFE_INTEGER` in the same condition is imported from `models.py`, so one bound tracks its source and the other does not. The patch below adds the test rather than a fifth constant, which keeps the change inside the files you already touch.

**The AGT test is named for the superseded bound.** `test_iat_must_be_a_bounded_non_negative_integer` at `tests/test_agt_adapter.py:300`; the sandbox one already reads correctly. @rajnisht7's point.

**The comment names the packaged schema copy rather than the source**, which @rajnisht7 also raised this morning. I checked the value rather than assuming: `schema/trace-claim.json` and `src/agentrust_trace/schema/trace-v0.2.json` both carry `"minimum": 1700000000`, and at this head the two paths are the same git blob, so it is a pointer to fix and not a wrong number. One thing I found next to it that is not yours and that I will send separately rather than add to your list: `tests/test_validate.py::test_packaged_schema_matches_the_normative_schema` compares the two as parsed JSON and says in its own docstring that it declines a byte comparison because the files differ in line endings. They are the same blob today, so the docstring has gone stale under the test it explains.

**The pull request body still describes the fix that was replaced.** It says the change mirrors "`_check_structure`'s bound (non-negative, <= JCS_SAFE_INTEGER)", true of `604a635` and not of this head, whose floor is `1700000000`. It also reports `1561 passed, 1 skipped`; this head is `1573`. The body is the document a reviewer reads first and the one surface a commit cannot reach, so it needs an edit from whoever carries the pull request from here.

**Blank lines, and what the linter does and does not say.** One blank line before `class TraceAGTAdapter` and three before `class TraceSandboxAdapter`, where PEP 8 asks for two, and one before the sandbox boundary test. CI runs `ruff check src tests scripts` at `.github/workflows/ci.yml:37`, and it passes on this head, so none of these fails CI.

## The patch

Against `f39d0c5`. Four files, +159 / -34, and it carries findings 1 and 2, the test name, the comment path and the blank lines. It is mine, written and run here rather than reviewed by anyone else, so read it before you take it. If you do take it, commit it with `git commit -s`: the sign-off finding 3 is about applies to this as much as to anything else you write. On it: `git apply` against a pristine `f39d0c5` is clean, `ruff` clean, `mypy --strict` clean on both adapters, **1583 passed, 1 skipped**. Reverting either half alone leaves **8 failed, 1575 passed**, so neither is decoration and the tests are not vacuous. The sign-off and the body edit are not in it, because a patch cannot do either; they belong to whoever carries this from here, which as of this morning may be @rajnisht7 rather than @harshnair75567-cloud.

```diff
diff --git a/src/agentrust_trace/adapters/agt.py b/src/agentrust_trace/adapters/agt.py
index 78b2fac..6ae59f6 100644
--- a/src/agentrust_trace/adapters/agt.py
+++ b/src/agentrust_trace/adapters/agt.py
@@ -23,12 +23,33 @@ from agentrust_trace.models import (
 )
 
 # TrustRecord.iat is Field(ge=1700000000) in models.py, matching "minimum" in
-# schema/trace-v0.2.json. Below it a record is schema-invalid, which is the whole
+# schema/trace-claim.json. Below it a record is schema-invalid, which is the whole
 # failure this check exists to stop -- so the floor is the contract's, not zero.
 TRACE_MIN_IAT = 1700000000
 
 
-@dataclass
+def _check_iat(value: object) -> None:
+    """The bound ``TrustRecord.iat`` carries, applied to whatever is about to be signed.
+
+    Called from ``__post_init__`` and again where the value is written into the
+    record. ``frozen=True`` stops the ordinary reassignment; it does not stop
+    ``vars()`` or ``object.__setattr__``, and a guard can only speak about the value
+    it is shown, so the one that matters runs where the record is built. #320 settled
+    the same point for ``provenance.build_record``.
+    """
+    if (
+        not isinstance(value, int)
+        or isinstance(value, bool)
+        or value < TRACE_MIN_IAT
+        or value > JCS_SAFE_INTEGER
+    ):
+        raise ValueError(
+            f"iat must be an integer Unix timestamp within the TRACE v0.2 range "
+            f"[{TRACE_MIN_IAT}, {JCS_SAFE_INTEGER}], got {value!r}"
+        )
+
+
+@dataclass(frozen=True)
 class AGTSessionResult:
     """Minimal AGT session data needed to build a Level 0 TRACE Trust Record.
 
@@ -65,16 +86,8 @@ class AGTSessionResult:
     """Issuance timestamp. Defaults to now."""
 
     def __post_init__(self) -> None:
-        if (
-            not isinstance(self.iat, int)
-            or isinstance(self.iat, bool)
-            or self.iat < TRACE_MIN_IAT
-            or self.iat > JCS_SAFE_INTEGER
-        ):
-            raise ValueError(
-                f"iat must be an integer Unix timestamp within the TRACE v0.2 range "
-                f"[{TRACE_MIN_IAT}, {JCS_SAFE_INTEGER}], got {self.iat!r}"
-            )
+        _check_iat(self.iat)
+
 
 class TraceAGTAdapter:
     """Build Level 0 TRACE Trust Records from AGT govern() session output.
@@ -167,6 +180,8 @@ class TraceAGTAdapter:
             else len(session.audit_entries)
         )
 
+        _check_iat(session.iat)
+
         record: dict[str, Any] = {
             "eat_profile": "tag:agentrust-io.com,2026:trace-v0.2",
             "iat": session.iat,
diff --git a/src/agentrust_trace/adapters/sandbox.py b/src/agentrust_trace/adapters/sandbox.py
index f9535fe..5cae8d3 100644
--- a/src/agentrust_trace/adapters/sandbox.py
+++ b/src/agentrust_trace/adapters/sandbox.py
@@ -89,10 +89,31 @@ from agentrust_trace.models import (
     ToolTranscript,
 )
 # TrustRecord.iat is Field(ge=1700000000) in models.py, matching "minimum" in
-# schema/trace-v0.2.json. Below it a record is schema-invalid, which is the whole
+# schema/trace-claim.json. Below it a record is schema-invalid, which is the whole
 # failure this check exists to stop -- so the floor is the contract's, not zero.
 TRACE_MIN_IAT = 1700000000
 
+
+def _check_iat(value: object) -> None:
+    """The bound ``TrustRecord.iat`` carries, applied to whatever is about to be signed.
+
+    Called from ``__post_init__`` and again where the value is written into the
+    record. ``frozen=True`` stops the ordinary reassignment; it does not stop
+    ``vars()`` or ``object.__setattr__``, and a guard can only speak about the value
+    it is shown, so the one that matters runs where the record is built. #320 settled
+    the same point for ``provenance.build_record``.
+    """
+    if (
+        not isinstance(value, int)
+        or isinstance(value, bool)
+        or value < TRACE_MIN_IAT
+        or value > JCS_SAFE_INTEGER
+    ):
+        raise ValueError(
+            f"iat must be an integer Unix timestamp within the TRACE v0.2 range "
+            f"[{TRACE_MIN_IAT}, {JCS_SAFE_INTEGER}], got {value!r}"
+        )
+
 __all__ = ["SandboxAttestation", "SandboxSessionResult", "TraceSandboxAdapter"]
 
 # Read the accepted platforms off the model so this cannot drift from the schema when a
@@ -182,7 +203,7 @@ class SandboxAttestation:
             )
 
 
-@dataclass
+@dataclass(frozen=True)
 class SandboxSessionResult:
     """One sandbox session, as the runtime has it at close.
 
@@ -233,17 +254,7 @@ class SandboxSessionResult:
             raise ValueError(
                 f"image_digest {self.image_digest!r} must be a sha256: or sha384: digest."
             )
-        if (
-            not isinstance(self.iat, int)
-            or isinstance(self.iat, bool)
-            or self.iat < TRACE_MIN_IAT
-            or self.iat > JCS_SAFE_INTEGER
-        ):
-            raise ValueError(
-                f"iat must be an integer Unix timestamp within the TRACE v0.2 range "
-                f"[{TRACE_MIN_IAT}, {JCS_SAFE_INTEGER}], got {self.iat!r}"
-            )
-
+        _check_iat(self.iat)
 
 
 class TraceSandboxAdapter:
@@ -332,6 +343,8 @@ class TraceSandboxAdapter:
         bundle_hash = self.bundle_hash(session.policy_bundle_bytes)
         runtime = self._runtime(session, bundle_hash)
 
+        _check_iat(session.iat)
+
         record: dict[str, Any] = {
             "eat_profile": "tag:agentrust-io.com,2026:trace-v0.2",
             "iat": session.iat,
diff --git a/tests/test_agt_adapter.py b/tests/test_agt_adapter.py
index 1a4b19f..1ab37dd 100644
--- a/tests/test_agt_adapter.py
+++ b/tests/test_agt_adapter.py
@@ -2,6 +2,10 @@
 
 from __future__ import annotations
 
+import json
+from dataclasses import FrozenInstanceError
+from pathlib import Path
+
 import hashlib
 import rfc8785
 
@@ -10,7 +14,11 @@ from pydantic import ValidationError
 
 from agentrust_trace import TrustRecord, sign_record, generate_key, key_to_jwk, verify_record
 from agentrust_trace.adapters import AGTSessionResult, TraceAGTAdapter
+from agentrust_trace.adapters.agt import TRACE_MIN_IAT
 from agentrust_trace.models import JCS_SAFE_INTEGER
+from agentrust_trace.validate import validate_json
+
+SCHEMA_PATH = Path(__file__).resolve().parents[1] / "schema" / "trace-claim.json"
 
 # ---------------------------------------------------------------------------
 # Fixtures
@@ -297,7 +305,7 @@ def test_an_unappraised_record_still_signs_and_verifies() -> None:
     "iat",
     ["1800000000", True, False, -5, 0, 1, 1699999999, 1.5, 2**60, JCS_SAFE_INTEGER + 1],
 )
-def test_iat_must_be_a_bounded_non_negative_integer(iat) -> None:
+def test_iat_must_be_within_the_trace_v0_2_range(iat) -> None:
     """Every other field here reaches the record through a models.py pydantic
     constructor, which coerces or refuses a bad type before it is dumped. `iat`
     used to reach build_trust_record's top-level "iat" key untouched, with
@@ -319,7 +327,48 @@ def test_valid_iat_round_trips_as_an_int_on_the_wire() -> None:
 
 
 @pytest.mark.parametrize("iat", [1700000000, JCS_SAFE_INTEGER])
-def test_boundary_iat_values_are_accepted(iat) -> None:
-    """Exact minimum and maximum from the TRACE v0.2 range should construct fine."""
-    session = _make_session(iat=iat)
-    assert session.iat == iat
+def test_boundary_iat_values_reach_the_wire_and_validate(iat) -> None:
+    """The exact contract bounds must survive to the record, not merely construct.
+
+    Constructing the session says nothing about the value that gets signed, which
+    is what this file's iat tests are about, so the assertion is on the wire form
+    and on the schema."""
+    record = _make_adapter().build_trust_record(_make_session(iat=iat))
+    assert record["iat"] == iat
+    assert isinstance(record["iat"], int) and not isinstance(record["iat"], bool)
+    validate_json(record)
+
+
+@pytest.mark.parametrize("iat", ["1800000000", 1, 1699999999, JCS_SAFE_INTEGER + 1])
+def test_a_checked_iat_cannot_be_replaced_before_the_record_is_built(iat) -> None:
+    """Passing __post_init__ has to be a property of the value that gets signed.
+
+    While the session was a mutable dataclass it was not: one assignment between
+    construction and build_trust_record put any of these on the wire, where the
+    schema then rejected the record. #320 settled the same point for
+    provenance.build_record, which is why its check sits at the point of use."""
+    with pytest.raises(FrozenInstanceError):
+        _make_session().iat = iat
+    # frozen=True guards __setattr__ and nothing else. Each of these reaches the
+    # field, so the check that decides what gets signed is the one in
+    # build_trust_record. A fresh session per route, or the second assertion would
+    # pass on damage the first one did.
+    for reach in (lambda s: vars(s).__setitem__("iat", iat),
+                  lambda s: object.__setattr__(s, "iat", iat)):
+        session = _make_session()
+        reach(session)
+        assert session.iat == iat
+        with pytest.raises(ValueError, match="iat must be an integer Unix timestamp"):
+            _make_adapter().build_trust_record(session)
+
+
+def test_the_adapter_floor_is_the_one_the_record_contract_carries() -> None:
+    """TRACE_MIN_IAT is a literal here and in models.py. If they ever diverge the
+    adapter refuses records the schema accepts, or emits ones it rejects."""
+    schema = json.loads(SCHEMA_PATH.read_text(encoding="utf-8"))
+    assert schema["properties"]["iat"]["minimum"] == TRACE_MIN_IAT
+    assert schema["properties"]["iat"]["maximum"] == JCS_SAFE_INTEGER
+    bound = next(
+        m for m in TrustRecord.model_fields["iat"].metadata if getattr(m, "ge", None) is not None
+    )
+    assert bound.ge == TRACE_MIN_IAT
diff --git a/tests/test_sandbox_adapter.py b/tests/test_sandbox_adapter.py
index 348ebce..2797be4 100644
--- a/tests/test_sandbox_adapter.py
+++ b/tests/test_sandbox_adapter.py
@@ -2,6 +2,9 @@
 
 from __future__ import annotations
 
+from dataclasses import FrozenInstanceError
+from pathlib import Path
+
 import hashlib
 import json
 from typing import Any, get_args
@@ -25,8 +28,11 @@ from agentrust_trace.adapters import (
     SandboxSessionResult,
     TraceSandboxAdapter,
 )
+from agentrust_trace.adapters.sandbox import TRACE_MIN_IAT
 from agentrust_trace.models import JCS_SAFE_INTEGER
 
+SCHEMA_PATH = Path(__file__).resolve().parents[1] / "schema" / "trace-claim.json"
+
 # ---------------------------------------------------------------------------
 # Fixtures
 # ---------------------------------------------------------------------------
@@ -341,13 +347,55 @@ def test_valid_iat_round_trips_as_an_int_on_the_wire() -> None:
     assert record["iat"] == 1800000000
     assert isinstance(record["iat"], int)
 
+
 @pytest.mark.parametrize("iat", [1700000000, JCS_SAFE_INTEGER])
-def test_boundary_iat_values_are_accepted(iat) -> None:
-    """The exact contract bounds from models.py must construct, not just values
-    inside them."""
-    assert _make_session(iat=iat).iat == iat
+def test_boundary_iat_values_reach_the_wire_and_validate(iat) -> None:
+    """The exact contract bounds must survive to the record, not merely construct.
+
+    Constructing the session says nothing about the value that gets signed, which
+    is what this file's iat tests are about, so the assertion is on the wire form
+    and on the schema."""
+    record = _make_adapter().build_trust_record(_make_session(iat=iat))
+    assert record["iat"] == iat
+    assert isinstance(record["iat"], int) and not isinstance(record["iat"], bool)
+    validate_json(record)
 
 
+
+@pytest.mark.parametrize("iat", ["1800000000", 1, 1699999999, JCS_SAFE_INTEGER + 1])
+def test_a_checked_iat_cannot_be_replaced_before_the_record_is_built(iat) -> None:
+    """Passing __post_init__ has to be a property of the value that gets signed.
+
+    While the session was a mutable dataclass it was not: one assignment between
+    construction and build_trust_record put any of these on the wire, where the
+    schema then rejected the record. #320 settled the same point for
+    provenance.build_record, which is why its check sits at the point of use."""
+    with pytest.raises(FrozenInstanceError):
+        _make_session().iat = iat
+    # frozen=True guards __setattr__ and nothing else. Each of these reaches the
+    # field, so the check that decides what gets signed is the one in
+    # build_trust_record. A fresh session per route, or the second assertion would
+    # pass on damage the first one did.
+    for reach in (lambda s: vars(s).__setitem__("iat", iat),
+                  lambda s: object.__setattr__(s, "iat", iat)):
+        session = _make_session()
+        reach(session)
+        assert session.iat == iat
+        with pytest.raises(ValueError, match="iat must be an integer Unix timestamp"):
+            _make_adapter().build_trust_record(session)
+
+
+def test_the_adapter_floor_is_the_one_the_record_contract_carries() -> None:
+    """TRACE_MIN_IAT is a literal here and in models.py. If they ever diverge the
+    adapter refuses records the schema accepts, or emits ones it rejects."""
+    schema = json.loads(SCHEMA_PATH.read_text(encoding="utf-8"))
+    assert schema["properties"]["iat"]["minimum"] == TRACE_MIN_IAT
+    assert schema["properties"]["iat"]["maximum"] == JCS_SAFE_INTEGER
+    bound = next(
+        m for m in TrustRecord.model_fields["iat"].metadata if getattr(m, "ge", None) is not None
+    )
+    assert bound.ge == TRACE_MIN_IAT
+
 # ---------------------------------------------------------------------------
 # 6. Transcript hashing uses JCS
 # ---------------------------------------------------------------------------
```

## Checked and clean

`pytest` 1573 passed, 1 skipped at this head. `ruff check src tests` clean and `mypy --strict` clean on both changed source files. The four CI checks on `f39d0c5` are green except `gate`, which waits on a maintainer approval at the current head rather than reporting a defect. The rejection parametrize covers `"1800000000"`, `True`, `False`, `-5`, `0`, `1`, `1699999999`, `1.5`, `2**60` and `JCS_SAFE_INTEGER + 1` in both files, and all twenty cases raise. `subject`, `data_class` and `transparency` are the other three members these adapters write without a pydantic constructor in front of them, and sweeping all three for the same bypass is where this review was wrong until late: `iat` is not the only one written unchecked, it is the one this pull request fixes. `subject` is held to `_SUBJECT_RE` in `SandboxSessionResult.__post_init__` and to nothing at all in `AGTSessionResult`, where the schema's `subject` pattern catches a bad value afterwards. `transparency` has no check either and the schema catches it, on type and on `format: uri`.

`data_class` is the one worth a separate issue, which I will file rather than add to this pull request. It is checked nowhere in either adapter, and `schema/trace-claim.json` gives it `"examples"` rather than an `enum`, so nothing catches it at all:

```
TraceSandboxAdapter(data_class="")           record validates
TraceSandboxAdapter(data_class="bogus-xyz")  record validates
```

That is a wider hole than the one this pull request closes, since a bad `iat` was at least refused by the schema in the end. It is out of scope here and it is not yours. `True` assigned after construction does not reach the wire even at this head, because `build_trust_record` passes `session.iat` into `Appraisal(timestamp=...)` and that field rejects a bool; a string and an integer below the floor have no second line of defence.

Of the 18 commits on `main` since the fork point, one touches these four files: #329 edits a docstring in `tests/test_sandbox_adapter.py` that this branch does not touch. `git merge-tree` against `cd92e16` is conflict-free.

Not approving, on 2.

harshnair75567-cloud:
@imran-siddique  Following up on timing: if this is needed today, I'll hand the PR over to @rajnisht7 to complete, since he's already reviewed it and is familiar with the remaining items (boundary validation, regression tests, DCO sign-off, and the PR body update). If tomorrow (September 19) is acceptable, I can complete and push the full fix myself. Let me know which you'd prefer.

lywinged:
> @imran-siddique  Following up on timing: if this is needed today, I'll hand the PR over to @rajnisht7 to complete, since he's already reviewed it and is familiar with the remaining items (boundary validation, regression tests, DCO sign-off, and the PR body update). If tomorrow (September 19) is acceptable, I can complete and push the full fix myself. Let me know which you'd prefer.

Didn't you notice I already did it for you?

harshnair75567-cloud:
@lywinged Sorry ,I was following this over email notifications and the diff didn't render clearly there — just pulled it up on GitHub directly. Will Apply it today along with the PR body update and sign-off, will push by 4:30 PM UTC today.

harshnair75567-cloud:
@lywinged Thank you for the detailed review and the patch — applied it cleanly on top of f39d0c5 as 9496ef9 (signed off), and confirmed locally: 1583 passed, 1 skipped, ruff clean, mypy --strict clean on both adapters. Pushed to fix/iat-type-bypass-in-adapters. Updating the PR body now to reflect the current test count and sign-off status.

imran-siddique (approved):
Approving at `9496ef9`. Both adapters now refuse an `iat` outside `[1700000000, JCS_SAFE_INTEGER]`, the rejection tests cover 0, 1 and 1699999999, and both exact boundaries build a record that passes `validate_json`. Merged locally with current main: the adapter tests pass and nothing else changes. `f39d0c5` carries no sign-off; the squash commit will carry yours from the other two. Thanks to @lywinged and @rajnisht7 for the reviews that set the bound.

imran-siddique:
@lywinged, your patch is applied as `9496ef9` and I have approved it. Your changes-requested review is the last thing blocking the merge. Could you re-review by 22 September?

lywinged (approved):
Approved at `9496ef9`. Everything here was recomputed on that head rather than taken from the thread.

`git diff f39d0c5 9496ef9` is the patch I posted, byte for byte, four files, +159 / -34, and nothing else. On it: **1583 passed, 1 skipped**, `ruff check src tests` clean, `mypy --strict` clean on both adapters, which is what you reported. Both halves are still load-bearing alone: dropping `frozen=True` gives `8 failed, 1575 passed`, and dropping the two `_check_iat(session.iat)` calls gives the same `8 failed, 1575 passed`, the 8 being `test_a_checked_iat_cannot_be_replaced_before_the_record_is_built` over its four parameters in each adapter's test file. Merged with current `main` at `56c9de0` the suite is **1695 passed, 1 skipped** and ruff is clean, so nothing that landed since has moved under it.

Two things are still open, neither of them code and neither worth holding the merge for, so I am recording them rather than blocking on them.

`f39d0c5` still carries no `Signed-off-by`; `604a635` and `9496ef9` both do. The body's checklist now reads `- [x] DCO sign-off on all commits`, so the body and the commits disagree where before they agreed, and nothing under `.github/workflows/` checks either. Squash merge settles it for `main`: #313 went in as four commits and arrived as one, so the sign-off that lands is the one written on the merge, which makes this @imran-siddique's line rather than a fourth commit here.

The body still says the change mirrors `_check_structure`'s bound "(non-negative, <= JCS_SAFE_INTEGER)". That was true of `604a635`. This head's floor is `TRACE_MIN_IAT`, 1700000000, which is the whole point of the second commit, so the paragraph now describes a version that is not the one being merged.

One nit that is mine and not yours: the patch left one blank line before `__all__` in `sandbox.py` and a three line run before the replace test in `tests/test_sandbox_adapter.py`. `ruff check src tests scripts` passes on both, so it is cosmetic. Worth a second if you are in the file again.

Thank you for the quick turnaround, and for applying the patch as its own signed commit instead of folding it into the earlier two. It is why the check above is one command.

---

lywinged (approved):
Approved at `9496ef9`. Everything here was recomputed on that head rather than taken from the thread.

`git diff f39d0c5 9496ef9` is the patch I posted, byte for byte, four files, +159 / -34, and nothing else. On it: **1583 passed, 1 skipped**, `ruff check src tests` clean, `mypy --strict` clean on both adapters, which is what you reported. Both halves are still load-bearing alone: dropping `frozen=True` gives `8 failed, 1575 passed`, and dropping the two `_check_iat(session.iat)` calls gives the same `8 failed, 1575 passed`, the 8 being `test_a_checked_iat_cannot_be_replaced_before_the_record_is_built` over its four parameters in each adapter's test file. Merged with current `main` at `56c9de0` the suite is **1695 passed, 1 skipped** and ruff is clean, so nothing that landed since has moved under it.

Two things are still open, neither of them code and neither worth holding the merge for, so I am recording them rather than blocking on them.

`f39d0c5` still carries no `Signed-off-by`; `604a635` and `9496ef9` both do. The body's checklist now reads `- [x] DCO sign-off on all commits`, so the body and the commits disagree where before they agreed, and nothing under `.github/workflows/` checks either. Squash merge settles it for `main`: #313 went in as four commits and arrived as one, so the sign-off that lands is the one written on the merge, which makes this @imran-siddique's line rather than a fourth commit here.

The body still says the change mirrors `_check_structure`'s bound "(non-negative, <= JCS_SAFE_INTEGER)". That was true of `604a635`. This head's floor is `TRACE_MIN_IAT`, 1700000000, which is the whole point of the second commit, so the paragraph now describes a version that is not the one being merged.

One nit that is mine and not yours: the patch left one blank line before `__all__` in `sandbox.py` and a three line run before the replace test in `tests/test_sandbox_adapter.py`. `ruff check src tests scripts` passes on both, so it is cosmetic. Worth a second if you are in the file again.

Thank you for the quick turnaround, and for applying the patch as its own signed commit instead of folding it into the earlier two. It is why the check above is one command.

harshnair75567-cloud:
@lywinged Thank you for putting together the patch itself, not just flagging the gaps — I was tied up and couldn't get to it myself in time, so having it ready to apply directly made this a lot faster. Appreciate the thoroughness on the verification too (both halves being load-bearing, the merge check against main). I'll get the PR body updated .

rajnisht7:
The core fix looks good floor is 1700000000, bool rejected, check runs in __post_init__ and again in build_trust_record, sessions are frozen, and the boundary tests go through build_trust_record + validate_json. Matches the schema and TrustRecord.iat.
One thing commit f39d0c5 is not Signed-off. CONTRIBUTING requires DCO on every commit. Kindly sign that one.

harshnair75567-cloud:
@rajnisht7 Thanks for the review. On the sign-off — @imran-siddique and @lywinged already covered this above: f39d0c5 itself has no Signed-off-by, but since this will land as a squash merge (like #313), the sign-off that ends up on main is the one written on the merge commit, which carries forward from 604a635 and 9496ef9. So no extra commit needed here on my end — that's on imran's side at merge time.
```> Squash merge settles it for `main`: #313 went in as four commits and arrived as one, so the sign-off that lands is the one written on the merge, which makes this @imran-siddique's line rather than a fourth commit here.```

rajnisht7 (approved):

PR #345, opened 2026-09-14
docs(governance): record the executed LF Projects contribution agreement

imran-siddique:
GOVERNANCE.md and CHARTER.md both say the Project Contribution Agreement with LF Projects, LLC is still being executed. It was executed on 17 August 2026, and both files now say so. The sentence saying the technical home is not yet settled comes out of both.

CHARTER.md still describes the Technical Charter itself as not yet in effect. Only the agreement status changes there.

This amends GOVERNANCE.md, so under "Amendments" it stays open for the 14-day minimum comment period and needs LF Projects, LLC approval before it merges. Opened 14 September 2026, so the earliest merge is 28 September.

Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01CreQ6VB7fSoPVf67dBvxCw


lywinged (commented):
The status update is right and I have no question about the date. One thing about the removal,
and then two smaller ones.

**The paragraph being deleted names two documents, and only one of them is now executed.** In
full, as it stands in GOVERNANCE.md: "The technical home for ongoing specification work is not
yet settled. Formation paperwork is also still in progress: the Technical Charter and Project
Contribution Agreement are being executed with LF Projects, LLC." The agreement is executed, as
this PR records. About the charter's execution the PR says nothing, and the only thing either
file now says about it is that it "has not yet taken effect", which is a different fact: a
document can be executed and not yet effective. Both files still say governance transitions to
a Technical Steering Committee when it does. So half the paragraph's condition is settled, the
other half is not stated either way, and the whole paragraph comes out.

That is what I would like a sentence about, because there are two quite different versions of
this PR. If the Technical Charter is also executed and only awaits effect, say so, and the
removal follows. If it is not, then what settles where ongoing specification work lives is the
charter and the TSC it creates, neither of which is in effect, and the sentence saying so is
still describing the situation accurately. The PR records one of the two pieces and removes the
statement covering both.

**CHARTER.md's status line no longer says anything about the charter's own execution.** You note
in the PR that CHARTER.md still describes the charter as not yet in effect and that only the
agreement status changes there, so this is not something you missed, and it is close to the same
point. The line used to read "the Technical Charter and Project Contribution Agreement are still
being executed" and now names only the agreement. That is the same distinction as above, and this is
where a clause would fix it: if the charter is executed and only awaits effect, the status line
can say both things, and if it is not, it can say that.

**A question rather than a finding, about which rule this runs under.** Your note cites
GOVERNANCE.md's Amendments section, and the 14-day minimum, the LF Projects approval and the
28 September arithmetic all check out against it. CHARTER.md has its own rule at section 9:
"Amendments to this charter require a two-thirds TSC majority and a 30-day public comment
period. Before host organization acceptance, amendments require Project Lead approval and
14-day notice to contributors."  I do not know whether you read a status record
as an amendment of that file at all, and if the answer is no then none of this applies and it is
worth one line in the PR saying so. If the answer is yes, the exception appears to have lapsed
with the acceptance the file records, and there is no TSC yet.

One smaller thing, which supports the first: `spec/trace-v0.2.md` section 6.1 still reads
"Formation is in progress". Whether that is consistent after this PR depends on the same answer.

imran-siddique:
@lywinged the Technical Charter is executed and has not taken effect. 0e7f5d7 says so in GOVERNANCE.md, the CHARTER.md status line, and spec §6.1, so the sentence no longer drops half its condition.

On §9: CHARTER.md is not in effect, and GOVERNANCE.md names itself the authority until it is. So this runs under GOVERNANCE.md's Amendments rule, and the 28 September date holds. §9 applies once the TSC exists.

Qiang-Xu (approved):

Issue #206, opened 2026-08-24
test(conformance): portable PIC/TRACE bridge corpus and independent verifier


coreytshaffer:
**Spec section affected**

- `docs/integration/pic-trace-bridge-v1.md`
- Bridge verifier tests and the eventual conformance-corpus location

**Problem**

Issue #179 delivered the PIC/TRACE authorization bridge design, schema, reference verifier, and eight implementation tests. Its completion note records a narrower remaining gap: those tests use inline fixtures and are not a portable vector set that a second implementation can consume.

TRACE already has a general cross-language RFC 8785/JCS corpus from PR #126. The remaining gap is bridge-specific:

- Can another verifier reproduce the bridge-specific signed bytes and digests while using PR #126 for shared canonicalizer conformance?
- Does it reach the same result at trust-key, time-window, scope, digest, and transcript boundaries?
- Does each negative case fail for the intended rule rather than an unrelated earlier error?

**Proposed change**

This would be a non-normative, informative conformance addition. I propose a versioned JSON corpus for `pic-trace-bridge-v1`, an independent verifier, and a differential runner.

The case contract would contain:

- bridge artifact and trusted-authorizer JWK;
- declaration and PIC verifier outputs;
- tool call and optional transcript;
- fixed verification time;
- expected semantic result and stable reason code; and
- bridge-specific canonical bytes and reproducibility metadata where relevant.

Vector families would cover:

1. valid bound execution;
2. malformed and unknown-field rejection;
3. signature, trust-key, and `authorizer_key_id` failures;
4. deny, not-yet-valid, invalid-window, and expiry behavior;
5. tool and impact scope mismatch;
6. PIC, declaration, and tool-call digest mismatch;
7. missing, partial, or contradictory transcript evidence; and
8. cross-language canonicalization by reference to PR #126, with new signed-object fixtures only where the bridge introduces an additional input or binding.

The independent verifier would not import `agentrust_trace.intent_bridge`. The runner would execute the corpus against both implementations and report agreement or a preserved ambiguity. A bounded mutation check would verify that each negative vector makes its named rule load-bearing.

I would appreciate maintainer guidance on whether the corpus should live beside the bridge in `trace-spec`, in `trace-tests`, or initially in a standalone repository until the case contract and outcomes are reviewed.

**Backward compatibility**

- [x] Non-breaking (informative conformance corpus and tooling)
- [ ] Breaking

[...]

PR #362, opened 2026-09-15
docs(integration): cite the PIC specification from the PIC/TRACE bridge

### LINKED ISSUE #361: docs: cite the authoritative PIC specification from the PIC/TRACE bridge
**What is wrong**

The PIC/TRACE bridge names `PIC-CJSON/1.0` and preserves PIC-defined `intent_digest` and `args_digest` values, but it does not cite the upstream PIC specification or repository.

That leaves the bridge under-specified for independent implementers. A reader can see that the bridge consumes PIC-shaped values, but cannot tell from this repository where `PIC-CJSON/1.0`, `intent_digest`, `args_digest`, or the relevant PIC declaration semantics are defined.

This does not need to make PIC a normative dependency of TRACE. The bridge can remain optional and informative. But if the bridge uses PIC-defined profile names and digest semantics, it should identify the source of authority for those terms.

**Spec section or file**

`docs/integration/pic-trace-bridge-v1.md`

Related schema path:

`schema/pic-trace-bridge-v1.json`

`#/properties/authorization/properties/pic`

The schema currently contains:

```json
"profile": {
  "const": "PIC-CJSON/1.0"
}
```

and requires:

```json
"intent_digest"
"args_digest"
```

without pointing readers to the PIC specification that defines those values.

**Expected behavior**

The bridge documentation should cite the upstream PIC specification as the source for `PIC-CJSON/1.0`, `intent_digest`, `args_digest`, and the relevant PIC declaration semantics.

Suggested minimal addition to `docs/integration/pic-trace-bridge-v1.md`:

```md
## References

- PIC Standard: https://github.com/pic-standard/pic-standard
- PIC-CJSON/1.0: canonicalization and digest profile used for PIC `intent_digest` and `args_digest`
```

Equivalent wording would also be fine if maintainers prefer to place the reference elsewhere.

This keeps the current boundary intact: TRACE does not depend on PIC, and the bridge remains optional, but the optional bridge references the external profile it consumes.

**Impact**

This does not appear to cause invalid Trust Records or cause verifiers to accept invalid ones directly.

The impact is interoperability and authority/discovery. Independent implementers cannot determine from the TRACE repository where the PIC profile and digest semantics are defined, so they may treat the bridge document as the de facto definition of `PIC-CJSON/1.0` or implement those fields inconsistently.

Adding an upstream reference would make the bridge clearer, preserve composability, and avoid TRACE becoming the implicit source of truth for PIC-defined semantics.

lywinged:
## What this changes

Closes #361.

`docs/integration/pic-trace-bridge-v1.md` names `PIC-CJSON/1.0` and preserves PIC's
`intent_digest` and `args_digest`, and cites nothing for any of the three. A reader of this
repository can only take the bridge document as the definition of terms it does not own, which is
what #361 reports. Two additions to the document, and a test.

**A `References` section**, answering the block the issue drafted. It names the PIC Standard
repository and PIC Canonical JSON v1, and cites `docs/canonicalization.md` alongside the repository
root because section 8 is where the byte rules live: section 8.1 gives
`args_digest = SHA-256(canonicalize(action.args))`, section 8.3 gives
`intent_digest = SHA-256(UTF-8 bytes of intent string)` with no JSON wrapping, and each is "the
64-character lowercase hex encoding of the SHA-256 output". It also names PIC's `docs/vocabulary.md`,
which is the artifact behind the issue's fourth term: that file asks downstream specifications to
reference its definitions "rather than recoining the same concepts", and sources each entry. The
section says what this bridge does with those semantics rather than implying it binds them. It
compares `declaration.impact` against the signed `scope.impacts`, and the schema makes both
unconstrained strings, so the vocabulary is PIC's and an integrator aligning with it takes it from
PIC. The repository's own fixture uses `external-side-effect`, which is not in PIC's taxonomy, so
claiming otherwise would have been checkable and wrong.

**A subsection on the representation boundary**, `PIC digest values at the bridge boundary`, which
answers the second half of the issue's Impact section, implementers getting the fields
inconsistently. It states the two digests are computed differently, since a document written to
stop readers getting PIC's digest semantics wrong should not flatten the case PIC special-cases.
PIC's values are bare hexadecimal. This bridge's schema constrains both fields through its
`$defs/digest`, `^sha256:[0-9a-f]{64}$`, so an adapter serializes the PIC verifier's output as
`sha256:<pic_hex>`, on the artifact and on the arguments to the reference implementation alike. A
bare value is refused with `authorization.pic.intent_digest must be a sha256 digest` on the first
and `intent_digest must be a sha256 digest` on the second. The wording follows the four points the
reporter set out on the issue after agreeing to this change.

**A test pinning the quotations**, `tests/test_pic_bridge_doc_matches_the_code.py`. The new prose
quotes a schema pattern and two strings the code raises, and nothing in the suite read that
document, so the quotations could drift with everything green. Each of the three was mutated in the
document and fails exactly one of the six tests.

The fixture-contract half discussed on #206 is not in this change.

One process note, since I am a maintainer here. `require-maintainer-approval.yml` returns early for
a maintainer-authored pull request, so this one can merge on CI alone. I am not merging it: it
changes a document another member asked about, and it should have a reviewer.

## Type of change

- [ ] Schema change
- [ ] Example addition

Ticked Editorial as citations and a clarification in a document that already binds nothing. Say so
if you would rather it took the non-breaking path and its comment period.

## Spec section

None. `spec/trace-v0.2.md` is untouched, and so is every schema. The change is
`docs/integration/pic-trace-bridge-v1.md`, one test, and one `CHANGELOG.md` entry. PIC stays a
non-dependency: the bridge consumes the values and now cites their source.

## Checklist


imran-siddique (approved):
Checked on `625a606`. The references match PIC’s canonicalization sections 8.1 and 8.3: arguments use canonical JSON bytes, while intent uses raw UTF-8 bytes. The documentation correctly distinguishes those digest definitions from TRACE’s `sha256:<hex>` representation.

The clarification matches the scope agreed in #361. No runtime behavior or schema changes. The six new documentation tests and 47 existing bridge tests pass locally; CI and CodeQL pass on this head. DCO is present. Approved.

PR #363, opened 2026-09-16
Publish PIC/TRACE bridge corpus with case-003 successor reconciliation

coreytshaffer:
This is the **initial publication of the previously held PIC/TRACE bridge corpus plus the reviewed case-003 successor reconciliation**. It makes the held fixtures, contracts, generator, independent verifier, and scoped reconciliation tests available for repository review.

Related: #206 and the [September 15 PIC maintainer resolution](https://github.com/agentrust-io/trace-spec/issues/206#issuecomment-5682052917), following the successor-envelope model accepted in #338 and implemented in #340.

## Publication scope

The repository base is `cd92e16dde6658eedbe1058c903f3fce92bb3a58`. This PR has one commit and **118 added files, 0 modified files, and 0 deleted files**. All 338 existing upstream Git entries, including the runtime and normative bridge schema, are preserved exactly. The 19 upstream files whose archived checkout representation used different line endings retain their original Git bytes.

The held corpus had not previously been committed to this repository. Consequently, initial publication adds its existing material alongside the reconciliation. The reviewed **16-path reconciliation patch is a delta against the preserved held corpus**, not the complete repository diff. Its bytes and SHA-256 remain unchanged. The repository diff has its own SHA-256 below.

The 101 other fixture files, the held v1 schema, and the held global reason-code registry retain their reviewed bytes. Their publication does not represent broader semantic reconciliation or certification.

## Case-003 reconciliation

Only case 003 and its direct counterexamples opt into the revised three-axis contract:

- `pic_authorization_bound`: supplied PIC-defined verifier outputs match the signed authorization.
- `successor_envelope_bound`: the complete successor envelope matches the signed digest.
- `successor_outcome`: independently evaluated as `established`, `contradicted`, or `not-established`.

The positive vector contains a complete relevant envelope and establishes the fixture transition. Literal `after: {}` reaches the exact missing-envelope-fields refusal (`transcript_after_invalid`) with earlier prerequisites valid; no successor outcome is evaluated. Valid digest-bound envelopes containing empty or unrelated observations remain accepted at the binding layer and produce `not-established` outcomes.

PIC `allow` remains a pre-execution authorization result. It does not establish execution, success, or a successor state. PIC canonicalization and digest semantics are consumed as upstream verifier outputs, not redefined here.

The old `transcript_bound` positive credit remains **withdrawn**. The frozen conformance denominator remains **100**; rows **101/102 remain excluded diagnostics**. **Full-corpus certification remains held**, and the default full-corpus runner explicitly refuses certification. This PR does not broaden semantics beyond the reviewed checkout.

## Verification and hashes

Fresh verification on the publication checkout passed **95/95 scoped checks**, with zero failures, errors, or skips:

```text
python -m pytest tests/test_pic_bridge_case003.py tests/test_intent_bridge.py tests/test_successor_observation.py
```

This includes regeneration of all 105 fixtures from an empty fixture directory, all 101 frozen fixture SHA-256 comparisons, exact reason checks, and tests that demonstrate the intended guards affect the result. Both case-003 verifier paths agree. All 118 added files match the reviewed checkout byte-for-byte.

- Reviewed reconciliation patch SHA-256: `c7de8b3621aea7d85155a736005f6ed5fd7558d89176aed603ea8dfe439892bf`
- Complete repository-diff SHA-256: `17cb8501b537f98772cfa7610a27d2001a7966b17735a21d16518b8b7f5bbe08`
- Publication commit: `900fa7405099ad5fb8c359a2bc3c2e6708824a6d`

The repository hash identifies the 724,104-byte output of this command, separate from GitHub's rendered or downloadable patch formats:

```text
git -c diff.algorithm=myers -c core.quotepath=false diff --binary --full-index --no-ext-diff --no-textconv --no-renames --unified=3 --src-prefix=a/ --dst-prefix=b/ cd92e16dde6658eedbe1058c903f3fce92bb3a58 900fa7405099ad5fb8c359a2bc3c2e6708824a6d --
```

Maintainers may modify the branch. This PR is ready for review; no merge or auto-merge is requested.


madeinplutofabio:
Thanks for publishing this. I reviewed the PIC-facing fixture contract and the case-003 reconciliation, not the full 118-file TRACE corpus.

This is the fixture-contract review I offered in #206, now that #362 has handled the PIC-source reference and bridge-boundary documentation.

From the PIC side, the case-003 direction looks right.

The important boundary from #206 is preserved:

- PIC `allow` remains a pre-execution authorization result only.
- PIC verifier outputs are consumed by the bridge, not recomputed or redefined by TRACE.
- Successor-envelope binding is separate from successor/outcome sufficiency.
- Literal `after: {}` is treated as a malformed/missing successor-envelope refusal, not as bound successor evidence.
- Complete but insufficient successor envelopes can be digest-bound and still evaluate to `not-established`.

The 003 / 003a / 003b / 003c split looks correct to me.

I recommend two PIC-facing fixture-contract changes.

First, please make the digest-boundary wording explicit in the corpus contract or README, not only in the bridge documentation from #362. In native PIC, `intent_digest` and `args_digest` are bare 64 lowercase hex SHA-256 values. `args_digest` is computed over PIC-CJSON canonical bytes of `proposal.action.args`; `intent_digest` is computed over the raw UTF-8 bytes of `proposal.intent`.

In this corpus, the fields named `pic_intent_digest` and `pic_args_digest` are TRACE-bridge serialized PIC verifier outputs, because the bridge surface uses `sha256:<hex>`. That prefix is a TRACE bridge representation rule, not a change to PIC semantics.

Related to that, I recommend adding explicit bare-hex boundary vectors. The existing signed/supplied syntax cases cover invalid prefixed values such as uppercase hex under `sha256:<hex>`, which is useful, but they do not pin the specific #361 interop trap: a native PIC verifier output pasted verbatim as bare 64 lowercase hex into the TRACE bridge surface.

Suggested cases:

- signed `authorization.pic.intent_digest` is bare 64 lowercase hex, expected `pic_intent_digest_invalid`;
- supplied `pic_intent_digest` argument is bare 64 lowercase hex, expected `pic_intent_digest_invalid`;
- signed `authorization.pic.args_digest` is bare 64 lowercase hex, expected `pic_args_digest_invalid`;
- supplied `pic_args_digest` argument is bare 64 lowercase hex, expected `pic_args_digest_invalid`.

Those would make the #362 boundary testable: native PIC emits `<pic_hex>`, TRACE bridge carries `sha256:<pic_hex>`, and the bridge must not silently accept or reinterpret bare PIC-native values on its own surface.

Second, `transcript_after_invalid` now covers more than "`after` is not an object" in the case-003 path. Literal `after: {}` is an object, but it is not a valid successor-observation envelope because it lacks `observation`, `observer`, and `observed_at`. I recommend either widening the reason-code description or adding a very explicit note in the case-003 contract that this code is reused for malformed successor envelopes, including missing successor-envelope fields.

These are not objections to the case-003 reconciliation. They are fixture-contract clarity issues for independent implementers, especially around the PIC boundary fixed in #362.

lywinged (changes requested):
Reviewed at `900fa740`, base `cd92e16`. I read your PR body, both markdown files in the diff, and @madeinplutofabio's comment before writing, and I re-derived every number below from a clean checkout.

Three things block. The first is not the one either of us would have guessed.

## 1. The corpus is pinned to a reference 66 commits behind this PR's base, and three of the changes since then have moved under it

`run_bridge_vectors.py:45` sets `REFERENCE_BASELINE = "3c2f9637"`. Between that commit and `cd92e16` there are 66 commits, six of which touch `src/agentrust_trace/intent_bridge.py`: #305, #319/#322, #333, #325, #340, #318. Three of the six are visible in the corpus right now.

**#340 is the big one.** `schema/pic-trace-bridge-v1.json` gained `if transcript_required is true then successor_observation_digest is required`. Validating each `inputs.bridge` against that file:

```
fixtures carrying expected.bridge_schema : 105
agree                                    : 59
disagree                                 : 46      (all declare "valid", all measure "invalid")
```

Every one of the 46 has exactly one error, at `properties/authorization/allOf/0/then/required`. Your README defines that axis by relative path into this repository, so it names the live schema, and on 46 fixtures the declared verdict and the live schema disagree.

**#305 and #319/#322 account for two more, and the second is not cosmetic.** `082-decision-unknown.json` expects classification `denied` with `decision_not_allow`. The current reference raises `authorization.decision must be "allow" or "deny"`, which is `invalid`, not `denied`. That is a classification change, not a missing mapping: a corpus whose whole premise is that rejection for the wrong reason does not count now carries a vector that is refused for a different reason than the one it names. `016-signature-decoder-rejected.json` is the milder case, an exception now wrapped so the adapter's exact-type check no longer fires.

Running the corpus through its own `_frozen_run_corpus` gives 102 cases and 50 failures: 35 bridge-schema alone, 11 bridge-schema plus reference, 3 reference alone, 1 case-schema plus reference. **The independent verifier disagrees on none of them.** Your independent implementation is not what moved.

**One more thing the drift has already made false, and it is on fabio's exact subject.** `examples/pic-trace-bridge-v1/README.md:74` says "transcript `after` is checked only for object shape, so an empty object is accepted". Your own `003a-empty-after.json` carries `after: {}` and expects `transcript_after_invalid`, and the runtime refuses it. That sentence was true against `3c2f9637` and is false against your base, in the corpus's front-door document.

You have a defence and it belongs next to this rather than found later: `CASE003-REVIEW.md` says "Wider corpus reconciliation, commit, push, and PR publication are outside this patch", which is a fair description of what you set out to do. **I am not asking you to reconcile 46 fixtures here.** I am asking for three things that are paragraphs rather than patches: say in the README that the corpus is scored against `REFERENCE_BASELINE` and not against the current head, delete the `after: {}` sentence at line 74, and put the number in the PR body. `CASE003-REVIEW.md` opens by saying the patch is against `cd92e16`, and the baseline constant says otherwise in six places, including a `const` in both contract schemas.

## 2. The suite fails on this head, and the fix is independent of finding 1

`tests/test_adequacy_all_sets.py::test_every_vector_set_on_disk_is_measured_somewhere`:

```
AssertionError: vector sets with nothing grading them: ['pic-trace-bridge-v1'].
Add a loader to SETS, or name the test that covers the set in MEASURED_ELSEWHERE.
A set in neither is not adequate or inadequate, it is unmeasured.
```

Full run at this head: 1 failed, 1677 passed, 1 skipped. The same file on `cd92e16` is 13 passed. This is the only thing in the review that fails CI on every push, and it is caused entirely by this diff.

It names two remedies and the first works on this head, with finding 1 wholly unresolved. The adequacy criteria are decidable from the fixtures alone and never run a verifier, so the 46 disagreements are irrelevant to them. I built the loader and ran it on the unmodified head:

```
vectors                  : 105
trivially_satisfied_by   : None     (no unconditional answer passes the set)
boundaries without margin: none     (every code boundary carried by two or more vectors)
```

One trap worth handing you: `tests/adequacy.py` has `ACCEPTING = frozenset({"accept", "verified", "pass"})`, so a loader passing your `accepted` through unmapped reports `trivially_satisfied_by = "reject"`. Map it and it is clean, with no `KNOWN_THIN` entry needed.

The other branch is the one to avoid. A `MEASURED_ELSEWHERE` entry naming `tests/test_pic_bridge_case003.py` would not be true: that test grades 4 of the 105, and of the rest one harness compares hashes, another compares regenerated bytes, and `run_corpus()` is called once in the whole suite, inside `pytest.raises`.

**Related, and worth fixing in the same pass.** `CASE_PATTERN = "[0-9][0-9][0-9]-*.json"` does not match `003a`, `003b` or `003c`. So the full-corpus runner sees 102 of the 105, and the three it cannot see are exactly the counterexamples this PR exists to add. Re-enabling it later, which is where finding 1 points, would grade the corpus while silently dropping three of the four reconciliation vectors. That is the hand-maintained-list defect, in the runner rather than in the registry.

## 3. No DCO sign-off

The single commit `900fa740` carries no trailer block. `CONTRIBUTING.md` line 42 requires it and line 48 says a pull request without it will not be merged. No workflow checks it, so it falls to review. `git commit --amend -s` fixes it.

## A question rather than a blocker: one code crosses a boundary you set

`CASE003-REVIEW.md` says "The reason labels in the case-003 adapter are review-local normalization, not additional entries in the approved global taxonomy." The rest of the PR keeps that rule.

`case003_independent.binding()` returns two code strings, `transcript_after_invalid` and `successor_envelope_digest_mismatch`, and `verify_bridge_independent.py:482` puts both into `codes`. The first is one of the approved 45; the second is in no registry. `tests/test_pic_bridge_case003.py:111` asserts `result["codes"] == ["successor_envelope_digest_mismatch"]` across three mutations, so it is not a latent path.

I started to recommend moving it into `successor_reason` and then checked that against your own contract. `README.md:97` says non-accepted results carry null "and at least one exact code", so moving it leaves a `mismatch` with `codes: []` and breaks a different rule in the same document. The string is also emitted at `case003_independent.py:30`, matched at `:76`, hardcoded again in the reference adapter at `case003_runner.py:77`, and asserted in the test: four edits across three files, not one line.

So it is a real question with two answers I cannot pick for you: register a 46th code, which is a taxonomy change and the Project Lead's, or move it and say what `codes` carries instead. It is not blocking because neither answer is yours alone to give.

The same shape, unreachable and worth knowing: `case003_runner.REASONS` maps 8 upstream reasons and `evaluate_successor_observation` can return a ninth, `"successor observation is absent"`, on `after is None`. The adapter would raise `ReferenceSignalError`. `contract/case003-v1.schema.json` closes `id` to the four vectors that exist, so nothing can reach it.

## @madeinplutofabio's two recommendations

**The second is already implemented and I should say so first.** You offered widening the description *or* a note in the case-003 contract. `CASE003-REVIEW.md:45` already reads "`transcript_after_invalid` is reused for the exact missing-envelope-fields refusal, anchored to exception type and full message." That is the second option, in the diff. If the registry description is also worth widening, the minimal change is one clause on its first sentence naming the case-003 path: the three fixtures carrying that code are `{}` (003a), `null` (099) and `[]` (100), and only the first is an object.

**The first is right and the gap is real, and two of the four proposed cases need one more step.** Bare 64-hex appears at no JSON path in any of the 105 fixtures. The two supplied digest fields appear in all 105 and the two signed ones in the 99 that carry `authorization.pic`; the only invalid-syntax coverage is one uppercase case for each of the four. It is worth pinning because `docs/integration/pic-trace-bridge-v1.md:57`, from #362, states that a bare hex value is refused, and #361 raised the interop question that document answers.

Run through both verifiers:

| case | reference | independent |
|---|---|---|
| supplied `pic_intent_digest` bare hex | `pic_intent_digest_invalid` | `pic_intent_digest_invalid` |
| supplied `pic_args_digest` bare hex | `pic_args_digest_invalid` | `pic_args_digest_invalid` |
| signed `authorization.pic.intent_digest` bare hex | `authorization_unverifiable` | `authorization_unverifiable` |
| signed `authorization.pic.args_digest` bare hex | `authorization_unverifiable` | `authorization_unverifiable` |

The supplied pair is as expected. The signed pair is not: editing inside the signed authorization invalidates the Ed25519 signature, so the case fails there before the digest-syntax rule is reached, which is a vector passing for an unrelated earlier error. Re-signing with `gen_bridge_vectors._resign`, which `tests/test_pic_bridge_case003.py:152` already uses, gives `pic_intent_digest_invalid` and `pic_args_digest_invalid`, both verifiers agreeing on all four. One note for when they land: the registry entry for `pic_intent_digest_invalid` already reads "Signed or PIC-verifier-supplied intent digest has invalid SHA-256 syntax", so the new vectors share a code with the uppercase ones and their discriminating value is on the bridge-schema axis.

## What I checked and could not fault

The scoped command gives 95 passed, matching your 95/95. Your diff command reproduces to the byte: 724,104 and `17cb8501…bbe08`, exactly as written, and dropping `--full-index` gives 716,316 and a different digest, so the flag in your command block is load-bearing rather than incidental. 118 files, all additions, zero modified, zero deleted; 338 entries at base. The generator regenerates byte-identically from an emptied directory and is collected as `test_generator_reproduces_its_committed_fixtures[pic-trace-bridge-v1]`. No build artifacts. The `known_defect` treatment is what #206 asked for. Independence holds transitively: neither `verify_bridge_independent.py` nor `case003_independent.py` imports `agentrust_trace`, and the latter is the only repository module the former pulls in.

On the `codes` axis the discipline is exactly right: 45 used across the 105, 45 registered, equal in both directions.

And a correction in your favour. CI runs `ruff check src tests scripts`, `tools/check_dashes.py` and `mypy src/agentrust_trace`, all clean, but none of them looks at 117 of your 118 added files, which are under `examples/`. So I ran ruff over the corpus myself: also clean, and there is not one em dash in the 118 files. Nothing in CI would have told you that.

## Smaller things

`case003_runner.py` lines 165 and 166 emit `frozen_conformance_denominator: 100` and `known_defect_rows_excluded` as literals, and `test_four_declared_targets_agree` asserts four of the report's seven fields, leaving those two and `scope` unchecked. 100 is derivable, as 102 loaded minus the two known-defect rows; it is just not derived.

`run_bridge_vectors.py` carries 322 unreachable lines across 13 functions out of 749, counting everything reachable only from `_frozen_run_corpus` and `_frozen_mutation_report`, which have no callers at all. `score_case` alone is 104 of them. That is the grader for the whole corpus, disabled.

Four fixtures declare a case-schema verdict they do not meet. The README says every committed case records `valid` for `expected.case_schema` against `contract/case-v1.schema.json`; the four case-003 vectors record `valid` and validate against `case003-v1.schema.json` instead. Created by this PR, and a one-sentence fix.

`successor_outcome` is enumerated `established`, `contradicted`, `not-established`, and the string `contradicted` appears in none of the 105 fixtures. `tests/test_pic_bridge_case003.py::test_contradiction_requires_affirmative_relevant_evidence` does assert the outcome, so it is exercised; what nothing pins is the reason it carries. A fifth case-003 vector means editing the `id` enum, so this is a recommendation on #338 requirement 3 and #279 rule 1 rather than an ask.

The README names `tests/test_pic_trace_bridge_vectors.py` and `tests/test_pic_trace_bridge_completeness.py`; neither exists on any of the 437 refs in my clone. And it says the bare `run_bridge_vectors.py` scores every case, where it exits 2 with `FATAL RuntimeError: Full-corpus certification remains held; run --case-003 for the bounded review`, which is correct behaviour, documented in the file's own header block, and stale in the body.

## Not yours

Measured 2026-09-16 16:22Z, `900fa740` has four workflow runs: `CI` and `CodeQL` both `action_required`, so neither has executed your code, and `Contributor reputation check` and `Require maintainer approval`, which are `pull_request_target` workflows running the base branch. Releasing the runs is a maintainer action. It is also why finding 2 was invisible on both sides: your three-file command cannot reach `tests/test_adequacy_all_sets.py`, and `ci.yml` runs the whole suite under coverage with no path filter, so it would have.

## Where this leaves it

The publication mechanics are the most carefully evidenced I have reviewed here, the diff hash reproduces to the byte, and your independent verifier disagrees with nothing in the corpus. Finding 2 is one loader and it is the one failing CI. Finding 3 is one flag. Finding 1 is the one that changes what merging means, and the ask there is three paragraphs and a deleted sentence, not 46 fixtures: what lands should say what it is scored against.

The loader, so it is not an ask:

```python
def pic_trace_bridge() -> list[Vector]:
    return _load("pic-trace-bridge-v1",
                 lambda e: "accept" if e["conformant_runtime"]["classification"] == "accepted"
                           else e["conformant_runtime"]["classification"],
                 lambda e: list(e["conformant_runtime"]["codes"]))
```

with `"pic-trace-bridge-v1": (pic_trace_bridge, None)` in `SETS`. The four bare-hex vectors are the existing uppercase cases with the value changed to bare hex and `gen_bridge_vectors._resign(inputs)` called on the two signed ones.

madeinplutofabio:
Thanks @lywinged, this answers my PIC-facing review points.

Agreed on the refinement for the bare-hex vectors: the supplied-value cases should reach `pic_intent_digest_invalid` / `pic_args_digest_invalid` directly, while the signed `authorization.pic.*` cases need to be re-signed after changing the value, otherwise the earlier signature failure is the thing being tested.

That is exactly the boundary I wanted pinned: native PIC emits bare 64 lowercase hex, the TRACE bridge surface carries `sha256:<pic_hex>`, and a bare native PIC value must not be silently accepted or reinterpreted on the TRACE bridge surface.

On `transcript_after_invalid`, I agree the existing case-003 note covers the missing-envelope-fields reuse. If the registry wording is widened later, that is enough from my side.

I’ll leave the baseline, adequacy-loader, DCO, and full-corpus drift issues to the PR author and maintainers. Any assistance needed, please ping me and I'll be happy to oblige.

imran-siddique:
Ruling on the code question from @lywinged's review: register it as the 46th rule, named `transcript_after_mismatch`.

The registry already pairs every bound value this way. `pic_intent_digest`, `pic_args_digest`, `declaration_digest` and `tool_call_digest` each have `_invalid` and `_mismatch`, and the transcript has `transcript_before_mismatch`. The after side has only `transcript_after_invalid`, so a complete envelope whose digest differs from the signed one has no code of its own. That gap is the registry's, not the case-003 adapter's. The name follows the existing pattern rather than introducing `successor_envelope_digest_mismatch`.

@coreytshaffer the three blocking findings above stand as written: the baseline and the 46 fixtures, the adequacy loader and `CASE_PATTERN`, and the sign-off.

PR #371, opened 2026-09-18
docs(platforms): add the Bernstein software-only vendor annex (#231)

chernistry:
## What this changes

The vendor annex accepted on #231, sequenced after the reproducibility claim and opened now that #364 has merged. Informative, per GOVERNANCE: reviewed by the vendor author and one Maintainer, binds no implementation, and names the section behind every requirement it quotes (`tests/test_requirement_keywords_are_attributable.py` passes on it).

`docs/platforms/bernstein.md` maps a software-only producer, a deterministic scheduler with no hardware root, onto the Trust Record. Everything in it is pinned to `sipyourdrink-ltd/bernstein@4281e3c1`, the commit the supplementary-plane pair was minted at, and follows the emitter's own module docstring rather than restating it from memory:

- **What the producer records.** The hash-chained journal (`event_hash = H(prev_hash, event_type, payload_hash, index)`, timing excluded from `payload_hash`), the seal boundary the emitter states for itself, and one distinction a verifier needs: the journal's own hashes use a code-point key order with ASCII escaping and are not TRACE digests; every digest that reaches a record is RFC 8785.
- **TRACE representation.** A field-by-field table: value on the record, and the journal events it is derived from. Optional members omitted, never `null`.
- **Delegation and aggregate.** One record per hop, linked over the complete signed parent under section 3.1.3; the run-level aggregate's rollup rules and its `member-execution` references with the digest in `digest`, as corrected earlier in the thread. Both `rel` values are named as unregistered.
- **Two commitments.** Where `runtime.measurement` (all-zero, pending #242) and the journal commitment (`tool_transcript.hash`, over the calls; the head is not on the record) sit today.
- **The reproducibility claim.** The producer's `replay --re-derive` check is the function: recorded plan and outcomes in, coordination steps out, no agent re-executed, no network. The mapping onto `function` / `code_identity` / `code_resolver` / `input_closure` / `transcript_digest`, with two cautions: the re-derived journal head is not `transcript_digest` (different encoding), and the re-derivation's two refusal codes are both `diverged` in section 3.1.4's vocabulary. Stated plainly: the emitter at the pinned commit does not carry the block yet; the schema shape is #370.
- **What a verifier can check without the producer**, in four steps, and which of them need the journal.
- **Assurance boundary.** Level 0 by construction, with the three conformance checks that make it so, and this repository's own example having the same property.
- **Vectors.** The seven emitter-minted records, the supplementary-plane pair among them, with the mint and the reference-suite command.

`docs/platforms/index.md` gains a vendor-annex table with the one row; `mkdocs.yml` gains the nav entry under Platforms; `CHANGELOG.md` gains the entry.

## Type of change

- [ ] Schema change
- [ ] Example addition

## Spec section

None changed. The annex reads 3.1.1 (`software-only`), 3.1.2 (`references`), 3.1.3 (the link preimage), 3.1.4 (the claim) and 3.4 (delegation), and says so where it does. Section 4.4's annex list names hardware slots only; whether a software-only row belongs there is a spec question I have left alone.

## Checklist

- [x] `CHANGELOG.md` updated
- [ ] Breaking changes marked (none)
- [ ] Backward compatibility statement (not applicable: documentation only)

Docs guards, `tools/check_dashes.py` and the full suite (1655 passed) are clean on the branch.

Refs #231.


imran-siddique (dismissed):
Approving at the merge commit I just pushed. The annex is the one accepted on #231 and it stays inside what a software-only producer can claim: `software-only` on every record, Level 0 with `TR-RTE-001` named as the reason Level 1 is unreachable, and everything pinned to `sipyourdrink-ltd/bernstein@4281e3c1`. The nav entry is in `mkdocs.yml` and the requirement-attribution guard passes.

The `CHANGELOG.md` conflict came from merging #370 just now, so I resolved it in the branch with both entries kept.

PR #372, opened 2026-09-18
Keep the v0.3 draft schema canonical and document key-bound runtime evidence

### LINKED ISSUE #368: v0.3 draft schema drifted from canonical, missing cnf.jwk private-key guards from #296/#311
**What is wrong**

`schema/trace-claim-v0.3-draft.json` is supposed to be canonical plus two v0.3-only changes (`eat_profile` and `runtime.evidence`). It was actually forked from an older revision and never picked up:

- #296: refuse private-key members in `cnf.jwk` (`d`, `p`, `q`, …)
- #311: RSA branch (`n` / `e`)

Checked on `main` @ `cd92e16` with a v0.3 vector that adds `cnf.jwk.d`:

- draft schema → accepts it
- canonical `trace-claim.json` → rejects it

**Not a live bypass.** `verify_record()` already refuses any profile other than TRACE v0.2 before schema checks, so the shipped verifier never hits this file. Still worth fixing: the draft is what a v0.3 implementer reads, and it currently allows the private half of a confirmation key.

**Expected**

Regenerate the draft from canonical HEAD, keep only the two declared v0.3 diffs, and add a small drift test so this doesn't happen again. Suite stays green (existing vectors are OKP, no private members).

**Related**

#296, #311, #136, #247. Doesn't conflict with open PR #341.

imran-siddique:
The v0.3 draft had drifted from canonical key restrictions, while the runtime-evidence RFC still described the top grade as unreachable despite a later key-binding capture. This PR combines #372's schema generation with #373's bounded documentation update.

## Changes

- Generate the draft from the canonical schema, preserving only its declared profile and runtime-evidence differences. Include canonical private-key refusal, RSA handling and the reproducibility fields now on main.

- Check generated-byte drift, the allowed difference set and private-key rejection. Regenerate after merging current main, which exposed two failing drift assertions before regeneration.

- Record the September 14 TDX key-binding capture in RFC section 7.2 and correct its section 7.1 cross-reference.

- Use portable paths in the schema inventory test so Windows separators do not create false classification failures.

Both original PR histories are preserved. #373 remains the link to the original capture discussion. This changes neither the released verifier's supported profile set nor the evidence grade's trust assumptions.

## Validation

Combined head `b7f6305e702158415f6f94bd43844a4eca498518`: 77 focused schema/inventory tests pass; generator `--check`, Ruff, mypy (12 source files), house-style and diff checks pass. Full Windows suite: 1,911 passed, six skipped and five generator-fixture line-ending failures. Those same five failures reproduce on unchanged current main; Both Python 3.11/3.12 Linux CI runs and the runtime-evidence appraisal jobs passed on the combined head. All nine reported checks passed, and the exact-head Actions inventory contains no held or failed run.

The capture's previously recorded verification used the pinned external verifier and produced grade `attested`, model claim `absent`. The quote was checked against a pinned Intel root without TCB/QE identity appraisal; the record makes no model or policy claim. This is one historical capture outside the corpus, not a newly executed hardware experiment or a guarantee upgrade.

Closes #368. Original schema-drift report: @saintmalik. DCO sign-offs are retained. Auto-merge is disabled pending independent review.


imran-siddique:
@Qiang-Xu could you review the combined changes at `b7f6305e7021`? This now includes #373: the bounded key-binding capture update, alongside canonical draft-schema generation. I regenerated the draft against current main; the combined Linux CI passed. The title and description include the full scope and evidence limits.

imran-siddique:
@lywinged could you review the combined head, `b7f6305`? The main review points are the canonical-to-draft schema differences and the limits stated for the historical key-binding capture. All nine checks passed, and the current-head Actions inventory has no held or failed runs.

The capture does not establish model identity or full TCB/QE appraisal. The PR description preserves those limits and links the original discussions. Qiang is also requested for review.

lywinged (approved):
**Verdict: Approve** at `b7f6305`. Three things, in the order I would act on them, none of them
holding the merge:

1. **One sentence to change before it ships.** §7.2 says the capture put `sha256(cnf.jwk.x)` in
   `REPORT_DATA`. It put the digest of the bytes that string decodes to. Recomputed; the string
   reading matches nothing anywhere in the quote. It leads because it is in the diff and because
   it is now the only construction the RFC gives for the binding the whole top grade turns on.
   The correct version is already published in the capture program, so this is a one-line fix.
2. **A gap the PR neither creates nor closes**, offered as a suggestion rather than a condition:
   the published v0.3 capture record is schema-validated by nothing, in either repository, and
   `binds: "cnf-key"` is the one value of a five-value enum that no record here carries.
3. **One commit is not signed off**, `d9b8266`, while the description says they all are. Moot for
   `main` under squash merge, raised because the description makes the claim.

Everything else reproduced, including the parts I could only check by fetching the artifacts and
running the pinned external verifier. I am not holding the merge for any of them.

Read first: #368 and its report by @saintmalik, #373 and the capture it records, #296 and #311 for
the two rules the draft had lost, §4 rules 4 and 6, §5.2, §6, §7.1 and §8 of the RFC, and the
workflow at `.github/workflows/runtime-evidence.yml`.

## 1. `sha256(cnf.jwk.x)` is not what the capture computed, and this is now the only place that says how

§7.2 as written:

> On 2026-09-14 a GCP C3 trust domain generated an Ed25519 key, put `sha256(cnf.jwk.x)` followed by 32 zero bytes in `REPORT_DATA`

`cnf.jwk.x` is a base64url string. The digest in `REPORT_DATA` is over the 32 bytes that string
decodes to, not over the string. Recomputed against the published fixture:

```
sha256(cnf.jwk.x as ASCII)        == REPORT_DATA[0:32]   False
sha256(base64url-decode(x))       == REPORT_DATA[0:32]   True
REPORT_DATA found in the quote at offset 568, 64 bytes, last 32 all zero
```

This matters more than a wording nit because nothing else in the document says how. §5.2 settles
*what* to bind ("the record-signing key") and never the pre-image. §4 rule 6 says the guest field
is "compared against the record's `cnf` key". The table row at line 102 says `binds` records "what
the guest committed to". So after this change §7.2 line 223 is the only construction in the RFC,
and it is the one an implementer writing a second verifier will copy. Getting it wrong produces a
clean-looking failure: the quote verifies, the record verifies, and the binding check says no.

**The correct version is published, in code, in the directory §7.2 links to**, which is what
makes this a one-line fix rather than an open question. `gcp-tdx-2026-09-14-capture.py`:

```python
public = key.public_key().public_bytes(serialization.Encoding.Raw, serialization.PublicFormat.Raw)
report_data = hashlib.sha256(public).digest() + bytes(32)
```

`Raw, Raw` is the 32 key bytes. So a reader who opens the capture program gets the right answer
and a reader who stops at the RFC does not, and the RFC is the document with the normative
ambition.

I tried to kill this finding before reporting it rather than after. Six readings of
"`sha256(cnf.jwk.x)`" against the 8000 quote bytes: the ASCII string, the ASCII string with its
padding restored, UTF-16, the standard-alphabet decode, the url-safe decode, and the canonical
JSON of the whole `cnf.jwk` object. The two decodes match at offset 568 and are the same bytes as
each other. The other four produce a digest that appears nowhere in the quote at all, so this is
not a case of my having guessed one pre-image and missed another.

The charitable reading is that §7.2 is describing what one capture did rather than specifying a construction, and that is probably the intent. It does not help a reader: §5.2 settles what to bind and stops, so this sentence is the only thing in the document that looks like a construction, and it is what gets copied. In JOSE `x` is itself defined as the base64url encoding of the key, so "sha256 of `cnf.jwk.x`" reads most naturally as the digest of the member's value, which is the string, which is the reading that does not verify.

The paragraph already contains the right answer, four sentences later: **"The pre-image is the
public key, and it is in the record."** That sentence is correct and the formula contradicts it.
Suggested replacement for the clause:

> put `sha256` of the 32 raw public-key bytes, that is `base64url-decode(cnf.jwk.x)`, followed by 32 zero bytes in `REPORT_DATA`

**The same formula is in three places, and one of them cannot be edited later.** Besides §7.2, the
commit that introduced it, `d9b8266`, carries "binds sha256(cnf.jwk.x) in REPORT_DATA" in its own
message. Under squash merge the message that lands on `main` is written at merge time, so that one
is yours to get right in the squash body rather than in a follow-up.

One thing worth knowing while you are in there, and it is not yours to fix in this repository:
`tools/check-key-binding.py` in `agentrust-io.github.io` carries the same shorthand in its own
docstring, line 10, "REPORT_DATA[0:32] is SHA-256 of the record's cnf.jwk.x", while its code at
line 59 does `key = unb64u(jwk['x'])` and then hashes `key`. So two documents say the same
imprecise thing and one implementation does the right one. Two documents agreeing is one
instrument.

A second, smaller thing in the same neighbourhood, also not this PR's: that checker asserts
`parsed.report_data == sha256(key) + bytes(32)`, which requires the second 32 bytes to be zero.
§5.2 says the opposite in as many words: "the verifier does not derive a grade from them, and it
does not require them to be zero." The published checker is stricter than the profile it checks.
Noting it because §7.2 now points readers at that tool.

## 2. The only v0.3 record in existence is not a fixture, so the drift guard protects a schema nothing exercises

This is the thing the two halves of the PR do to each other, and I do not think it has been looked
at, because doing it needs a file from another repository. One half regenerates
`schema/trace-claim-v0.3-draft.json`. The other half points §7.2 at a published v0.3 record.
Nothing in the PR validates the second against the first.

Neither does the other side, and it says so in as many words. The capture's own checker,
`tools/check-key-binding.py`, opens by stating what it leaves to someone else: "It does not
validate the record against the TRACE v0.3 draft schema, which lives in trace-spec." So the tool
that publishes the record defers the schema check to this repository, and this repository does not
have the record. That is the whole of the gap, and both halves are written down. I checked the rest of that
repository too rather than taking the docstring's word for it: `check-tdx-verifier.mjs`,
`verify/verify-page.js` and `verify/home-panel.js` all reference the record and none of them
mentions a schema, `ajv`, or a validate call.

I did it by hand. `gcp-tdx-2026-09-14-keybind_record.json` from `agentrust-io.github.io` at
`f172f48c`, against both revisions of the draft:

```
capture record vs draft BEFORE this PR (b92fd02):  0 errors
capture record vs draft AFTER  this PR (b7f6305):  0 errors
```

So the regeneration does not break the one real artifact, which is the answer you want and which
the PR does not currently claim. The control that makes that mean something, the same record with
`cnf.jwk.d` added:

```
with cnf.jwk.d, before: ACCEPTED
with cnf.jwk.d, after:  REFUSED
```

That is #368's defect reproduced on a real v0.3 record rather than on a v0.2 example with its
profile string swapped, and the fix proven load-bearing on it.

The gap this leaves is narrower than I first wrote it, and I am stating the enumeration rather
than the category because my first attempt at this paragraph was false.
`tests/test_runtime_evidence_vectors.py::test_record_is_valid_against_the_draft_schema` already
validates all fourteen corpus records against the draft, and all fourteen declare the v0.3
profile, so the draft is exercised. What is not exercised is the shape §7.2 is about.
`evidence.binds` is a closed five-value enum in the draft, `cnf-key`, `weights-digest`,
`manifest-digest`, `nonce`, `unspecified`. Across the fourteen vectors it is absent eleven times,
`weights-digest` twice, and one vector carries no evidence block at all. `cnf-key` appears in no
record in this repository. The only record that carries it is the published capture, in another
repository, validated by nothing here.

So the drift guard covers the member, and the one value of that member which the whole top grade
turns on has no vector. Carrying the capture record, or a redacted copy, as a fixture would close
it. Not a condition of this merge; the guard is the right shape and this is the next thing it
wants. Separately, `tests/test_v03_draft_schema.py` builds its own record by taking
`examples/amd-sev-snp.json`, a v0.2 record, and setting `eat_profile` to the v0.3 identifier. That
is a reasonable stand-in and I checked it is not vacuous: the unmutated record validates with 0
errors against both schemas, so the private-member assertions really are about the private member.

## 3. One commit is not signed off, and the description says they all are

`d9b8266`, the RFC commit, carries no `Signed-off-by`. The other four do, all
`Imran Siddique <imran.siddique@opaque.co>`. The description says "DCO sign-offs are retained" and
`CONTRIBUTING.md` asks for one on every commit, so the body and the commits disagree.

Not blocking, and for the same reason it was not blocking on #343: `main` is squash merged, so the
sign-off that lands is the one written on the merge. Raising it because the body makes the claim
explicitly, and because nothing under `.github/workflows/` checks it.

## 4. Smaller

**The description understates what the generator preserves.** "Preserving only its declared
profile and runtime-evidence differences" is two members; `build()` also carries `$id`, `title` and
`description` across from the committed draft, which is five. The generator's own docstring lists
all five correctly and the draft's `description` field means record members by "members", so this
is a loose sentence rather than a wrong one. Worth a word because the allowed difference set is
the contract this PR is adding.

**The PR body cites itself.** "This PR combines #372's schema generation with #373's bounded
documentation update." #372 is this pull request. #368 is presumably meant.

**A normalisation in the new test cannot fire.** `test_committed_draft_is_what_the_generator_produces`
compares `committed.replace("\r\n", "\n")` against the generator's output. `Path.read_text` opens in
text mode, where universal newlines translate `\r\n` on read, so the string it returns does not
carry `\r\n`. I verified that on Linux by writing a CRLF file and reading it back; the Windows
case is the documented behaviour of text mode rather than something I ran, and I have no
Windows here. Harmless, and the part that is load-bearing on
Windows is `write_text(..., newline="\n")` in `main()`, which is right. Worth deleting so a reader
does not conclude the file might arrive with CRLF.

**The private-member parametrization is three of seven.** `d`, `p`, `q`, where
`_JWK_PRIVATE_PARAMS` carries `d`, `dp`, `dq`, `k`, `p`, `q`, `qi`. I checked all seven against the
regenerated draft and all seven are refused, and
`test_draft_differs_from_canonical_only_in_its_declared_members` covers the whole `cnf` block
structurally anyway, so the narrowness costs nothing today. Raising it only so the choice is a
choice.

## Checked and clean

Run here rather than read off the description.

**The head, `b7f6305`.** `pytest` 1916 passed, 6 skipped. `python scripts/gen_v03_draft_schema.py
--check` exit 0. `ruff check src tests scripts`, `mypy src/agentrust_trace` (12 files) and
`tools/check_dashes.py` all clean.

**Merged with current `main`, `892dbcf`**, which is three commits past this PR's base:
`--check` still exit 0, so the regeneration has not gone stale under the three commits that have
landed since: `882a0af` (verify_record reports whether cited objects resolve), `f5f3526` (the
Bernstein adopters entry) and `892dbcf` (freshness context with verification results). 1966
passed, 6 skipped; ruff clean; mypy clean on 13 files.

**"Two failing drift assertions before regeneration" reproduces exactly.** I put the draft back
to its pre-merge state, `d4cba53`'s copy, on an otherwise unchanged head: `2 failed, 3 passed`, and
the two are `test_committed_draft_is_what_the_generator_produces` and
`test_draft_differs_from_canonical_only_in_its_declared_members`. That is the description's number,
and it also proves both drift assertions are load-bearing rather than decorative.

**The regenerated draft really does carry what the description says it carries.** The RSA branch is
there with `required: ["n", "e"]`, the private-member refusal is there, and `reproducibility` and
`appraisal.method` are both present at top level, matching canonical.

**"77 focused schema/inventory tests" reproduces exactly.** `tests/test_v03_draft_schema.py` plus
`tests/test_safe_integer_range.py` is 77 passed.

**The §6 cross-reference correction is right.** §7.1 is where the middle grade's limit is shown:
`limit-substituted-quote-from-the-same-td` is the vector that demonstrates rule 4 binds a
measurement and not a quote. §7.2 is the top grade. The old pointer sent a reader to the wrong one.

**Every figure in the new §7.2, recomputed from the fixtures.** `gcp-tdx-2026-09-14-keybind_quote.bin`
is 8000 bytes with sha256 `2217b3d640b2e4cdabd34604ea59df7f4ea23ed9702d3ec040689dca20ce1d61`. The
record carries `runtime.evidence.binds: "cnf-key"`, `collateral: "embedded"`, `format:
"tdx-quote-v4"`, `model.provider: "none"`, `appraisal.status: "none"`, and `policy.bundle_hash`
`sha256:e3b0c442...b7852b855`, which is the SHA-256 of the empty string. `cnf.jwk` is OKP /
Ed25519. The quote embedded in the record is byte-identical to the published `.bin`, and the
published capture program hashes to the `capture_script_sha256` the fixture declares. The platform
string is "GCP C3 confidential VM, Intel TDX, configfs-tsm provider tdx_guest", so "a GCP C3 trust
domain" is supported.

**`appraise()` reproduces, with `agent-manifest` at the pinned
`934809709a2815695d65cfacb45dc0a164286046`.** On the capture record it returns `attested`, and
`grade_model_claim` returns `model claim: absent`, which is what §7.2 says. Running it on that
record is the part nothing in either repository does. The control that makes it a measurement
rather than a single observation is the same function over the repository's own fourteen corpus
vectors: `platform-attested` six times, `unattested` three times, `Reject` five times, and
`attested` **zero** times.

Credit where it belongs, because I nearly presented that sweep as a new fact and it is not:
`examples/runtime-evidence/test_appraisal.py::test_appraisal_matches_committed_expectation`
already runs `appraise()` over every vector and asserts both the grade and the model claim, and
`tests/test_runtime_evidence_vectors.py::test_no_vector_in_the_corpus_reaches_attested` already
asserts the `attested` count on the vectors' declared expectations. My sweep re-derives what those
two hold, which makes it a confirmation from a second direction rather than a discovery. Worth
knowing that the first of them is not in the default `pytest` run: `testpaths = ["tests"]`, so it
executes only in the `runtime-evidence` workflow.

My first two controls were worthless and I say so: mutating the record breaks its signature, so
both died at `record envelope failed: InvalidSignature` before reaching any grade logic, which is
why the control that counts is the corpus sweep rather than a mutation.

**The CI claim in §7.2 holds.** `.github/workflows/runtime-evidence.yml` pins `TDX_CAPTURE` to the
`gcp-tdx-2026-07-21` fixture directory, so the job does not touch the 09-14 capture, exactly as the
new text says.

**The generator tested where it cannot fire**, rather than where it does. Rename in canonical the
key that `evidence` follows and `evidence` still lands last rather than being dropped. Remove
`evidence` from the committed draft and `build()` raises `ValueError: 'evidence' is not in list`,
which is loud rather than silent. Both acceptable.

**The strongest thing in this PR is the second test, and it deserves saying.**
`test_draft_differs_from_canonical_only_in_its_declared_members` compares the two schemas to each
other, not the committed file to the generator's own output, so it is independent of `build()` and
it is what fails if the allowed difference set is ever widened. The first test pins bytes to the
generator and could not catch a wrong `build()` on its own. Having both is the right shape, and it
is a better answer to #368 than simply copying canonical over the draft once.

## What I could not check

Two figures in the description I have no route to: the Windows run, "1,911 passed, six skipped and
five generator-fixture line-ending failures", and the claim that those five reproduce on unchanged
current `main`. I have no Windows here. Everything I report above is Linux. Likewise "all nine
reported checks passed" is your reading of the Actions inventory and not mine; I read the workflow
files rather than the run list.

## What I did not do

No independent cold read of this review. The instruments above are mine and were chosen by me,
which is exactly what a cold read is for. Said rather than skipped.

PR #373, opened 2026-09-19
docs(rfc): record the 2026-09-14 key-binding capture in runtime evidence §7.2

imran-siddique:
§7.2 of the runtime evidence RFC said `attested` was unreachable because no capture bound a record-signing key. That stopped being true on 2026-09-14.

The capture is published at https://agentrust-io.com/verify/ with fixtures in https://github.com/agentrust-io/agentrust-io.github.io/tree/main/verify/fixtures. `gcp-tdx-2026-09-14-keybind_quote.bin` is an 8000 byte TDX v4 quote, sha256 `2217b3d640b2e4cdabd34604ea59df7f4ea23ed9702d3ec040689dca20ce1d61`. REPORT_DATA is sha256 of the record's `cnf.jwk.x` followed by 32 zero bytes.

What I ran:

- `python tools/check-key-binding.py` in agentrust-io.github.io at `7cba24d`: PASS (quote verifies, REPORT_DATA commits to the record key, record signature verifies).
- `appraise()` from `examples/runtime-evidence/generate.py` on the same record, with agent-manifest at the pinned `934809709a28`: grade `attested`, model claim `absent`.

The new text says what that grade does not cover. The quote is checked against the pinned Intel root only, with no TCB or QE identity appraisal, and the record says `appraisal.status: "none"`. It makes no model or policy claim (`policy.bundle_hash` is the empty-string sha256). It is not a corpus vector, and agent-manifest still binds the manifest digest, so the reference producer has not reached this grade.

Also fixes the §6 pointer to §7.1, which is where the middle grade's limit is actually shown.


imran-siddique:
Consolidated into #372 at b7f6305e702158415f6f94bd43844a4eca498518. The combined branch preserves this PR's complete head and history, and its title/description now cover both changes. Closing this duplicate as superseded; this is not a merge or a resolution of outstanding review requirements.

Issue #280, opened 2026-09-03
Records cite evidence by URI alone, so an appraisal cannot be reproduced against what was appraised


opento-suggestions:
**Spec section affected**
§3.1 (Logical schema), §3.2 (Wire format), §3.3 (Verification), and §4.3
(Bindings TRACE adds), as carried by `schema/trace-claim.json`.

**Problem**

## The observation

Four citation fields in the Trust Record carry a bare URI and no digest, so
what a record cites cannot be checked to be what was appraised.

At 5c69dc1679f866bdc6e997b782a91bb992458db1, `schema/trace-claim.json` carries
four citation fields that are a bare URI with no digest member beside them:
`appraisal.policy_ref` ("URI to the appraisal policy used."),
`transparency` (the SCITT receipt URI), `runtime.rim_uri`, and
`model.aibom_uri` (`appraisal.verifier` is also a bare URI but names the
verifier rather than an artifact, and is not counted). Three surfaces pair the URI with a
digest: `policy.policy_uri` beside `policy.bundle_hash`,
`tool_transcript.transcript_uri` beside `tool_transcript.hash`, and
`build_provenance.provenance_uri` beside `build_provenance.digest`; the
`references[].digest` member is optional, "when the producer holds it at issue
time". The four are top-level fields of the record, not entries in
`references`; the rules agentrust-io/trace-spec#226 quotes for that mechanism
do not reach them. `appraisal.policy_ref` appears in `spec/trace-v0.2.md` once, at line
241, as an example value.

The one bound object shows the spec already knows how: `policy.bundle_hash`
is required, patterned `^sha(256:[0-9a-f]{64}|384:[0-9a-f]{96})$`, described
as "SHA-256 or SHA-384 digest of the policy bundle in force at execution
time", and the spec prose states the binding in its own words (line 590):
"TRACE adds the binding: the policy bundle hash is sealed to the TEE
measurement". agentrust-io/trace-spec#276 (opened 2026-09-03) is doing the same work
for `delegation.parent_record_hash`, stating a preimage the schema alone left
open.

## The consequence

A second verifier resolving the same URI at appraisal-check time verifies
whatever is deployed at that moment, not what was appraised. Where
resolution fails outright, agentrust-io/trace-spec#190 and now
agentrust-io/trace-spec#279 govern what the verifier records; where it
succeeds, nothing at this head makes "the same object" checkable. The tree
states the mechanism for schemas in its own words:

[...]

PR #374, opened 2026-09-19
feat: verify_record reports whether the objects a record cites resolve

opento-suggestions:
## What this changes

A Trust Record cites objects it does not carry. `appraisal.policy_ref`, `runtime.rim_uri` and `model.aibom_uri` are URIs the schema constrains to `format: uri`, two of which the adapters write, and, until this change, nothing under `src/` read: a verifier could not say whether the object a record points at could still be obtained. agentrust-io/trace-spec#190 asks what a verifier records when a citation cannot be resolved, and where; agentrust-io/trace-spec#280 records that these fields carry no digest. This PR answers the first question at the result-type layer, the way agentrust-io/trace-spec#271 did for revocation, and leaves the second where it is. The resolver is called last, after the signature verifies and after every check that can raise, so a record that fails verification drives no resolution.

`verify_record()` gains `citation_resolver`, a caller-supplied function from URI to bytes, defaulting to `None`, and `VerificationResult` gains `citations`: one row per surface, in a fixed order, each a `CitationCheck` with `outcome`, `cause` and `evidence`. `resolved` carries the URI, the SHA-256 over exactly the bytes the resolver returned, and their count. `unresolvable` carries the cause and the exception's class name when the resolver raised, or the returned value's type name when it returned non-bytes, never any message text. `not_attempted` carries the reason: no resolver supplied, field absent, or surface deferred. `transparency` is deferred with its reason in the row: agentrust-io/trace-tests#92 holds the transparency-binding question, and section 3.1.2 refers the `transparency` reachability gap to section 7 open question 3.

What the check does not do is stated in the module docstring rather than implied. It records resolvability and asserts no binding: a `resolved` row says bytes were produced and hashed, not that the object was in force or that it binds the record, which is agentrust-io/trace-spec#280's question. It never reads `references[]`, which keeps it inside section 3.1.2 rule 3 by construction, and a signed record carrying a `references[]` entry that names a resolver verifies with `citations` byte-identical to the same record without it. It takes its resolver from the caller only, for the reason section 3.1.2 gives about TR-POL-003: a record that names its own checker can name one that agrees with it. No outcome changes the revocation check, the thumbprint, or whether verification raises; every revocation vector `verify_record()` accepted at `0ec04531` still verifies with the same revocation result, and a test holds that over the 23 it accepts. A resolver that is neither callable nor `None` is refused at entry beside the existing argument checks. The outcome names are not accepted normative text; agentrust-io/trace-spec#279, the thread naming that primitive across surfaces, is where they would be settled, and a consumer using other names is not thereby wrong.

Two calls this PR takes that agentrust-io/trace-spec#190 did not settle, offered for review rather than assumed. It consumes all three bare-URI surfaces rather than `policy_ref` alone; reversing that is contained: the surfaces tuple, its five vectors, and their README rows. And a `resolved` row retains the digest over the returned bytes, so a second verifier has a determinate object to compare against once the binding question is decided; reversing that is contained: the digest, its evidence key, its invariant, and the resolved vectors' expected blocks.

`examples/citation-resolution/` carries sixteen generated vectors, the generator that writes them, and a README. Each is a signed v0.2 record with the cited bytes in hand, base64 in `context.resolutions`, so every vector is offline: per consumed surface, no resolver, a resolver that lacks the URI, one that has it, a record without the field, and a record that cites an unresolvable URI and still verifies; plus one that resolves all three. The README lists the consumer defects the set goes red on, vector by vector, the evidence the 2026-09-02 comment on agentrust-io/trace-spec#247 requires of a new drift test, applied here to a conformance set. `tests/test_citation_resolution.py` holds eleven invariants over the set, carries its own copy of the network import sweep, and regenerates the vectors byte-for-byte. The set is registered under `MEASURED_ELSEWHERE` in the adequacy tests, because its outcomes are per-surface and the criteria grade accept/reject; the new function and parameter are entered in the public-surface sweep.

Stated limitations. `agentrust_trace.citation` is not exported from `__init__`; `CitationCheck` is reached by module path, and the export is a small follow-up kept out of this diff. On the `verify_record()` path the schema has already required each surface to be a string; a direct `check_citations()` call with a non-string value at a surface places that value in `evidence["uri"]` as given. `VerificationResult` gains a defaulted field after `trusted_key_thumbprint`, and agentrust-io/trace-spec#347 adds two more in the same place; whichever lands second rebases the field order and one signature hunk. `docs/verification.md` gains a short section beside Step 5, Resolving cited objects, stating the three outcomes and what the check does not do; its `examples/` link follows the absolute form agentrust-io/trace-spec#299 set and resolves on merge. Both network import sweeps glob the package's top level; `adapters/` is not swept by either, which predates this PR.

Verification was run on Windows 10 with Python 3.13.5 (CI's 3.11 and 3.12 were not run), with the five known Windows generator and path failures identified by name, and the same five red in both runs: 5 failed, 1700 passed, 1 skipped against 5 failed, 1650 passed, 1 skipped at `0ec04531`. ruff resolved to 0.16.8 where CI pins 0.16.7; mypy strict and `check_dashes` are clean. No Linux run is claimed.

Refs agentrust-io/trace-spec#190.

## Type of change

- [ ] Schema change
- [x] Example addition

Reference implementation, which the list above has no box for: `verify_record()` gains an optional keyword argument and `VerificationResult` gains a defaulted field. Callers passing nothing new see the same result plus four `not_attempted` rows. `CHANGELOG.md` entry added.

## Spec section

Section 3.1.2 (rule 3 and the resolver paragraph; no normative text changes). Section 3.3 step 4 for the object behind `runtime.rim_uri` and step 6 for the deferred `transparency`; no step names `appraisal.policy_ref` or `model.aibom_uri`, and step 5's policy is the enforced bundle under `policy.bundle_hash`, a different object from the appraisal policy. Section 7 open question 3, by way of section 3.1.2's reference to it, for the deferred surface.

## Checklist

- [ ] Breaking changes marked with `<!-- CHANGED: #NNN: description -->` in spec text: no spec text changes
- [ ] Backward compatibility statement included (for breaking changes): not a breaking spec change; the API note above is the compatibility statement for the package

Tool-assisted: the build, the witness rounds and this description. I answer for all of it.


imran-siddique:
@opento-suggestions both design choices make sense within the scope agreed on #190: check all three citation fields and retain the digest of the bytes returned.

Keep the limit explicit: that digest identifies what this check retrieved. It does not establish that those bytes were used during the original run or appraisal. That remains #280’s question.

Each field should keep its own result and reason when retrieval fails or no check runs. Keeping transparency deferred with a stated reason also matches the existing separation.

This accepts the two design choices. It does not approve the implementation or settle the specification’s outcome names. @lywinged’s review can continue on those terms.

imran-siddique (dismissed):
Approving at the merge commit I just pushed. I ran the behaviour rather than reading it: a resolver that raises gives `unresolvable` with `resolver_raised` and the record still verifies, an absent field gives `not_attempted` with `field_absent`, `transparency` reports `surface_deferred`, a successful resolve records the URI, the SHA-256 of the bytes returned and their count, and a record whose signature fails raises with the resolver never called.

Merging #347 put `profile` and `accepted_profiles` on `VerificationResult`, which conflicted your `citations` field and the public-function keyword test. I kept both sides in `revocation.py` and added both keywords to that test; nothing of yours was dropped.

Issue #342, opened 2026-09-13
verification: successful results can collapse distinct decision-relevant verifier contexts


altrudev:
Reproduced on frozen TRACE revision `6737005e1feefa5739d971a2b71ac123a4235220` using governed DSR job `fe2b6851dcb866fb0ddcb43d639d8a10` on worker `vps-377a113a`.

## Observed behavior

For the same signed TRACE v0.2 record and same trusted key:

- `max_age_seconds=86400` vs `max_age_seconds=None` -> identical successful `VerificationResult`
- `max_future_skew_seconds=300` vs `3600` -> identical successful `VerificationResult`
- `now=iat+60` vs `now=iat+120` -> identical successful `VerificationResult`

The result object contains only:

```
revocation
trusted_key_thumbprint
```

Control: revocation context is preserved. No revocation source yields `no_check_performed`; an empty consulted store yields `verified`, so those results differ.

Opposite-direction control: age policy is decision-relevant. At `now=iat+86401`, `max_age_seconds=86400` rejects as stale while `max_age_seconds=None` accepts.

DSR output:

```json
{
  "age_policy_collision": true,
  "skew_policy_collision": true,
  "verification_time_collision": true,
  "age_policy_semantically_live": true,
  "revocation_noncollision_control": true,
  "result_fields": ["revocation", "trusted_key_thumbprint"]
}
```

DSR status: PASS, exit 0, radial ALLOW, Parallax shadow ALLOW, signed worker result and DDC Action Receipt present.

## Internal consistency point

TRACE's own non-normative `docs/verification-outcome-statements.md` says that, for a "Verified at the required floor" statement, evidence worth retaining includes:

- the record itself;
- the trusted-key thumbprint used;
- **the freshness bounds applied**;
- revocation evidence when consulted;
- the complete appraisal.

`VerificationResult` currently preserves the trusted-key thumbprint and revocation result, but not the freshness bounds or verification time used to reach the successful decision.

That does not prove the result type is wrong by itself: the page is informative, and a caller could retain verifier context separately. It does show that the repository already treats freshness configuration as part of the evidence needed to interpret a retained verification statement.

## Governing invariant

If changing a verifier-side input can change the verification decision for the same signed record, that input is decision-relevant.

A successful result does **not** have to serialize every verifier input. But if a decision-relevant input is omitted from the retained result, then at least one of these has to be true for a later consumer to interpret the result without silently widening it:

[...]

PR #377, opened 2026-09-20
docs: retain freshness context with verification results

altrudev:
Closes the documentation part of #342.

Imran clarified that `VerificationResult` is the outcome of one verification call under its supplied inputs, not a complete assurance statement that can be retained and interpreted independently later.

This adds a small non-normative example to `docs/verification-outcome-statements.md` showing how a caller can retain:

- the exact signed record;
- the complete `VerificationResult`;
- the effective `max_age_seconds`, including explicit `None` when disabled;
- `max_future_skew_seconds`;
- the actual verification time;
- an optional retrievable policy/configuration version.

The example pins `now` explicitly so the retained verification context records the same time the verifier actually used. It does not change the signed Trust Record, schema, verifier API, or conformance behavior.

Based on current main `882a0af`.

AI-assisted; Altru.dev remains responsible for the change and verification.

Signed-off-by: altrudev <266135212+altrudev@users.noreply.github.com>


imran-siddique (approved):
Approving. Pinning `now` in the call is the part that makes the retained context worth keeping: the record says the time the verifier actually used rather than the time a later reader assumes. Keeping an explicit `None` for `max_age_seconds` carries the same weight, since a disabled check and an unrecorded one are different things. Non-normative, already in the nav, and no verifier behaviour changes.

PR #386, opened 2026-09-21
fix: preserve record verification when revocation bundles are malformed

### LINKED ISSUE #382: verify: a revocation bundle with a lone surrogate rejects the record instead of reporting `unverified_for_revocation`
**What is wrong**

A revocation bundle whose text contains a lone surrogate (`"log.example/trace\ud800"` inside `log_id`) makes `verify_record` reject the record:

```
CanonicalizationError: input contains non-UTF-8 codepoints
```

The bundle is untrusted input the verifier fetched; §3.2.3 says a bundle a verifier cannot ground an answer on is reported as `unverified_for_revocation`, and the record's own verification continues. `revocation.py:16` states the same contract: "The outcome is a value in the result, not an exception." The escape is `revocation.py:166-172`: `_canonical_bytes(bundle)` raises `rfc8785.CanonicalizationError`, which subclasses `ValueError`, and the `except ValueError: raise` arm re-raises it out of the digest function and out of `verify_record`.

Reproducer, on the base record from #379 (same `BASE`, `KEY`, `NOW`), with the repository's own example bundle:

```python
import pathlib
ex = json.loads(pathlib.Path("examples/revocation-bundle/01-fresh-well-inside-both-bounds.json").read_text())
good, keys = ex["context"]["bundle"], ex["context"]["trusted_bundle_keys"]
rec = sign_record(json.loads(json.dumps(BASE)), KEY)
verify_record(rec, key_to_jwk(KEY), now=NOW, revocation_bundle=good, trusted_bundle_keys=keys).revocation.outcome
# 'verified'
bad = json.loads(json.dumps(good).replace("log.example/trace", "log.example/trace\\ud800"))
verify_record(rec, key_to_jwk(KEY), now=NOW, revocation_bundle=bad, trusted_bundle_keys=keys)
# CanonicalizationError: input contains non-UTF-8 codepoints
```

The rule it tests: §3.2.3, the `unverified_for_revocation` outcome for a bundle the verifier cannot evaluate.

**Spec section or file**

- `src/agentrust_trace/revocation.py:166-172`
- `spec/trace-v0.2.md` §3.2.3

**Expected behavior**

`VerificationResult` with `revocation.outcome == "unverified_for_revocation"` and `cause == "bundle_malformed"`, evidence naming the canonicalization error; the record itself verifies. `c22` in the differential contexts (a bundle that is not a `TraceRevocationBundle`) already takes that path.

**Impact**

A malformed bundle turns a valid record into a rejection. Since the bundle is fetched, not signed by the record's producer, whoever serves it decides whether records verify: a denial surface on the verifier, and a broken contract for callers reading `result.revocation`.

Found by the differential run in #376 (context c23).

imran-siddique:
## What this changes

A revocation bundle containing a lone surrogate could raise a canonicalization exception and prevent an otherwise valid record from being verified. Catch failure to compute the bundle digest inside `check_bundle` and return `unverified_for_revocation` with cause `bundle_malformed`, a root path and the error. No digest is claimed for bytes that cannot be canonicalized.

The record's own signature is still checked. The direct `bundle_digest` API still raises `ValueError` for invalid input, and authenticated revocation statements continue to reject records. This change does not make unknown revocation status an affirmative appraisal.

Closes #382. Reported by @chernistry.

## Type of change

Implementation bug fix. No schema, wire-format, or normative text changes.

## Spec section

TRACE v0.2 section 3.2.3 and the existing revocation-result contract.

## Validation

- Four regression cases: malformed text in two bundle fields, continued rejection of a bad record signature, and the direct digest API's refusal. Three failed before the fix; the direct-API control passed.
- All 205 signing and revocation-bundle tests pass after the fix.
- Ruff, mypy (13 source files), house-style and diff checks pass.
- Full Windows Python 3.12 suite: 1,965 passed, 6 skipped, and the same five generator-fixture failures reproduced on unchanged main during #383 validation. Hosted Linux results are pending.

## Checklist

- [x] CHANGELOG.md updated
- Breaking-change markers and compatibility statement: not applicable.


lywinged (approved):
Checked at `159194c` against `main` at `3ddcbbf`, which moved under this PR when 385 merged at
16:32. It still merges clean. Your 205 signing and revocation-bundle tests pass. CI has since gone
green on all nine checks, which closes the "Hosted Linux results are pending" line in your
description.

#382 reproduces on `3ddcbbf` and is closed here: the same record and the same lone-surrogate
bundle go from `CanonicalizationError` to `unverified_for_revocation` / `bundle_malformed`, with a
good-bundle control staying `verified` on both trees.

Your three checkable claims hold, checked because approving co-signs them. No `bundle_digest` key
appears in the malformed result; a record tampered after signing still leaves as
`InvalidSignature` with the same bad bundle; and the direct `bundle_digest` still raises, as
`rfc8785.CanonicalizationError`, which subclasses `ValueError`.

I went looking for a reason `except ValueError` is wider than the failure, since `bundle_digest`
also raises a plain `ValueError` for a non-dict bundle. It is not reachable: the shape check
earlier in `check_bundle` already returns `bundle_malformed` for `[1, 2]` before the digest is
computed, so the only `ValueError` that arrives is the canonicalization one. The caveat worth
having is the opposite of the one I started writing: that ordering is what makes the broad catch
sound, so the shape check and the digest call should not be reordered.

Not checked: "authenticated revocation statements continue to reject records". It needs a signed
statement against the record's own key and I did not build one.

---

## What I could not check

  than anything I can run here.
- "authenticated revocation statements continue to reject records" on 386, as above.

Issue #379, opened 2026-09-21
verify: schema `pattern` and `format: uri` are evaluated with Python `re` semantics, so a trailing U+000A, U+000D or U+2028 in `subject` and `appraisal.verifier` verifies


chernistry:
**What is wrong**

The schema's `pattern` constraints and its `format: uri` values are evaluated with Python `re` semantics. JSON Schema 2020-12 specifies ECMA-262 for `pattern`, and the two dialects differ on exactly the characters a display cannot show: in Python `$` also matches before a final newline, and `.` matches U+000D and U+2028; in ECMA-262 `$` matches only at the end and `.` excludes all line terminators. `validate.py:20` builds `Draft202012Validator(..., format_checker=jsonschema.FormatChecker())`, whose `uri` checker anchors the same way.

Four records that the schema excludes under ECMA-262 are accepted and verify (agentrust-trace 0.10.0):

| record | reference |
|---|---|
| `subject` + U+000A | verified |
| `subject` + U+000D | verified |
| `subject` + U+2028 | verified |
| `appraisal.verifier` + U+000A | verified |

Reproducer, public API only:

```python
import hashlib, json
from cryptography.hazmat.primitives.asymmetric.ed25519 import Ed25519PrivateKey
from agentrust_trace import sign_record, key_to_jwk, verify_record

NOW = 1785000000
KEY = Ed25519PrivateKey.from_private_bytes(hashlib.sha256(b"repro").digest())
BASE = {
    "eat_profile": "tag:agentrust-io.com,2026:trace-v0.2", "iat": NOW,
    "subject": "spiffe://factory.example/agent/payments/prod",
    "model": {"provider": "anthropic", "model_id": "claude-sonnet-4-6"},
    "runtime": {"platform": "software-only", "measurement": "sha256:" + "0" * 64, "nonce": "a-nonce-value"},
    "policy": {"bundle_hash": "sha256:" + "a" * 64, "enforcement_mode": "enforce"},
    "data_class": "internal",
    "build_provenance": {"slsa_level": 0, "digest": "sha256:" + "b" * 64},
    "appraisal": {"status": "affirming", "verifier": "https://verifier.example/v1"},
}
for tail in ["\n", "\r", " "]:
    body = json.loads(json.dumps(BASE)); body["subject"] += tail
    verify_record(sign_record(body, KEY), key_to_jwk(KEY), now=NOW)   # returns, does not raise
[...]

PR #387, opened 2026-09-21
fix: reject malformed embedded IPv4 URI literals

### LINKED ISSUE #380: verify: `format: uri` accepts `https://[::1.2.3.04]/`, which RFC 3986 `dec-octet` excludes
**What is wrong**

`appraisal.verifier` is `format: uri`. The value `https://[::1.2.3.04]/` is accepted and the record verifies (agentrust-trace 0.10.0). RFC 3986 §3.2.2 defines `dec-octet` without a leading zero (`"0" / %x31-39 DIGIT / ...`), so `1.2.3.04` is not an IPv4address, the bracketed literal is not an IPv6address, and the string is not a URI.

Reproducer, on the base record from #379 (same `BASE`, `KEY`, `NOW`):

```python
body = json.loads(json.dumps(BASE)); body["appraisal"]["verifier"] = "https://[::1.2.3.04]/"
verify_record(sign_record(body, KEY), key_to_jwk(KEY), now=NOW)   # returns, does not raise
```

The rule it tests: RFC 3986 §3.2.2 `dec-octet`, applied through `format: uri` on `#/properties/appraisal/properties/verifier`.

**Spec section or file**

- `schema/trace-claim.json` `#/properties/appraisal/properties/verifier`
- `src/agentrust_trace/validate.py:20` (`jsonschema.FormatChecker()` and whichever URI validator it resolves at import time)

**Expected behavior**

Rejected as schema-invalid. Either the format checker applies the RFC 3986 grammar for IP literals, or the schema states that `format: uri` is an annotation here and verifiers do not rely on it.

**Impact**

Small on its own: `verifier` is an identifier a relying party resolves, and an unresolvable one fails later. The cost is the disagreement: an implementation that applies RFC 3986 rejects a record the reference accepts, on a field the schema marks with a format.

Found by the differential run in #376 (mutation m32).

imran-siddique:
## What this changes

A signed record containing `https://[::1.2.3.04]/` currently verifies, even though RFC 3986 excludes leading zeros in embedded IPv4 octets. Add an IPv6 address check after the existing URI format check so all eight URI fields reject this value as schema-invalid.

The check is local to TRACE's validator. Valid IPv6 literals, IPvFuture syntax accepted by the existing checker, user information and ports, registered host names, and non-web URIs retain their existing behavior. This does not replace the general URI parser or change the schema.

Closes #380. Thanks to @chernistry for the reproducer.

## Type of change

Python verifier bug fix; no normative or schema change.

## Spec section

TRACE v0.2 section 3.3, schema validation; [RFC 3986 section 3.2.2](https://www.rfc-editor.org/rfc/rfc3986.html#section-3.2.2), IP literals and `dec-octet`.

## Validation

- Before the fix: 34 rejection cases failed, including the signed-record reproducer; 132 controls passed.
- After the fix: 340 URI, schema, and signing tests passed. Tests exercise each IPv4 octet in all eight URI fields, error paths, valid addresses and identifiers, and isolation from jsonschema's global format checker registry.
- Full Windows run: 2,077 passed, 6 skipped, and the same five generator-fixture failures previously reproduced on unchanged `ccbd5b9` (conformance, gap-disclosure, canonicalization-boundary, delegation-link, verifier-compatibility).
- Ruff, mypy (13 source files), the repository style check, and `git diff --check` passed.

The Windows environment uses the pinned development requirements plus `colorama==0.4.6`. The separate regex and trailing-newline issue #379 is outside this change.

## Checklist


lywinged (approved):
Checked at `5c0d1dc` against `3ddcbbf`, merges clean, 340 tests pass, CI green.

The guard reaches further than the reported case, which is worth having in the record. Twelve
shapes, `main` against this branch:

| shape | `3ddcbbf` | 387 |
|---|---|---|
| `https://[::1.2.3.04]/` (the issue) | accept | **reject** |
| `https://[::01.2.3.4]/` first octet | accept | **reject** |
| `https://[::ffff:1.2.3.04]/` v4-mapped | accept | **reject** |
| `HTTPS://[::1.2.3.04]/` uppercase scheme | accept | **reject** |
| `https://user@[::1.2.3.04]:443/` userinfo and port | accept | **reject** |
| `coap+tcp://[::1.2.3.04]/` `+` in scheme | accept | **reject** |
| `https://[::1]/`, `[v7.x]`, `[::ffff:1.2.3.4]`, `example.com` | accept | accept |
| `https://[fe80::1%25eth0]/`, `https://[::1.2.3.4]./` | reject | reject |

Six newly rejected. No regression among the twelve, which is the domain I swept rather than a
statement about every URI; the last row is the base checker refusing on both sides
rather than anything you added, so RFC 6874 zone ids are not yours to answer here.

**`raises=ValueError` on the decorator is load-bearing, and it is worth a comment saying why.**
Exactly one of the malformed shapes depends on it: `https://[::1.2.3.04]/` is the only one
`base_uri_checker.conforms` lets through, so it is the only one that reaches `IPv6Address()` and
raises. The other three are refused earlier and return `False`. Drop the argument and the issue's
own case leaves as `AddressValueError` with an `ipaddress` message instead of the schema refusal.

---

## What I could not check

  than anything I can run here.

Issue #226, opened 2026-08-26
The references rel registry: four governance requests are one question


imran-siddique:
## Why this exists

Four governance-shaped requests have arrived against this repository in about a week, from four unconnected contributors:

| issue | ask |
|---|---|
| #191 | binding attributable human-approval outcomes to `step_up` and `defer` execution evidence |
| #209 | a public-sector auditability profile: policy exceptions, incident and escalation triggers |
| #220 | per-action governance-decision evidence, independently verifiable with keys the operator does not control |
| external | a transition-closure layer, raised on #191 |

@lywinged identified the pattern on #209 before the fourth arrived:

> All three need the same thing, which is a way to bind a governance fact that lives outside the record to the record itself. The demand is real and it is arriving faster than the profiles for it.

That reading is correct, and the consequence is that these should not be answered as four schema proposals.

## The mechanism already exists

§3.1.2's `references` block, merged in #197, binds an external fact by `rel`, `id`, `resolver`, optional `digest` and optional `retention`, with the limit stated normatively: a verifier MUST NOT reject a record because a reference cannot be resolved, and MUST NOT treat a resolved reference as attested evidence.

Two registered values already cover part of the ground:

- `authorized-intent`, an authorization decided before execution, held in another system
- `approval-outcome`, an attributable human approval attached to a step-up or defer decision

So the composability question is settled. What is not settled is which further relationships get registered values, and what a verifier may conclude from each.

## What this issue tracks

The `rel` registry as one question:

1. **Which relationships need a registered value.** Candidates from the four threads: a per-action policy evaluation (#220), a policy exception or rule violation (#209), an incident or escalation trigger (#209), and whatever #191 settles on beyond `approval-outcome`.
2. **What each one's referenced object is**, precisely enough that two implementations resolve the same thing.
3. **What a verifier may and may not conclude** from a resolved reference of each kind. §3.1.2 rule 3 sets the floor; each `rel` may need to say more.
4. **Whether the registry is normative or informative**, and therefore whether it needs an organisational sponsor under `GOVERNANCE.md#who-may-author-normative-text`.

Point 4 should be answered first, because it determines who can write the rest.

## What this is not

Not a new schema layer, and not a new artifact. Each of the four proposals is smaller when expressed as a `rel` than as its own structure, and answering them separately would produce four overlapping mechanisms for one relationship type.

## Related

#191, #209, #220, and #197 which merged the `references` block.

nutstrut:
[...]

PR #389, opened 2026-09-21
references: register condition-appraisal and document the rel registry (#226)

chernistry:
## What this changes

Refs #226. Implements the ruling there: the registry stays informative and open, the verifier rules of §3.1.2 stay normative, and `condition-appraisal` is proposed in one focused change.

- **`condition-appraisal` joins the registered `rel` values in §3.1.2**: an independent check's finding on whether a stated condition is established by a stated subject.
- **`docs/references-registry.md`** (new): what registration does and does not do; the referenced object for `condition-appraisal` as facts rather than a wire format; what a verifier may conclude from a resolved one; how a name is added.
- **One paragraph in §3.1.2, placed below the four rules**, stating that the registry is informative and open and that registering a value changes none of them. It carries no requirement keyword. If it reads as a change to normative text rather than a statement about the list, I will move it to the registry document and leave the list entry only.
- **`examples/condition-appraisal/`**: five signed records from a seeded generator against an appraisal store committed beside them: confirmed, altered after issue, a fail that verifies exactly as the pass does, unresolvable, and an issuer whose key the relying party does not hold. `tests/test_condition_appraisal_fixtures.py` recomputes every verdict from the committed bytes and re-runs the generator against them.
- The schema's `rel` description and its packaged copy, the regenerated v0.3 draft, the `rel` row in `docs/schema.md`, `examples/README.md`, the `registered` tuple in `tests/test_references_block.py` (which now also checks that the registry document names every registered value), `tests/test_adequacy_all_sets.py`, `mkdocs.yml`, and `CHANGELOG.md`.

The quoted normative text this leans on is rule 3 of §3.1.2, unchanged by this PR:

> A verifier MUST NOT reject a record because an entry in `references` cannot be resolved, and MUST NOT treat a resolved reference as attested evidence.

## What this does not do

- No requirement, schema shape or wire-format change. A `pass` is not attested evidence and a `fail` does not touch validity: cases `01` and `03` differ only in the reference they carry and verify identically, and `04` verifies with a reference that resolves to nothing.
- Re-running a cited check stays a separate proposal, as ruled on #226.
- The three earlier registered values are listed with the documents that define them; none is redefined here.

## Type of change

- [x] Example addition

The schema change is the `rel` description gaining the name; `rel` stays a plain string, so no shape changes.

## Spec section

§3.1.2 `references`: the registered-value list, and one paragraph below the four rules.

## Checklist

- [x] `CHANGELOG.md` updated
- [x] Changes marked with `<!-- CHANGED: #226 - ... -->` in spec text
- [ ] Backward compatibility statement included (for breaking changes) — not a breaking change

## Checks

`pytest` (1987 passed, 6 skipped), `ruff check src tests scripts`, `tools/check_dashes.py` and `mypy src/agentrust_trace` are green locally on 08c74f5. Regenerating the vectors reproduces the committed bytes; `schema/trace-claim.json` and its packaged copy are byte-identical.


imran-siddique:
Alex, keep the paragraph below the four rules in §3.1.2. It makes our earlier decision clear: the registry is informative and open, while the verifier rules remain binding. No need to move it.

imran-siddique (approved):
Checked at `08c74f5`. This follows the decision on #226. The registry remains open and informative, and the four verifier rules are unchanged. All 149 related tests pass, including altered, missing and failed appraisal examples. Approved.

